PR #1, opened 2018-01-28
adds overview and redux-promise-middleware skeleton

priley86:
* adds skeleton mapping wizard / planning wizard components
* adds redux-promise-middleware, redux-logger, seamless-immutable
* adds react-equalizer (as opposed to jquery-match-height 😸 ) for cards (will start the contribution for cards in pf-react soon)
* adds prettier-eslint
* adds axios (to be followed soon with some axios-mock-store examples)
* updates patternfly-react to latest
* downgrades react-bootstrap to `0.31.5` b/c there is some annoying modal warnings in `0.32` still


priley86:
a few updates from today... 

* add some redux-promise-middleware action snapshots 🏎 🏎 🏎 
* implement modal `onExited` so exit animation is smoother
* use `componentRegistry.registerMultiple` (and update ui-classic v2v branch w/ latest)
* pass urls to containers via the componentRegistry instead of using default props
* add temporary `markupWithProps` extension to componentRegistry b/c `data` is currently not merged with `currentComponent.data` (this was to support `ownProps` being passed in)
* use `<Route>` `render` prop as opposed to `component` prop so that we can pass the `componentRegistry.markup()` content instead 
* mock the componentRegistry so tests pass (*window.MiqReact.componentRegistry* will fail it w/o a mock)

PR #22, opened 2018-01-31
Add gettext initialization and locale subdirectory

mzazrivec:
This pull request contains:
1. initializer to register the plugin's i18n domain with ManageIQ
2. `locale/` subdirectory with gettext catalogs (currently empty). The catalogs were extracted by running the following command in the ManageIQ (core) directory:
```
bundle exec rake locale:plugin:find[MiqV2vUI]
```
The same command will be used to update the catalogs in the future.

### DIFF
--- a/config/initializers/gettext.rb
+++ b/config/initializers/gettext.rb
@@ -1,0 +1,1 @@
+Vmdb::Gettext::Domains.add_domain('MiqV2vUI', MiqV2vUI::Engine.root.join('locale').to_s, :po)
--- a/locale/MiqV2vUI.pot
+++ b/locale/MiqV2vUI.pot
@@ -1,0 +1,19 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the MiqV2vUI package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: MiqV2vUI 1.0.0\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2018-01-31 15:44+0100\n"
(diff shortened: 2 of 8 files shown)


mzazrivec:
Come to think of it, changes in this PR will only make sure strings in `.haml` `.erb` `.rb` and standard javascript files will be extracted. We still need to address / figure out the way to annotate & extract strings from react sources (another PR).

priley86:
ok good to know - ok this to merge this for those templates then? I would propose looking into `intl`,`intl-messageformat` for this... do you want me to get started on this or would you like to take a look? There is a pretty good pattern in oVirt existing today here: https://github.com/oVirt/ovirt-engine-dashboard/blob/master/package.json#L29

AparnaKarve:
We also have inline html within the JS files as shown below -

For e.g. -
```js
const loadingContents = () => (
  <EmptyState>
    <Spinner size="lg" className="blank-slate-pf-icon" loading />
    <EmptyState.Action>
      <h3>Loading Wizard</h3>
    </EmptyState.Action>
    <EmptyState.Action secondary>
      <p>
        Lorem ipsum dolor sit amet, porta at suspendisse ac, ut wisi vivamus,
        lorem sociosqu eget nunc amet.{' '}
      </p>
    </EmptyState.Action>
  </EmptyState>
);
```

So the text here would need to be set with React Props from the component (I'm guessing)

We may require https://www.npmjs.com/package/i18n-react or an equivalent package.
/cc @priley86

ohadlevy:
we can simply wrap the text with ```__('some string')```

Foreman did a lot of work on this already - see https://github.com/theforeman/foreman/pull/5184 for some details.

priley86:
ok thanks @ohadlevy - will start there.

priley86:
@mzazrivec @ohadlevy i've pushed an initial commit for the client side *only* which should enable us to start using i18n for dates/times/messages. Can you please take a look and try wiring up the server side? My Ruby is not so great...

https://github.com/priley86/miq_v2v_ui_plugin/commit/9db79b62ae7c26daf5db551eac325c625f3d209c

priley86:
included in #37

PR #24, opened 2018-01-31
remove package lock and use yarn

priley86:


priley86:
oops needs rebase.. fixing...

priley86:
@mturley @AparnaKarve @ZitaNemeckova ok to merge?

.npmrc
```diff
@@ -0,0 +1,2 @@
+# using yarn
+package-lock = false
```
AparnaKarve: @priley86 Is `yarn.lock = true` implicit here?

priley86: @AparnaKarve i'm not sure - maybe you have another suggestion? This was the setting in ui-classic.

AparnaKarve: Maybe it is implicit.
Let's go with this setting, since classic-ui also has it.

himdel: `yarn.lock` is generated always when you run yarn, and there's no real way to disable it,
but we can just keep `.gitignore`ing it, same as ui-classic.

Issue #30, opened 2018-02-02
Stub "Create Migration Plan" Wizard


serenamarie125:
Simple wizard flow at this point first - so we can enable the other wizard steps to be developed simultaneously. 

There are 3 steps in the wizard:
- General
- VMs
- Results

Stub the wizard steps indicators and the wizard back/next in the footer for infra mapping wizard that would be great. Let's just tackle that one first and try the planning wizard next...

mturley:
This will be closed by #38.

Issue #2, opened 2018-01-30
Stub "Create Infrastructure Mapping" Wizard


mturley:
This will probably need to be broken down further, but adding it as a high-level task to assign myself.

mturley:
Tentative plan to start on this:

- [ ] Add a barebones wizard with only one step (or maybe a stub for each real step).
- [ ] in that "Step 1" render area, add an empty dropdown for source clusters (and maybe a reload button? probably just for testing, wasn't in the mock)
- [ ] create a reducer for clusters
- [ ] wire up the API (or stub API) to the cluster reducer
- [ ] wire up the cluster reducer to the clusters dropdown

priley86:
@mturley do you mind starting this one? we just need a simple wizard flow at this point first - so we can enable the other wizard steps to be developed simultaneously. If you can stub the wizard steps indicators and the wizard back/next in the footer for infra mapping wizard that would be great. Let's just tackle that one first and try the planning wizard next...

mturley:
Yep, I'm on it.

mturley:
This will be closed by #38.

PR #38, opened 2018-02-07
Add ModalWizard, MappingWizardBody and PlanWizardBody components

mturley:
Includes i18n out of the box. Tried to break this down into logical components. I think maybe the `ModalWizard`, `ModalWizardBody` and `ModalWizardStateProvider` can be moved up to patternfly-react?

### DIFF
--- a/app/javascript/common/helpers.js
+++ b/app/javascript/common/helpers.js
@@ -1,14 +1,16 @@
 // todo: expose this from patternfly-react instead?
+export const getDisplayName = Component =>
+  Component.displayName || Component.name || 'Component';
+
+// TODO upgrade to latest patternfly-react after next release, these are exposed now:
 export const bindMethods = (context, methods) => {
   methods.forEach(method => {
     context[method] = context[method].bind(context);
   });
 };
-
 export const noop = Function.prototype;
-
-export const getDisplayName = Component =>
-  Component.displayName || Component.name || 'Component';
+export const selectKeys = (obj, keys, fn = val => val) =>
+  keys.reduce((values, key) => ({ ...values, [key]: fn(obj[key]) }), {});
 
 // https://github.com/yahoo/react-intl/wiki/Upgrade-Guide#flatten-messages-object
 export const flattenMessages = (nestedMessages, prefix = '') =>
--- a/app/javascript/common/helpers.js
+++ b/app/javascript/common/helpers.js
@@ -2,16 +2,6 @@
 export const getDisplayName = Component =>
   Component.displayName || Component.name || 'Component';
 
-// TODO upgrade to latest patternfly-react after next release, these are exposed now:
-export const bindMethods = (context, methods) => {
-  methods.forEach(method => {
-    context[method] = context[method].bind(context);
-  });
-};
-export const noop = Function.prototype;
-export const selectKeys = (obj, keys, fn = val => val) =>
-  keys.reduce((values, key) => ({ ...values, [key]: fn(obj[key]) }), {});
-
 // https://github.com/yahoo/react-intl/wiki/Upgrade-Guide#flatten-messages-object
 export const flattenMessages = (nestedMessages, prefix = '') =>
   Object.keys(nestedMessages).reduce((messages, key) => {
--- a/app/javascript/common/messages.js
+++ b/app/javascript/common/messages.js
@@ -9,11 +9,27 @@ export default {
       emptyPlanCardInfo: 'Create a migration plan to start migrating.',
       emptyPlanCardButton: 'Create Migration Plan'
     },
-    mappingWizard: {
-      title: 'Infrastructure Mapping Wizard',
+    wizard: {
       cancel: 'Cancel',
       back: 'Back',
-      next: 'Next'
+      next: 'Next',
+      close: 'Close'
+    },
+    mappingWizard: {
+      title: 'Infrastructure Mapping Wizard',
+      loadingTitle: 'Loading Infrastructure Mappings...',
+      loadingMessage: 'Just a sec.',
+      general: 'General',
+      clusters: 'Clusters',
+      datastores: 'Datastores',
+      networks: 'Networks',
+      results: 'Results'
+    },
+    planWizard: {
+      title: 'Migration Plan Wizard',
+      loadingTitle: 'Loading Clusters...',
+      loadingMessage: 'This may take a minute.',
+      vms: 'VMs'
     }
   }
 };
--- a/app/javascript/components/MappingWizard/index.js
+++ b/app/javascript/components/MappingWizard/index.js
@@ -2,31 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
-import {
-  Button,
-  Icon,
-  EmptyState,
-  Modal,
-  Spinner,
-  Wizard
-} from 'patternfly-react';
+import { noop, selectKeys } from '../../common/helpers';
 import * as MappingWizardActions from '../../redux/actions/mappingWizard';
-import { noop } from '../../common/helpers';
-
-const loadingContents = () => (
-  <EmptyState>
-    <Spinner size="lg" className="blank-slate-pf-icon" loading />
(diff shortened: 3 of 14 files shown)


mturley:
@serenamarie125, here they are! Sorry they are a day and a half later than expected.

screen shot 2018-02-07 at 6 30 43 am

screen shot 2018-02-07 at 6 31 37 am

mturley:
Ok @priley86 Squashed all my changes for these wizard stubs. Looks like I was having some CI failures, waiting on that again. Tests well in the browser on my end.

mturley:
I think I just fixed the eslint issue that was failing CI.

mturley:
Closes #2 
Closes #30

mturley:
Updated snapshots and squashed again. That should do it.

app/javascript/common/messages.js
```diff
@@ -9,11 +9,27 @@ export default {
...
-      title: 'Infrastructure Mapping Wizard',
+    wizard: {
       cancel: 'Cancel',
       back: 'Back',
-      next: 'Next'
+      next: 'Next',
+      close: 'Close'
+    },
+    mappingWizard: {
+      title: 'Infrastructure Mapping Wizard',
```
ohadlevy: why not to extract the strings using ```__('Infrastructure Mapping Wizard)```?

@priley86 did I miss understand how we handling i18n? (sorry was away for a few days).

mturley:
@ohadlevy that's what I thought we might be doing too, but I was just copying what I'd seen in @priley86's plugin already.

app/javascript/common/helpers.js
```diff
@@ -1,14 +1,16 @@
 // todo: expose this from patternfly-react instead?
+export const getDisplayName = Component =>
+  Component.displayName || Component.name || 'Component';
+
+// TODO upgrade to latest patternfly-react after next release, these are exposed now:
```
priley86: ok yep - we can do this in a subsequent PR...

ohadlevy:
or just do it now and rebase? :)

app/javascript/common/messages.js
priley86: hey @ohadlevy would you like to use i18n.gettext the same way which Foreman does here for text?
https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/i18n.js

I somehow missed this was being done already when looking at the other i18n PR for Dates. I can quickly switch to this format if it works better here in mIQ (although I think that using `react-intl` messages is a better long term solution than exposing it as a global on the window).  Can you detail your plans a bit further there?

Happy to adjust this but I think it's outside the scope of this PR and we can fix it in a subsequent one.

ohadlevy:
I think that the easiest path, I agree that we need to revisit future
plans, but for now that should work with the existing UI infrastructure?

app/javascript/components/MappingWizard/index.js
```diff
@@ -2,84 +2,39 @@ import React from 'react';
...
   componentDidMount() {
     const { url, fetchSourceClusters } = this.props;
 
     fetchSourceClusters(url);
   }
+
   render() {
-    const { showWizard, onHide, onExited, sourceClusters } = this.props;
+    const { sourceClusters } = this.props;
+    const loaded = sourceClusters && sourceClusters.length > 0;
```
priley86: Redux is already injecting `isFetching` inside `this.props` here so i think we can just use that instead...

app/javascript/components/ModalWizard/index.js
```diff
@@ -0,0 +1,9 @@
+import ModalWizard from './ModalWizard';
+import ModalWizardBody from './ModalWizardBody';
+import ModalWizardStateProvider from './ModalWizardStateProvider';
+
+// TODO maybe move these up to patternfly-react?
```
priley86: potentially ModalWizard can...i think `ModalWizardStateProvider` can be lifted to Redux, but I would like to go ahead with this approach for now to keep us moving...

app/javascript/components/ModalWizard/ModalWizardStateProvider.js
```diff
@@ -0,0 +1,55 @@
...
+class ModalWizardStateProvider extends React.Component {
+  constructor() {
+    super();
+    this.state = { activeStepIndex: 0 };
+    bindMethods(this, ['prevStep', 'nextStep']);
+  }
+
+  prevStep() {
+    const { activeStepIndex } = this.state;
+    this.setState({ activeStepIndex: Math.max(activeStepIndex - 1, 0) });
```
priley86: i think `activeStepIndex` can be lifted to Redux... let's go with this for now though! I will revisit it some in spare time soon and open a subsequent PR. This will keep us iterating fast for now and enable development of the other steps.

app/javascript/common/helpers.js
priley86: ok fine... let's go ahead and fix this!

priley86: just update pf-react...and it should be there...

priley86:
@mturley i am happy to see this PR merge to keep us moving if we can just tackle two things:

* reference the pf-react helpers that are currently available
* use `isFetching` coming from redux store...as it's already there.

priley86:
@ohadlevy re: 18n. I responded in [here](https://github.com/priley86/miq_v2v_ui_plugin/issues/19#issuecomment-363807373). I think this is potentially the best option but still considering this. Injecting `gettext` language data into ReactIntl messages should work as an iterim step?

app/javascript/common/helpers.js
mturley: Done.

app/javascript/components/MappingWizard/index.js
mturley: Ok. Done.

mturley:
Pushing updates in a moment...

app/javascript/components/ModalWizard/index.js
mturley: Yeah, sounds good. That was more a long-term idea-- i'll remove the comment for now.

app/javascript/components/ModalWizard/ModalWizardStateProvider.js
mturley: Yep! The reason I contained the state in a small `StateProvider` component is so it could be easily replaced later.

mturley:
@serenamarie125 that will probably involve some minor CSS tweaks or perhaps just className tweaks. I'll have to look into that as a separate issue.

mturley:
@priley86 this should be all good to go now, I think.

priley86:
great job!... thanks a lot.

app/javascript/common/messages.js
AparnaKarve: The `__("Text")` usage involves 2 things -

- extraction
A rake task that extracts all the strings with `__("...")` from the repo, which are then sent to the translators

- translation
The actual translation itself. All instances of `__("...")` would be substituted with the actual text from the language locale files that we receive from the translators

@priley86 With `react-intl`, do we need to explicitly extract the strings out?
With the `__("Text")` approach, we can skip the extraction step since we already have a process in place to do that.

priley86: hey @AparnaKarve - thanks for clarifying. I was slightly confused on this.

If it's all done via Rake tasks now and that process works, it's OK with me. But maybe you can weigh how much additional effort it would be to add `babel-plugin-react-intl` to do the extraction like the guide [here](https://medium.freecodecamp.org/internationalization-in-react-7264738274a0) does? (yes, `react-intl` would also do the extraction, but additional effort).

I'm really torn on using both i18n approaches (`react-intl` for dates and numbers, and using `gettext` for messages). But will trust your best judgement here...

AparnaKarve: @priley86 I think the process of extracting strings with `babel-plugin-react-intl`  might be comparable with the extraction process in Service UI where we use the `translate` angular directive.
```js
<span class="heading-class">{{notificationGroup.heading | translate}}</span>
```

@AllenBW and @chalettu may have some insight into that.

AparnaKarve: keeping @mzazrivec in the loop too.

PR #42, opened 2018-02-12
Mapping wizard step1 Form

AparnaKarve:
Form for Infrastructure Mapping Wizard / Step 1

- Introduces `redux-form`
- Implements Mapping Wizard / Step 1 form
- Implements validation in the form for the `Required` field
- blocks stepping further into the wizard if form has validation errors
- destroys the state of the form on Cancel/Close

Next Button disabled due to validation errors in the form -

screen shot 2018-02-15 at 4 58 33 pm

----------------

Next Button enabled since the form is valid -

screen shot 2018-02-15 at 4 56 57 pm

----------------

Error displayed on the field since the user went past the required field -

screen shot 2018-02-15 at 4 57 16 pm

Fixes #29 #45 #46 #47

PR #40, opened 2018-02-09
[WIP] Create DualPaneMapper component and implement step 2 of mapping wizard

michaelkro:
**[WIP]**

Initial implementation and integration of `DualPaneMapper`

CI failure is due to a `setState` call in `componentDidMount`.  Just doing this to stub data for now

https://github.com/priley86/miq_v2v_ui_plugin/issues/26

# Notes
1. `DualPaneMapper.scss` was copied and pasted from [here](https://github.com/strategicdesignteam/labs-console/blob/master/public/app.css#L166), with some minor customization.
2.  The Target and Source clusters panes are less wide than what you see in the mocks here [V2V POC](https://docs.google.com/presentation/d/1BLPWLYswfMNaddsozEMokiNXicnnmkF6I88necLJzfs/edit#slide=id.g2faecbaabf_0_12).  This is because our actual Wizard Modal is significantly less wide than what is depicted in the mocks.
# Todo:

- [ ] Sort clusters in lists.  Do we want them sorted alphabetically?  This could also be handled by a redux selector function.
- [x] Implement `x of y items selected` info at bottom of Source Clusters pane as shown in mocks

PR #43, opened 2018-02-13
refactor app structure, start clusters actions/reducers

### LINKED ISSUE #3: Actions and reducers for loading Clusters

priley86:
apologies this took so long... i was doing my due diligence here first 😸 

Here's what this PR does:
* Adopts Foreman domain level approach being introduced [here](https://github.com/theforeman/foreman/pull/5240) for redux connected components.
* Migrates app structure to Ryan Florence's [shared modules structure](https://gist.github.com/ryanflorence/daafb1e3cb8ad740b346#shared-modules) to more closely mimic routing/navigation structure
* connects the Overview / PlanWizard / MappingWizard to the Redux store instead of passing props (so we can time travel a bit better ✈️ 😄). 
* adds source/target clusters actions/reducers/tests and connects them to Step 2 to prepare for #40 
* Adds MappingWizard/PlanWizard selectors
* Adds base css/scss structure started previously

Would like to merge this first (if we aren't opposed), then adopt #42 followed by #40 

Closes #3


app/javascript/app/screens/App/Overview/OverviewReducer.js
```diff
@@ -0,0 +1,42 @@
...
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case SHOW_MAPPING_WIZARD:
+      return Immutable.merge(state, {
+        mappingWizardVisible: true,
+        hideMappingWizard: false
+      });
+    case HIDE_MAPPING_WIZARD:
+      return state.set('hideMappingWizard', true);
```
michaelkro: Are you using `Immutable.set()` here?

app/javascript/app/index.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { Provider } from 'react-redux';
 import { BrowserRouter as Router } from 'react-router-dom';
-import Routes from './Routes';
-import store from '../../../redux';
+import Routes from './config/Routes';
+import store from '../redux';
 
 // todo: convert this to use ManageIQ.redux.addReducer, ManageIQ.redux.store
 // <Provider store={store}>{componentRegistry.markup(component, data, store)}</Provider>,
```
AparnaKarve: `app/javascript/app...` is a little repetitive.

I know it follows the convention [here](https://gist.github.com/ryanflorence/daafb1e3cb8ad740b346#shared-modules) -- but just checking if we should consider a different folder name instead of `app`

app/javascript/app/screens/App/Overview/Overview.js
```diff
@@ -2,77 +2,33 @@ import React from 'react';
...
+    bindMethods(this, ['getNodes']);
+
+    this.mappingWizard = componentRegistry.markup(
+      'MappingWizardContainer',
+      props.store
+    );
+    this.planWizard = componentRegistry.markup(
+      'PlanWizardContainer',
+      props.store
+    );
```
AparnaKarve: Nice - this looks much cleaner.

michaelkro:
fwiw, I was able to integrate the layout of https://github.com/priley86/miq_v2v_ui_plugin/pull/40 with minimal effort.

fullscreen_2_13_18__3_07_pm

app/javascript/app/index.js
priley86: hey @AparnaKarve - do you have another suggestion? I am happy to change it quickly today... I was on the fence myself (knowing that `app` is Rails plugin app and also the React front end app). I think this was more to prepare for long term efforts.

priley86: i think per our conversation, we can go with "react" instead. I am updating this now...

priley86: updated... can you confirm @AparnaKarve ? If it's good... please merge. ;)

priley86:
One thing I missed here was some simple reducer tests... will add these in subsequent PRs. Avi has a good example here: https://github.com/theforeman/foreman/pull/5240/files#diff-3c5666f4594243aff3722dc37633a930

priley86:
thanks @michaelkro

michaelkro:
🎉

PR #48, opened 2018-02-16
Menu items are added via plugin

ZitaNemeckova:
**Wait** for [this PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/3347) to be merged!

Moves definition of Menu items from `manageiq-ui-classic` to `miq_v2v_ui_plugin` as suggested [here](https://github.com/martinpovolny/miq_plugin_example/blob/master/lib/miq_plugin_example/engine.rb).
Update `README.md`. Special commit with menu items is no longer needed.

### DIFF
--- a/lib/miq_v2v_ui/engine.rb
+++ b/lib/miq_v2v_ui/engine.rb
@@ -10,5 +10,14 @@ module MiqV2vUI
       app.config.assets.paths << root.join('assets', 'images').to_s
     end
 
+    initializer 'plugin' do
+      Menu::CustomLoader.register(
+        Menu::Section.new(:compute, N_("Compute"), 'pficon pficon-cpu', [
+          Menu::Section.new(:migration, N_("Migration"), 'fa fa-plus', [
+            Menu::Item.new('overview', N_('Overview'), 'miq_report', {:feature => 'miq_report', :any => true}, '/migration')
+          ])
+        ])
+      )
+    end
   end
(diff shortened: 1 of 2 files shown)


lib/miq_v2v_ui/engine.rb
```diff
@@ -10,5 +10,12 @@ def vmdb_plugin?
       app.config.assets.paths << root.join('assets', 'images').to_s
     end
 
+    initializer 'plugin' do
+      Menu::CustomLoader.register(
+        Menu::Section.new(:migration, N_("Migration"), 'fa fa-plus', [
+          Menu::Item.new('overview', N_('Overview'), 'miq_report', {:feature => 'miq_report', :any => true}, '/migration')
```
priley86: can you also add this one @ZitaNemeckova ?
https://github.com/priley86/manageiq-ui-classic/commit/f8792b86b97ee9e8133e700529d093f9845060fd

priley86: after we merge #62 ..

priley86:
We should be target `migration-poc` as the base branch now too...

himdel:
Just merged https://github.com/ManageIQ/manageiq-ui-classic/pull/3347 .. so this one should be good to go :)

ZitaNemeckova:
@priley86 please have a look when you can :)

PR #49, opened 2018-02-16
Update plugin name in README.md

ZitaNemeckova:
Gem has incorrect name. It should be `miq_v2v_ui` not `miq_v2v_ui_plugin`. And minor specification where to add the command.


priley86:
thanks!

PR #53, opened 2018-02-20
Fix ModalWizard next button disabled state

michaelkro:
Currently, ModalWizard reports an error if _any_ form in the entire
wizard is invalid. This results in a bug where, if a user is on a step
with a form error, and then clicks the back button, the `next` button
becomes permanently disabled (even if the current form is valid)

To prevent this, we also check to make sure that we are on the proper
step before disabling the button.

Included in this PR
* Disable the next button on the MappingWizardClustersStep if there are
no mappings
* Fix the global error behavior
* Remove lodash as it is no longer being used

# Notes
1. Not sure if this will play nice with with i18n [reduxFormMap](https://github.com/priley86/miq_v2v_ui_plugin/pull/53/files#diff-efbb747e6ab39e19b2b58601cfbc901fR6)

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizard.js
+++ b/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizard.js
@@ -4,11 +4,11 @@ import { noop, Modal, Wizard, Icon, Button } from 'patternfly-react';
 import { connect } from 'react-redux';
 
 const reduxFormMap = {
-  'Infrastructure Mapping Wizard': [
+  [__('Infrastructure Mapping Wizard')]: [
     'mappingWizardGeneralStep',
     'mappingWizardClustersStep'
   ],
-  'Migration Plan Wizard': []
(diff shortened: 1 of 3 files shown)


app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizard.js
```diff
@@ -2,7 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { noop, Modal, Wizard, Icon, Button } from 'patternfly-react';
 import { connect } from 'react-redux';
-import { find } from 'lodash';
+
+const reduxFormMap = {
+  'Infrastructure Mapping Wizard': [
```
priley86: i think we just construct it using __('Infrastructure Mapping Wizard') so it lines up i guess...

nice job on this  fix...

priley86:
Closing... this has been captured in #58

PR #55, opened 2018-02-20
transformation_mappings

### LINKED ISSUE #7: Actions and reducers for managing Infrastructure Mapping objects
I'm going to include Cluster Mapping, Datastore Mapping and Network Mapping in this issue/task because it will largely all be the same code (probably).

priley86:
closes #7  (redux actions/reducers for transformation_mappings)

* go ahead and move Cards into individual components


priley86:
* injects transformationMappings into plan wizard (and handles disable/enable button logic)

priley86:
@AparnaKarve @mturley were you able to integrate this branch into Step 1 of Plan Wizard? Would it be OK to go ahead and merge this to close #7  ?

AparnaKarve:
@priley86 In Plan Wizard Step1, there would be no API calls to fetch the list of transformation mappings? And the dropdown there, should always rely on the injected `transformationMappings` array?

AparnaKarve:
@priley86 I was wondering if the GET request to fetch the transformation mappings needs to happen in the Plan Wizard Step1 rather than in the Overview -- in order to prolong the API call until you actually need it.

priley86:
@AparnaKarve per our discussion - we can keep it in Overview (since it is also used to drive enable/disable button).

also, i have changed the base branch here to `migration-poc`. Let me know if this is OK to merge.

app/javascript/react/screens/App/Overview/overview.fixtures.js
```diff
@@ -4,5 +4,31 @@ export const initialState = Immutable({
...
+    transformation_mapping_items: [
+      { source: '/api/clusters/10', destination: '/api/clusters/1' },
+      { source: '/api/clusters/11', destination: '/api/clusters/1' },
+      { source: '/api/date_stores/12', destination: '/api/data_stores/13' },
+      { source: '/api/data_stores/2', destination: '/api/data_stores/13' },
+      { source: '/api/lans/2', destination: '/api/lans/13' },
+      { source: '/api/lans/3', destination: '/api/lans/13' }
+    ]
+  }
+];
```
AparnaKarve: I did not see an `id` field. Can we add that?

priley86: updated.

AparnaKarve: Thanks - That looks good!

AparnaKarve:
Looks good. Merging.

PR #67, opened 2018-02-27
WIP (Low Priority): Remove ModalWizardStateProvider and lift all reduxForm related code out of ModalWizard

mturley:
I think merging this before we move on will help prevent confusion around the `ModalWizardStateProvider`, which was a misguided attempt at stubbing state behavior and a pattern I don't want to propagate. The behavior of having ModalWizard know about redux and redux-form was also confusing.  With this PR, I have lifted all redux stuff up to the connected components themselves, and kept it out of the presentational components.

This wasn't strictly required, but I think it'll save us some time we might have wasted getting confused by this stuff.


mturley:
NOTE: This is currently not testable, because it imports `{ controlled } from 'patternfly-react'`, and that export was broken.  After we merge this: https://github.com/patternfly/patternfly-react/pull/255, and update patternfly-react here, this PR will be viable.

@priley86, I'd like to merge this ASAP so we can resolve any conflicts it causes, and it is blocked by pf-react#255. So if we could merge that (will that automatically do a pf-react release), I'll update this to the new release of pf-react immediately after. If we can merge both early tomorrow we will unblock everyone.

mturley:
@priley86 sounds good!

The rationale here behind using `controlled` instead of redux for that one piece of state (`activeStepIndex`), is twofold:
* I want to try and make `ModalWizard` a patternfly-react component, and just like VerticalNav, I wanted it to be able to "just work" without redux optionally, so controlled state made sense.
* Moving that little piece of state up to redux just so that we can have its value above ModalWizard, and calculate a `disableNextStep` prop, seemed like overkill for this POC-- so instead, to lift that value up into that scope, I passed a `shouldDisableNextStep` callback which takes `activeStepIndex`.

I imagine that if we eventually do lift `activeStepIndex` up to redux, we can change that `shouldDisableNextStep` callback prop into an `disableNextStep` boolean, but I'd say we just leave the controlled code for `activeStepIndex` in place (for patternfly-react use later) and just override it.

What things do we differ on?  We can diverge the two wizard implementations I guess, but I was assuming we would consolidate some of the duplicated code and have them be literally the same implementation. at least where possible.

priley86:
so I guess the way I am thinking about this now (after looking at it some more and trying to understand that we are trying to fully abstract the pattern) is:

* i think there may be an interesting hybrid solution later on (but I would say neither of our solutions is fully generic as it is right now). But that's OK, we iterate!
* the way I am thinking about this is a bit different - I see Redux as the state manager (and thus the provider of state). So from that sense, if we are looking to abstract the pattern further, I would suggest a common Redux container pattern, where you are dispatching "next step" and "previous step" from your application Wizard, and letting a common Redux HOC connector handle the state logic (much like Redux form does w/ reset and validation). This would mean abstracting those next/prev callbacks into dispatches handled by a Wizard state manager HOC (and then letting it inject the active step index via Redux keys). That would be much more ideal to me b/c we get the added value of Redux logs/state management and it the state becomes consumable in any component via Redux keys (just like we have done w/ redux-form).
* As far as `disableNextStep`  goes - isn't that really specific to the consumer?  I guess we could wire that to Redux as well - but I took the approach of letting the consumer handle validation (but maybe it could be further abstracted as well).

In general - I think the `controlled()/getDervivedState/etc` pattern is good for simple UI component state in PF React - but i don't think the pattern scales as well to application state logic (but again, fine w/ your approach for the POC!). I guess the parallel analogy w/ Angular would be using a lot of $scope inside your Controllers or Directives... over time, i.m.o. this doesn't scale as well b/c you are having to move that $scope around on a per component level... v.s. just using a Provider or a Service to properly share that state. Redux does the same thing...but also gives us some really nice developer benefits (redux logger, redux time travel debugger, cross framework compatible, etc.)

does it make sense? I think we shelf this and revisit it after the POC! Hopefully we are a bit further in the "generator" territory soon and we can start exploring HOC Redux connected patterns (that would further abstract these common UI patterns).

mturley:
I totally agree with everything you said, and actually had exactly that in mind for later improvement. The reason I did it this way for now instead:

* I want to make iterating easier, and so I needed to get rid of the pattern we had-- which meant that state needed to go somewhere, and wiring it all the way up to redux seemed unnecessary for the POC, so I hit two birds with one stone by using controlled(), which will be nice for patternfly-react later. I fully expect to lift that `activeStateIndex` up to redux and pass it into `ModalWizard` as a prop, which will simply disable that state behavior. It also already has `onNext`, `onBack` and `onStepChanged` handlers, which get called at the same times as the existing setControlledState calls, so we can very easily wire it up to redux with no change to `ModalWizard`.
* I am fully on board with redux as a state manager, I'm just taking my temporary state code for the POC (saves a little bit of time versus building a reducer and action creator) and moving it somewhere where it won't be throwaway (controlled() in a component we can reuse). It's a step that gets things working now without wasting the effort of a one-off solution like `ModalWizardStateProvider` was.
* The thing about `disableNextStep` is, it is definitely specific to the consumer, but in order to determine it you need to know the current step and the form data. So I lifted `activeStateIndex` out of ModalWizard only as a callback argument, so that we could use it in the place where we already have access to form data, so that we don't have to put redux-form code in ModalWizard.  Of course, if we lift the `activeStepIndex` all the way up to redux, we won't need that callback, and we can just pass a prop for it.

I think we're on the same page, but I'm just trying to find clever ways to reuse code that is otherwise throwaway.

mturley:
Here you go @priley86 ... this is what I meant for us to really do with `ModalWizard`.  There's no need to duplicate any of that code... and look how nice it is to pass `onNext: this.createMappings` in the `steps` array.

I decided to flatten everything down into that steps array, which means no more `numSteps` prop (`steps.length` is fine), and no more passing `ModalWizardBody` as a child of `ModalWizard` (instead, we now pass its props directly to `ModalWizard`, and it is used internally in there only). This meant getting rid of the `MappingWizardBody` and `PlanWizardBody` components too...

...but don't worry, that's because everything that is not generic now fits in a nice small component, right in `MappingWizard` and `PlanWizard` themselves.

I think this simplifies things quite heavily.

mturley:
There we go. Should be ready now.

mturley:
@priley86 it is honestly easier to read the contents of the new `MappingWizard.js` and `PlanWizard.js` than to review the diff... i changed a lot.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
```diff
@@ -1,66 +1,40 @@
...
-    this.state = { activeStepIndex: 0 };
-    bindMethods(this, ['prevStep', 'nextStep', 'goToStep']);
-  }
-
-  prevStep() {
-    const { activeStepIndex } = this.state;
-    this.setState({ activeStepIndex: Math.max(activeStepIndex - 1, 0) });
+  constructor(props) {
+    super(props);
+    const register = (context, component) => {
```
priley86: if the idea is to make this a helper... i think it should go into helpers... also think we should name it something like `registerRegistryComponent` to indicate it is coming from the componentRegistry and we are binding it to `this`...

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
```diff
@@ -70,76 +44,55 @@ class MappingWizard extends React.Component {
...
-              {onFinalStep
-                ? __('Close')
-                : activeStepIndex === 3 ? __('Create') : __('Next')}
-              <Icon type="fa" name="angle-right" />
-            </Button>
-          </Modal.Footer>
-        </Wizard>
-      </Modal>
+        title={__('Infrastructure Mapping Wizard')}
+        shouldDisableNextStep={activeStepIndex => {
```
priley86: we now need to ensure this piece will work w/ our newest validations in Mapping Wizard (since this PR is conflating that piece too)... @michaelkro can you validate this?

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardSelectors.js
```diff
@@ -1,10 +1,13 @@
...
-});
+import { selectKeys } from 'patternfly-react';
 
-export const mappingWizardFormFilter = form => ({
-  mappingWizardGeneralStep: form.mappingWizardGeneralStep,
-  mappingWizardClustersStep: form.mappingWizardClustersStep,
-  mappingWizardDatastoresStep: form.mappingWizardDatastoresStep,
-  mappingWizardNetworksStep: form.mappingWizardNetworksStep
-});
+export const mappingWizardOverviewFilter = overview =>
```
priley86: i think it was mentioned in a recent Foreman PR (and in previous PF React Community meetings) that using a "memoized" selector is a performance boost. 
This library was mentioned (and currently evaluating that elsewhere):
https://github.com/reactjs/reselect

So for that reason... i don't think it makes sense to introduce `selectKeys`...but maybe a new memoized selector in pf-react later on... for now, can we just keep the current convention?

app/javascript/react/screens/App/Overview/screens/PlanWizard/__tests__/index.test.js
```diff
@@ -21,7 +21,11 @@ describe('Plan Wizard integration test', () => {
     createStore(
       combineReducers({ ...reducers }),
       {
-        overview: initialState
+        overview: initialState,
+        form: {
```
priley86: if you check latest, @michaelkro found a better fix for this last night...

app/javascript/react/screens/App/common/helpers.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
priley86: do we still need these empty "string" helpers?

package.json
```diff
@@ -55,7 +55,7 @@
     "cx": "^18.1.11",
     "intl": "~1.2.5",
     "patternfly": "^3.35.1",
-    "patternfly-react": "^1.9.2",
+    "patternfly-react": "^1.10.1",
```
priley86: thanks...

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
mturley: Yeah, I wasn't sure if we needed this, just wanted to save a few lines. Also, this code could be cleaner if we just called the registry right in the step render functions, is there a reason we need it to assign things to `this`? I assumed so when I wrote it this way.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
mturley: I didn't touch the validation code, and wired up the same form data to the same place in theory... Definitely should "just work", fingers crossed!

mturley: I didn't touch the validation code, and wired up the same form data to the same place in theory... Definitely should "just work", fingers crossed!

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardSelectors.js
mturley: Sure. Hadn't thought about the performance impact, I'll change this back.

app/javascript/react/screens/App/Overview/screens/PlanWizard/__tests__/index.test.js
mturley: Ok, I'll look!

app/javascript/react/screens/App/common/helpers.js
mturley: They are being used in only one place, and will be removed when Aparna finishes her stuff for plan wizard step 3. I moved them to helpers temporarily to declutter the PlanWizard code.

mturley:
Leaving this PR open for now, in case we find time to update and merge it after we finish all the existing work. Added a WIP flag and some relevant labels. If you guys think we should just close this, let me know.

mturley:
I'm closing this for now to get it out of the way. It'll be back when we don't have a deadline.

mturley:
This is now replaced by #160.

PR #71, opened 2018-02-27
Disable Next button on datastores and networks steps when there are no mappings

michaelkro:
* Uses redux-form-validators length helper to check that there is at
least one mapping in the array
* In order to leverage redux-form's validation capabilities, a Field
component must be present. So, switch from conditionally rendering
Field to hiding it when there is no selected source cluster


michaelkro:
fullscreen_2_27_18__4_23_pm

michaelkro:
fullscreen_2_27_18__4_25_pm

michaelkro:
fullscreen_2_27_18__4_26_pm

priley86:
thanks @michaelkro - this was one of the missing pieces yesterday. merging..

PR #72, opened 2018-02-27
Persist DualPaneMapper and TreeView when there are mappings

michaelkro:
* A redux-form decorated component is passed a "pristine" prop. This
is a boolean value that indicates whether the form data has changed
from its initialized values
* Also included: If there is only one source cluster being mapped from
step-2, auto-select it when entering the datastores step
* Handles both the datastores and networks steps


michaelkro:
## Before fix

michaelkro:
## After fix
after fix

PR #73, opened 2018-02-27
Persist CSV rows as objects with named fields instead of arrays

mturley:
Provides a single source of truth (the `columnNames` prop) for the format of the CSV files, taking the burden of that assumption away from the redux code. This way we can continue to merge the rest of our wizard code, and only have to come back and change `columnNames` when we know the final format of the CSV.


mturley:
From Bill Wei on gitter:

> For the card we just added, the columns to use will be: Name, Provider, Reference.

So I'm updating this PR to match those properties, `{ name, provider, reference }`.

priley86:
looks good to me, thanks! merging...

PR #74, opened 2018-02-28
Filter source clusters

michaelkro:
Original implementation was incorrect to remove target clusters from
list after creating a mapping. We want users to be able to continue
adding source clusters under a target cluster in an existing mapping.

Because we re-fetch all clusters when entering the clusters step, using
actions/reducers to manually add/remove clusters when
removing/adding mappings introduced a bug where it became possible to
"double map" a source cluster by navigating away from and back to the
clusters step.

* Removes all actions/reducers/constants for adding/removing source and
target clusters
* Use a filter instead to ensure that only unmapped source clusters
populate DualPaneMapper
* Modifies addMapping() to allow adding on to existing mappings
* Display loading spinners in Dual Pane Lists


michaelkro:
@priley86 doing 4ec1c21 in app/javascript/react/screens/App/Overview/screens/MappingWizard/__tests__/index.test.js
will get rid of that console error

priley86:
fantastic work @michaelkro ! much appreciated!

PR #83, opened 2018-03-01
integrated mappings apis

### LINKED ISSUE #78: Wire Mapping Wizard APIs into "real" APIs
* test the responses for clusters/networks/datastores and make sure they are valid
* test the responses for posting transformation_mappings
* change mock/dummy urls and remove the action creator exception responses (returning stubs now)

priley86:
closes #78 

I've tested the Mapping Wizard end-to-end against the backend Ruby APIs. This PR implements that behavior if you set `mockMode = false ` in all Actions and in `javascript/components/index.js` (where URLs are set). I wanted to be careful and ensure we preserve local development for future (so `mockMode = false` still runs w/ local stubs as well...

The other thing this does is fix an issue w/ not re-querying the transformation_mappings after a new one is created on Step 5 (continue to Plan). We needed to requery those again so that the newly constructed one would be present when continuing to Step 1 of Plan (previously had not done this).

So all that should really be left now is ensuring Plan Wizard APIs can be integrated, and that Status APIs can be integrated (but it looks like #82 pretty much covers this 😸  ).


AparnaKarve:
@priley86 I did a quick test around `mappingWizard.transformationsBody` since I had noticed some issues with regards to that before.
Noticed a couple of things -
- If I proceed to the last step by clicking on `5` (not the `Create` button), then I get this -
`All mappings in have been mapped.` (Infra mapping name missing here)

- However if I click on `Create`, then I see the mapping name appear in Step5. Although when I click `Close` (or Cancel), `mappingWizard.transformationsBody` is not cleared.

priley86:
Yes - so we should prevent the user from going to Step 5 before Step 4 is complete (I believe Mike R’s fix will cover this). I also think we should prevent user from going back to Step 4 after mapping has been constructed and we’ve proceeded to completion on Step 5 to avoid recreating the same mapping? Then it becomes additional logic to do an update vs a create - So I think we prevent from going back at that point.

priley86:
@michaelkro can you please rebase latest validations with this branch? I think we wrap this up pretty quickly...

Preventing the user from going back once we hit Step 5 is the last issue I think...and that should be very simple...

priley86:
FWIW doing an update flow later should be fairly straight forward too...

michaelkro:
@priley86 do you want me to add my commits here, or vice versa?

priley86:
closing as you've incorporated into #84 ...nice work!!!

PR #86, opened 2018-03-01
Add implementation of MigrationsCompletedCard

### LINKED ISSUE #11: Create the Migration Completed Cards
This issue will handle creating the UI components for the Migration in Completed Cards.

Subject to UX approval: 
These will be using the Aggregate Status pattern referenced here:
http://www.patternfly.org/pattern-library/cards/aggregate-status-card/

There is also a PF React issue here for Base Card :
https://github.com/patternfly/patternfly-react/issues/192

We will need to ensure the Base Card created supports Aggregate Status and then add any enhancements for Aggregate Status.

jeff-phillips-18:
Fixes #11 

image

Still need to test against backend.

@serenamarie125


jeff-phillips-18:
Updated to truncate name with tooltip:

image

PR #88, opened 2018-03-01
Plan Wizard Step3

AparnaKarve:
Current implementation includes a REST API POST call to create the Migration Plan.

Incorporating the second POST call to create the Migration Request is in progress...

screen shot 2018-03-01 at 7 21 15 am


AparnaKarve:
Un-WIPing the PR.
As discussed, the second POST call will be done in a separate PR.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -1,33 +1,151 @@
...
+class PlanWizard extends React.Component {
+  constructor() {
+    super();
+    this.state = { activeStepIndex: 0 };
+    bindMethods(this, ['prevStep', 'nextStep', 'goToStep']);
+  }
+
+  prevStep() {
+    const { activeStepIndex } = this.state;
+    this.setState({ activeStepIndex: Math.max(activeStepIndex - 1, 0) });
```
mturley: Duplicating all this state stuff in both Wizards makes me sad... but to @serenamarie125's point on Slack yesterday, we have bigger fish to fry. I'll save the cleanup for when there is time.

app/javascript/react/screens/App/Overview/screens/PlanWizard/helpers.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const createMigrationPlans = (
+  planWizardGeneralStep,
+  planWizardCSVStep
+) => {
+  const planName = planWizardGeneralStep.values.name;
+  const planDescription = planWizardGeneralStep.values.description;
+  const infrastructureMapping =
+    planWizardGeneralStep.values.infrastructure_mapping;
+  const vms = map(planWizardCSVStep.values.csvRows, 'reference').map(trim);
```
priley86: can someone yank lodash back out pretty please??? 

ES6 gives us map and trim natively now:

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_trim
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_map

priley86: good catch @jeff-phillips-18

AparnaKarve: @priley86 I will look into that.

In the previous versions of lodash, `map` used to be called `pluck`  - which is what I need here.

This link https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_map does not seem to be referring to the `map` (aka `pluck`) that I'm using here.
Maybe it goes by a different name in ES6?

PR #89, opened 2018-03-01
Fix top nav for mapping wizard

michaelkro:
Only allow users to use Wizard.Step to navigate
to steps they have already visited at least once

* We need to prevent users from skipping past the clusters step (step-2)
because if there are no cluster mappings, the dropdown selects in the
datastores (step-3) and networks (step-4) steps will not populate.
* When a redux-form element is mounted, the form's value key is
initialized with an empty object. So a Wizard.Step is disabled until
the user has visited that step at least once
* This forces the user to go through the mapping wizard sequentially
once, and afterwards, frees them to use the top nav to jump between all
of the steps, except step-5
* Entering step-5 automatically submits the infrastructure mapping, so
that Wizard.Step is permanently disabled. The user must hit the Create
button in step 4
* Changes the cursor to not-allowed when hovering over a disabled
Wizard.Step

## Note
The changes made to `ModalWizardBody` will not affect the PlanWizard

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
+++ b/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
@@ -17,9 +17,9 @@ class ModalWizardBody extends React.Component {
     bindMethods(this, ['onStepClick', 'stepProps', 'renderLoading']);
   }
 
-  onStepClick(stepIndex) {
+  onStepClick(stepIndex, disabled) {
     const { steps, goToStep, disableNextStep } = this.props;
-    if (disableNextStep) return;
+    if (disableNextStep || disabled) return;
     const step = steps[stepIndex];
     goToStep(stepIndex);
     if (step && step.onClick) {
@@ -76,7 +76,8 @@ class ModalWizardBody extends React.Component {
           steps={steps.map((stepObj, index) => (
             <Wizard.Step
               {...this.stepProps(index, stepObj.title)}
-              onClick={() => this.onStepClick(index)}
+              onClick={() => this.onStepClick(index, stepObj.disabled)}
+              className={stepObj.disabled && 'is-disabled'}
             />
           ))}
         />
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
@@ -126,8 +126,18 @@ class MappingWizard extends React.Component {
 
   goToStep(activeStepIndex) {
     const { activeStepIndex: currentStep } = this.state;
+    const targetStepProp = mappingWizardSteps[activeStepIndex];
+    const targetStepForm = this.props[targetStepProp];
 
-    if (currentStep !== 4) {
+    if (activeStepIndex === 4) {
+      this.setState(prevState => ({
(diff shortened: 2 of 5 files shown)


app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
```diff
@@ -126,8 +126,18 @@ class MappingWizard extends React.Component {
...
   goToStep(activeStepIndex) {
     const { activeStepIndex: currentStep } = this.state;
+    const targetStepProp = mappingWizardSteps[activeStepIndex];
+    const targetStepForm = this.props[targetStepProp];
 
-    if (currentStep !== 4) {
+    if (activeStepIndex === 4) {
+      this.setState(prevState => ({
+        activeStepIndex: prevState.activeStepIndex
+      }));
```
jeff-phillips-18: If clicking on this doesn't allow nav, please remove the pointer cursor from the step indicator for this step.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
```diff
@@ -126,8 +126,18 @@ class MappingWizard extends React.Component {
...
+    } else if (currentStep === 2 && activeStepIndex === 3) {
+      this.nextStep();
```
jeff-phillips-18: We really should be keeping track if the step's status is valid and not allow navigation forward based on the validity. The cursor over following steps in the step indicator should show disabled.

app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
```diff
@@ -17,9 +17,9 @@ class ModalWizardBody extends React.Component {
     bindMethods(this, ['onStepClick', 'stepProps', 'renderLoading']);
   }
 
-  onStepClick(stepIndex) {
+  onStepClick(stepIndex, disabled) {
     const { steps, goToStep, disableNextStep } = this.props;
-    if (disableNextStep) return;
+    if (disableNextStep || disabled) return;
```
mturley: I think you can pull in `this.props.activeStepIndex` here, and make this line into:
```
if (disableNextStep || stepIndex === activeStepIndex + 1 || disabled) return;
```

That way if the Next button works, the next number button also works.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardBody.js
```diff
@@ -35,27 +35,32 @@ class MappingWizardBody extends React.Component {
...
-            onClick: () => console.log('on step 2 click')
+            onClick: () => console.log('on step 2 click'),
+            disabled: !this.props.mappingWizardClustersStep.values
           },
           {
             title: __('Datastores'),
             render: () => this.mappingWizardDatastoresStepContainer,
-            onClick: () => console.log('on step 3 click')
+            onClick: () => console.log('on step 3 click'),
+            disabled: !this.props.mappingWizardDatastoresStep.values
```
mturley: Actually, I think this `disabled` property should be called `unopened` or `uninitialized` or `new` or something. Or it could even be called `visited` and be the reverse value.  I think it's fine to use the code you have, but let's call it what it is, `disabled` implies it's using the validation.

mturley:
My review comments above are fixed by #101.

PR #92, opened 2018-03-01
Display diskspace usage for datastores mappings

michaelkro:
Construct a string from
* Total size of target datastore
* Target datastore available space
* Sum used space over all mapped source datastores


priley86:
rebase please ;)

priley86:
so there is a locale setting for `numeral` (see Locales : http://numeraljs.com/) ...but I don't think we care if it's only being used for disk storage conversions..

priley86:
Right now... we are still using getText...but it might be nice to explore `FormattedNumbers` in the future too (still a back burner i18n issue it seems): https://github.com/yahoo/react-intl/wiki/Components#formattednumber

PR #101, opened 2018-03-02
Allow navigation to past steps even if the current step is incomplete.

mturley:
Minor cleanup from my comments on #89.

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
+++ b/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
@@ -22,7 +22,7 @@ class ModalWizardBody extends React.Component {
   onStepClick(stepIndex, disableGoto) {
     const { steps, goToStep, disableNextStep, activeStepIndex } = this.props;
     const nextStepClicked = stepIndex === activeStepIndex + 1;
-    if (disableNextStep || (!nextStepClicked && disableGoto)) return;
+    if (disableGoto || (nextStepClicked && disableNextStep)) return;
     const step = steps[stepIndex];
     goToStep(stepIndex);
     if (step && step.onClick) {
@@ -100,7 +100,7 @@ class ModalWizardBody extends React.Component {
             <Wizard.Step
               {...this.stepProps(index, stepObj.title)}
               onClick={() => this.onStepClick(index, stepObj.disableGoto)}
-              className={stepObj.disabled && 'is-disabled'}
+              className={stepObj.disableGoto && 'is-disabled'}
             />
           ))}
         />
--- a/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
+++ b/app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
@@ -19,9 +19,10 @@ class ModalWizardBody extends React.Component {
     bindMethods(this, ['onStepClick', 'stepProps', 'renderLoading']);
   }
 
-  onStepClick(stepIndex, disabled) {
-    const { steps, goToStep, disableNextStep } = this.props;
-    if (disableNextStep || disabled) return;
+  onStepClick(stepIndex, disableGoto) {
+    const { steps, goToStep, disableNextStep, activeStepIndex } = this.props;
+    const nextStepClicked = stepIndex === activeStepIndex + 1;
+    if (disableNextStep || (!nextStepClicked && disableGoto)) return;
     const step = steps[stepIndex];
     goToStep(stepIndex);
     if (step && step.onClick) {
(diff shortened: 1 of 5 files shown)


app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardBody.js
```diff
@@ -35,32 +35,27 @@ class MappingWizardBody extends React.Component {
           {
             title: __('General'),
             render: () => <MappingWizardGeneralStep />,
-            onClick: () => console.log('on step 1 click'),
-            disabled: !this.props.mappingWizardGeneralStep.values
+            disableGoto: !this.props.mappingWizardGeneralStep.values
```
priley86: i have no issue w/ this... mainly b/c `mappingWizardGeneralStep.values` is being inserted via parent props (and Redux) and localized to Mapping Wizard. There are no callbacks here, just redux-keyed props coming down the state tree. 😸

app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardBody.js
```diff
@@ -17,9 +17,10 @@ class ModalWizardBody extends React.Component {
     bindMethods(this, ['onStepClick', 'stepProps', 'renderLoading']);
   }
 
-  onStepClick(stepIndex, disabled) {
-    const { steps, goToStep, disableNextStep } = this.props;
-    if (disableNextStep || disabled) return;
+  onStepClick(stepIndex, disableGoto) {
+    const { steps, goToStep, disableNextStep, activeStepIndex } = this.props;
+    const nextStepClicked = stepIndex === activeStepIndex + 1;
+    if (disableNextStep || (!nextStepClicked && disableGoto)) return;
```
priley86: I don't know if this conditional above is safe for all consumers, but it seems safe for our Wizards now. I would think that `onStepClick` can be an action injected into `MappingWizard` (via prop how we inject actions today) in the future along with `activeStepIndex`, etc. and the consumer can determine how to handle this. If we want to make the assumption for them, it should at least be overridable.

This is OK for now though! ;)

priley86: so long term...

we end up seeing something like this later on:
```
const MappingWizardContainer = compose(WizardContainer)(MappingWizard);
export default connect(mapStateToProps, MappingWizardActions, mergeProps)(
  MappingWizardContainer
);

priley86: It is also probably still fine to render a `ModalWizard` inside of `MappingWizard` that is stateless and just renders those common child elements...

mturley: That sounds good! I am all for the HoC pattern.

As for this conditional not being safe for all consumers, I actually thought about that and i was originally going to have the boolean be `visited`, and flip it, to be more explicit. But that would have been a problem, because the consumer would have to know to put a "visited: true" property on their steps to enable those buttons.  Keeping it flipped means that if you simply don't define a `disableGoto` property, it's the same as having `disableGoto: false` and it retains the functionality. Seems safe enough to me.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardBody.js
AparnaKarve: `!this.props.mappingWizardGeneralStep.values` condition won't be enough to block a GoTo

`mappingWizardGeneralStep.syncErrors` is what decides if the current form is valid or not

michaelkro:
Oops, sorry, on second inspection I am seeing some issues here.  My screenshots don't show my cursor, so can't provide a picture, but the `not-allowed` cursor isn't working anymore

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardBody.js
mturley: Right, but we want to allow the user to hop to/from forms that are invalid/empty/incomplete, as long as they have been visited already (and values gets initialized on first visit). We just don't want users getting to a step for the first time via any other means than the Next button (which does use that validation).

mturley:
@michaelkro the issue you found with being able to skip from step 4 to step 5 was because i wanted to have it allow the user to use the number instead of the Next button if they want. (a), in order to trigger the same request that is on the Next button, that would have depended on an abstraction i forgot we didn't have because we didn't merge my refactor, and (b) it's dumb anyway, Next button should be all you can use to proceed. So that's how it works now.

priley86:
i think we can review this alongside #109  today...

priley86:
since this only applies to step clicks, it should be safe to merge this for now. We are planning to disable "step clicks" **for the POC** in #121  and then revisit this abstraction for Wizards here early next sprint. I agree w/ @mturley that we should revisit the abstraction as soon we know the final intended UX...

PR #112, opened 2018-03-02
update migrations-in-progress url

priley86:
really funny 😸 

@AparnaKarve can you confirm these URLs for us please?


priley86:
also - may need to review response data... 

screen shot 2018-03-02 at 5 10 12 pm

priley86:
another update:

`"filter[]=state='active'"` will never be visible locally... (b/c the backend will take it from `pending` to `active` for us, but this won't happen on our machines). May need to investigate further next week...

priley86:
I think the best option may be to show another row of cards for `pending` status - per Aparna - a background queue will change the request to `active` ... so I believe we should poll the status of migrations every few seconds and update the Overview dashboard appropriately.

AparnaKarve:
LGTM -- Merging

PR #113, opened 2018-03-05
Add target cluster name to drop downs

michaelkro:
Display the associated target cluster's name alongside each source
cluster in the dropdown selects

closes https://github.com/priley86/miq_v2v_ui_plugin/issues/103

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/SourceClusterSelect/SourceClusterSelect.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/SourceClusterSelect/SourceClusterSelect.js
@@ -14,11 +14,27 @@ import {
 
 const SourceClusterSelect = ({
   form,
-  sourceClusters,
+  clusterMappings,
   selectSourceCluster,
-  selectedCluster
+  selectedCluster,
+  selectedClusterMapping
 }) => {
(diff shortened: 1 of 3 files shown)


michaelkro:
fullscreen_3_5_18__9_13_am

michaelkro:
fullscreen_3_5_18__9_11_am

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/SourceClusterSelect/SourceClusterSelect.js
```diff
@@ -14,12 +14,28 @@ import {
...
+    (mappings, targetClusterWithSourceClusters) => {
+      const {
+        nodes: sourceClusters,
+        ...targetCluster
+      } = targetClusterWithSourceClusters;
+      const sourceToTargetMappings = sourceClusters.map(sourceCluster => ({
+        sourceCluster,
+        targetCluster
+      }));
+      return mappings.concat(sourceToTargetMappings);
```
mturley: For future reference you can also use spread syntax (`...`) in arrays instead of `.concat()`. We may as well use it wherever we can if we're using it.

```
return [...mappings, ...sourceToTargetMappings];
```

I'll approve this one anyway because it's perfectly readable as is and it's more of a preference thing anyway.

PR #114, opened 2018-03-05
introduce globalMock setting and polling for active/completed migrations

priley86:
* introduce a `globalMock` setting for globally toggling all mock modes
* poll the active/completed migrations every 10 seconds. Ensure the spinner animation lasts 3 seconds (and doesn't go away immediately).

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Cards/MigrationsInProgressCard/MigrationsInProgressCard.js
+++ b/app/javascript/react/screens/App/Overview/components/Cards/MigrationsInProgressCard/MigrationsInProgressCard.js
@@ -14,7 +14,7 @@ class MigrationsInProgressCard extends React.Component {
     const { fetchMigrationsInProgressAction } = this.props;
     // fetch migrations in progress initially, then poll them
     fetchMigrationsInProgressAction();
-    setInterval(fetchMigrationsInProgressAction, 10000);
+    this.pollingInterval = setInterval(fetchMigrationsInProgressAction, 10000);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -30,6 +30,10 @@ class MigrationsInProgressCard extends React.Component {
         });
       }, 3000);
     }
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.pollingInterval);
   }
 
   renderActiveMigrations() {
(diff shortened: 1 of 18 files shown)


app/javascript/react/screens/App/Overview/components/Cards/MigrationsInProgressCard/MigrationsInProgressCard.js
```diff
@@ -4,9 +4,36 @@ import { Grid, Icon, Spinner } from 'patternfly-react';
...
+      });
+    } else {
+      setTimeout(() => {
+        this.setState({
+          isFetchingMigrationsInProgress: false
+        });
+      }, 3000);
+    }
+  }
+
```
jeff-phillips-18: I think it will annoy users if we are constantly showing the wait cursor on each update. I don't believe we should change the cursor during auto-updates.

priley86: good point - we can restrict this to **not** poll when a modal is open...does that work @jeff-phillips-18 ?

priley86: (basically passing down the `mappingWizardVisible` and `planWizardVisible` state from Overview...)

priley86: will update this now... give me 10 minutes ;)

AparnaKarve: @priley86 Should the auto-updates be done silently - even when the modals are not visible?

i.e. no visual indication to the user about refreshing the migration status. (as in no spinners)

priley86: i've made a change to prevent polling while modal wizards are open. LMK if this OK.

AparnaKarve:
@priley86 If the polling is happening every 3 seconds, I'm not seeing any API related state changes (PENDING, FULFILLED etc)

AparnaKarve:
@priley86 Probably some environment related issues at my end earlier.
Everything is working now, as expected.

Merging...

PR #128, opened 2018-03-07
Unhappy errors

### LINKED ISSUE #109: Mapping Wizard - "unhappy" error APIs
Steps 2 - 4 need error messages for when APIs return errors instead of clusters/networks/datastores.

priley86:
Closes #109 

Step 2: 
Display the error message centrally and wire up the "retry" button to re-fetch clusters. Next button is disabled (as no clusters can be mapped).

screen shot 2018-03-07 at 2 23 33 pm

Step 3:
Display the error message in an alert. On load, we will always be selecting the first source cluster, and if it fails, display the alert. If source cluster is selected again and fails or any other source cluster fails, update the alert. Alert can be closed at any point while mappings are still in display.

**Update**: since these are really two API calls to retrieve data stores for source and data stores for target, update the messaging to reflect this. Also - adding some adequate spacing to distinguish the alert from the wizard.

Selected Source Cluster Fails to load Datastores:
screen shot 2018-03-07 at 2 16 20 pm

Selected Target Cluster Fails to load Datastores:
screen shot 2018-03-07 at 2 14 38 pm
 

Step 4:
Selected Source Cluster Fails to load Networks:

screen shot 2018-03-07 at 2 19 18 pm

Selected Target Cluster Fails to load Networks:
screen shot 2018-03-07 at 2 20 02 pm

Adding a retry button to the alert in steps 3/4 is a bit of stretch (have to wire up more actions for that). Leaving this alone for now.

@michaelkro can you sanity check this?

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
@@ -1,13 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { noop } from 'patternfly-react';
+import { Button, noop, bindMethods } from 'patternfly-react';
 import { Field, reduxForm } from 'redux-form';
 import { length } from 'redux-form-validators';
 
 import ClustersStepForm from './components/ClustersStepForm';
 
 class MappingWizardClustersStep extends React.Component {
+  constructor(props) {
+    super(props);
+    bindMethods(this, ['fetchClusters']);
+  }
   componentDidMount() {
+    this.fetchClusters();
+  }
+
+  fetchClusters() {
     const {
       fetchSourceClustersUrl,
       fetchSourceClustersAction,
@@ -24,9 +32,29 @@ class MappingWizardClustersStep extends React.Component {
       isFetchingSourceClusters,
       sourceClusters,
       isFetchingTargetClusters,
-      targetClusters
+      targetClusters,
+      isRejectedSourceClusters,
+      isRejectedTargetClusters
     } = this.props;
 
+    if (isRejectedSourceClusters || isRejectedTargetClusters) {
+      return (
+        <div className="wizard-pf-complete blank-slate-pf">
+          <div className="wizard-pf-success-icon">
+            <span className="pficon pficon-error-circle-o" />
+          </div>
+          <h3 className="blank-slate-pf-main-action">
+            Error Retreiving Clusters
+          </h3>
+          <p className="blank-slate-pf-secondary-action">
+            We&apos;re sorry, something went wrong. Please try again.
+          </p>
+          <Button bsStyle="primary" onClick={this.fetchClusters}>
+            Retry
+          </Button>
+        </div>
+      );
+    }
     return (
       <Field
         name="clusterMappings"
@@ -49,7 +77,9 @@ MappingWizardClustersStep.propTypes = {
   sourceClusters: PropTypes.arrayOf(PropTypes.object),
   targetClusters: PropTypes.arrayOf(PropTypes.object),
   isFetchingSourceClusters: PropTypes.bool,
-  isFetchingTargetClusters: PropTypes.bool
+  isFetchingTargetClusters: PropTypes.bool,
+  isRejectedSourceClusters: PropTypes.bool,
+  isRejectedTargetClusters: PropTypes.bool
 };
 MappingWizardClustersStep.defaultProps = {
   fetchSourceClustersUrl: '',
@@ -57,7 +87,9 @@ MappingWizardClustersStep.defaultProps = {
   fetchTargetClustersUrl: '',
   fetchTargetClustersAction: noop,
   isFetchingSourceClusters: true,
-  isFetchingTargetClusters: true
+  isFetchingTargetClusters: true,
+  isRejectedSourceClusters: false,
+  isRejectedTargetClusters: false
 };
 
 export default reduxForm({
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
@@ -44,13 +44,13 @@ class MappingWizardClustersStep extends React.Component {
             <span className="pficon pficon-error-circle-o" />
           </div>
           <h3 className="blank-slate-pf-main-action">
-            Error Retreiving Clusters
+            {__('Error Retrieving Clusters')}
           </h3>
           <p className="blank-slate-pf-secondary-action">
-            We&apos;re sorry, something went wrong. Please try again.
+            {__("We're sorry, something went wrong. Please try again.")}
           </p>
           <Button bsStyle="primary" onClick={this.fetchClusters}>
-            Retry
+            {__('Retry')}
           </Button>
         </div>
       );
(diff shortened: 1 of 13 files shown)


michaelkro:
If there is only one source cluster mapped in step-2, we auto select it in steps 3 & 4.  Looks like that behavior is bypassing error validation

fullscreen_3_7_18__7_54_am

michaelkro:
So, it looks like the auto selection has nothing to do with the screen I posted above.  I'm testing by just going into the `mockMode` blocks and removing the `catch` blocks.  Maybe this isn't the right way to be testing these steps?

michaelkro:
Ok, my bad.  I was accidentally removing the `return reject()`

All is well!

priley86:
I was testing by changing the "real" source/target cluster urls to "dummy" urls... that way it fires REJECTED. Removing the catch will actually break it? I believe changing the URLs is the way to test that...

michaelkro:
Ok, if I remove the `mockMode` block, leaving the `return reject()`, this is the screen I see.  If this is not a valid test case, or if we don't mind the spinner, then we can ignore!

fullscreen_3_7_18__8_03_am

priley86:
fixed...

>Oops, forgot to set `isRejectedTargetClusters` to false here [clusters step reducer]( https://github.com/priley86/miq_v2v_ui_plugin/blob/unhappy-errors/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStepReducer.js#L57)

michaelkro:
If I error out fetching target datastores, but not source datastores, I see this

fullscreen_3_7_18__8_11_am

priley86:
that's a good point - if we fail fetching source clusters, but not target clusters...what should we display @serenamarie125 @vconzola ? Right now, I am lumping them together to match the selection, however when each source cluster is selected, we actually query both...

vconzola:
@priley86 @michaelkro I'd expect the error message to specify the failing cluster name and what can not be fetched. For example, "Error retrieving cluster XYZ." if we can't get to the cluster itself, or "Error retrieving datastores/networks for cluster XYZ." if we can't retrieve the cluster's datastores or networks. I don't think it would matter if it's source or target clusters. If both fail, I'd display the error for the source. Does this answer your question?

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
```diff
@@ -24,9 +32,29 @@ class MappingWizardClustersStep extends React.Component {
...
     } = this.props;
 
+    if (isRejectedSourceClusters || isRejectedTargetClusters) {
+      return (
+        <div className="wizard-pf-complete blank-slate-pf">
+          <div className="wizard-pf-success-icon">
+            <span className="pficon pficon-error-circle-o" />
+          </div>
+          <h3 className="blank-slate-pf-main-action">
+            Error Retreiving Clusters
```
michaelkro: typo, should be `Retrieving`

priley86:
ok - updated screenshots just now... apologies for just now getting to this ;)

michaelkro:
For failed target datastore fetching, the text does not seem to be updating.  Works fine when it's the source datastore retrieval that has failed

fullscreen_3_7_18__2_35_pm

michaelkro:
Just async things.  Not sure if we care about this, but figured it was worth pointing out

vconzola:
@Mike Ro The cluster in the message (RHV Cluster1) matches the target in
the dropdown, so not sure what you're saying. But the fact that it's
unclear to you tells me we need to change the message. Would "Error
retrieving datastores for cluster RHV Cluster1" be more clear?  Do you have
suggestions that would make the message more clear?

VINCE CONZOLA

SR. INTERACTION DESIGNER, USER EXPERIENCE DESIGN

Red Hat <https://www.redhat.com/>

vconzola@redhat.com    T: 9197165041
<https://red.ht/sig>

michaelkro:
@vconzola, it's not the message, but rather that if both requests are failing, the message changes as each request fails.

It's not a big deal, but when it's not slowed down like my gif, the message flickers as it changes.

priley86:
@michaelkro can you try again? think it should be fixed...

michaelkro:
Everything looks good to me!

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
```diff
@@ -33,6 +33,40 @@ class MappingWizardDatastoresStep extends React.Component {
...
+    const { selectedCluster, selectedClusterMapping } = this.state;
+    if (
+      isRejectedSourceDatastores !== nextProps.isRejectedSourceDatastores &&
+      nextProps.isRejectedSourceDatastores
+    ) {
+      showAlertAction(
+        __(
+          `Error retrieving cluster datastores: ${selectedCluster.name}, ID: ${
+            selectedCluster.id
+          }`
```
michaelkro: I think we're supposed to be using `sprintf` to translate dynamic strings

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
```diff
@@ -33,6 +33,40 @@ class MappingWizardDatastoresStep extends React.Component {
...
+    } else if (
+      isRejectedTargetDatastores !== nextProps.isRejectedTargetDatastores &&
+      nextProps.isRejectedTargetDatastores &&
+      !isRejectedSourceDatastores
+    ) {
+      showAlertAction(
+        __(
+          `Error retrieving cluster datastores: ${
+            selectedClusterMapping.name
+          }, ID: ${selectedClusterMapping.id}`
```
michaelkro: ditto

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/MappingWizardNetworksStep.js
```diff
@@ -33,6 +33,41 @@ class MappingWizardNetworksStep extends React.Component {
...
+
+    if (
+      isRejectedSourceNetworks !== nextProps.isRejectedSourceNetworks &&
+      nextProps.isRejectedSourceNetworks
+    ) {
+      showAlertAction(
+        __(
+          `Error retrieving cluster networks: ${selectedCluster.name}, ID: ${
+            selectedCluster.id
+          }`
```
michaelkro: ditto

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/MappingWizardNetworksStep.js
```diff
@@ -33,6 +33,41 @@ class MappingWizardNetworksStep extends React.Component {
...
+      isRejectedTargetNetworks !== nextProps.isRejectedTargetNetworks &&
+      nextProps.isRejectedTargetNetworks &&
+      !isRejectedSourceNetworks
+    ) {
+      showAlertAction(
+        __(
+          `Error retrieving cluster networks: ${
+            selectedClusterMapping.name
+          }, ID: ${selectedClusterMapping.id}`
+        )
```
michaelkro: ditto

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
priley86: interesting... not really sure what the distinction is there yet...

priley86: will update...

priley86:
* updated tagged template literals to `sprintf`. Hadn't noticed those until now (and was still in `react-intl` land). `sprintf` reminded me of writing C programs on Lexmark printers as an intern about 13 years ago... just thankful i'm not doing pointer arithmetic anymore 🤣 

thanks very much for the through reviews @michaelkro ! 

ok to merge?

PR #139, opened 2018-03-09
Validate infrastructure mapping name

michaelkro:
* Use redux-form async validation to check whether the currently entered
name is a duplicate
* If a duplicate is found, show alert, otherwise move forward to next
wizard step
* Automatically dismiss alert when user changes name

Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/134

### Note
Was seeing the same snapshot diffs on `migration-poc`, so went ahead and updated them


michaelkro:
* Show errors after hitting `next` button
* Automatically dismiss alert if user changes name

michaelkro:
* Show errors if the user clicks away from the name field after having entered a duplicate name

app/javascript/react/screens/App/Overview/__test__/__snapshots__/Overview.test.js.snap
```diff
@@ -13,16 +13,26 @@ exports[`Overview component renders the overview 1`] = `
     }
   >
     <Col
+      bsClass="col"
```
priley86: this is odd... i cleared my `node_modules` and reinstalled a few days ago... and it was changing these snaps...and now they are updating again (guessing we had different dependencies installed). Would anyone be opposed to reintroducing the package lock file? Kind of annoying...

priley86: The semantic updates to various packages can also cause these...

priley86: sorry...`yarn.lock` can be added back to prevent this...

PR #140, opened 2018-03-09
Custom dropdown for Infrastructure Mappings in Plan Wizard Step1

AparnaKarve:
Custom dropdown for Infrastructure Mappings in Plan Wizard Step1

screen shot 2018-03-09 at 11 09 50 am

screen shot 2018-03-09 at 11 12 45 am

screen shot 2018-03-09 at 11 13 27 am

Fixes https://github.com/priley86/miq_v2v_ui_plugin/issues/107

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.scss
+++ b/app/javascript/react/screens/App/Overview/Overview.scss
@@ -1,4 +1,5 @@
 @import './screens/MappingWizard/MappingWizard.scss';
+@import './screens/PlanWizard/components/DropdownButtonSelect/DropdownButtonSelect.scss';
 @import './screens/PlanWizard/components/PlanWizardResultsStep/PlanWizardResultsStep.scss';
 @import 'components/Cards/MigrationsInProgressCard/MigrationsInProgress.scss';
 @import 'components/Cards/MigrationsCompletedCard/MigrationsCompleted.scss';
--- a/app/javascript/react/screens/App/Overview/__test__/__snapshots__/Overview.test.js.snap
+++ b/app/javascript/react/screens/App/Overview/__test__/__snapshots__/Overview.test.js.snap
@@ -13,16 +13,26 @@ exports[`Overview component renders the overview 1`] = `
     }
   >
     <Col
+      bsClass="col"
+      componentClass="div"
       md={12}
     >
-      <Row>
+      <Row
+        bsClass="row"
+        componentClass="div"
+      >
         <div
           className="spacer"
         />
         <Equalizer
+          byRow={true}
+          enabled={[Function]}
           nodes={[Function]}
+          property="height"
         >
           <Col
+            bsClass="col"
+            componentClass="div"
             md={6}
             xs={12}
           >
@@ -32,6 +42,8 @@ exports[`Overview component renders the overview 1`] = `
             />
           </Col>
           <Col
+            bsClass="col"
+            componentClass="div"
             md={6}
             xs={12}
           >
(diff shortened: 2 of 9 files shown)


priley86:
hey @AparnaKarve - thanks a lot for following up on this. This second screenshot though just looks weird to me (w/ the dropdown not going 100% like the native one does). 

@serenamarie125 @jeff-phillips-18 do you have any thoughts on how to handle this? I've asked Jenn G. here in Raleigh about styling the dropdowns for future, and it sounds like we do not yet have a standard (but will being trying to use better aria tags/semantic `select` elements to ensure accessibility). If that is the case, maybe we should just style the `DropdownButton` select option to be 100% width of the select element (but not sure how difficult this would be yet).

AparnaKarve:
@michaelkro I was thinking we could still keep that native select in there, just for reference purposes like @priley86 did above.

priley86:
@AparnaKarve would you mind seeing if we can match the [existing](https://silviomoreto.github.io/bootstrap-select/examples/) bootstrap-select "search" feature in use in MIQ today with this library: https://www.npmjs.com/package/react-bootstrap-select?

AparnaKarve:
@priley86 It looks like `react-bootstrap-select` is using an older version of React.

The Select.jsx code is calling `React.createClass` which is supported only in the older versions of React (I think in React version 14 or below) -- 
https://github.com/tjwebb/react-bootstrap-select/blob/master/jsx/Select.jsx#L9
As a result, I'm getting an error stating that `React.createClass` is an undefined function. 
Unless there is a workaround for this, `react-bootstrap-select` is pretty much unusable at the moment.


There is an open issue for this repo, which describes the above version problem -
https://github.com/tjwebb/react-bootstrap-select/issues/26

priley86:
yikes! Well that means we have some old dependencies, don't we ;)

At this point, I recommend just consuming `bootstrap-select` jquery version (also being consumed by Angular in MIQ), and just slightly tweak the React Wrapper there... 
https://github.com/tjwebb/react-bootstrap-select/blob/master/jsx/Select.jsx

It would look like this:
```
import React from 'react';
import '../path-to/bootstrap-select';
import {FormControl} from 'patternfly-react';  //i noted that it's using react-bootstrap just like us

class BootstrapSelect extends React.Component {
  getInitialState: function () {
    return {
      open: false
    };
  }
componentDidUpdate: function () {
    $(ReactDOM.findDOMNode(this)).find('select').selectpicker('refresh');
    var select = $(ReactDOM.findDOMNode(this)).find('div.bootstrap-select');
    select.toggleClass('open', this.state.open);
  }
componentWillUnmount: function () {
    var self = this;
    var select = $(ReactDOM.findDOMNode(this)).find('select');

    var button = $(ReactDOM.findDOMNode(this)).find('button');
    var dropdown = $(ReactDOM.findDOMNode(this)).find('.dropdown-menu.open');
    var items = $(ReactDOM.findDOMNode(this)).find('ul.dropdown-menu li a');

    $('html').off('click');
    button.off('click');
    items.off('click');
  }
  componentDidMount: function () {
    var self = this;
    var select = $(ReactDOM.findDOMNode(this)).find('select');
    $(select).selectpicker();

    var button = $(ReactDOM.findDOMNode(this)).find('button');
    var dropdown = $(ReactDOM.findDOMNode(this)).find('.dropdown-menu.open');
    var items = $(ReactDOM.findDOMNode(this)).find('ul.dropdown-menu li a');

    $('html').click(function () {
      self.setState({ open: false });
    });

    button.click(function (e) {
      e.stopPropagation();
      self.setState({ open: !self.state.open });
    });

    items.click(function () {
      if (self.props.multiple) return;
      self.setState({ open: !self.state.open });
    });
  }
  render: function () {
    return (
      <Input {...this.props} componentClass='select' />
    );
  }
}
```

https://daveceddia.com/convert-createclass-to-es6-class/

You are going to eslint ignore the `findDomNode` warnings as well...

I still think this is the easiest path - let me know if you need any more help w/ this...

priley86:
after further discussion... here is the two current options currently in use today. 

**Foreman** (uses Select2):

select2

You can read more about the appearance options here:
https://select2.org/appearance

I can also dig up some more examples w/ Roxanne...


**MIQ** (existing one noted above) uses jquery bootstrap select/ angular pf-select with `data-live-search` option. It looks like this:

pf-select

You can also see the options for bootstrap select here (under Live Search):
https://silviomoreto.github.io/bootstrap-select/examples/

These are the current dropdowns in use today. We could also compare w/ Open Shift in the future I'm guessing...

priley86:
closing this one...since we are resolving it in #148

PR #149, opened 2018-03-16
Group source networks with multiple switches

michaelkro:
* Organize source networks into groupings
  * Membership is based on the network's uid_ems
  * Stored as arrays on an object whose keys consist of each group's
  uid_ems
* Only display one representative of each group when listing source
networks in DualPaneMapper
* When adding a new networks mapping, for each source network selected,
map its entire grouping
* Filter networks step mappings when displaying in TreeView so that
only one representative of each mapped source network grouping is shown
* Use uid_ems to identify source networks, so that it does not matter
which particular group member happens to be displayed in the UI
* Update fixture data so that the uid_ems and name attributes are the
same for source networks (this is how it is in the test db)

# Notes
* It seems that the `uid_ems` and `name` attributes only match for source networks, and not for target networks in the test db, so I did not make any changes as to how we deal with target networks

### Target Network
fullscreen_3_16_18__2_00_pm

### Source Network
fullscreen_3_16_18__2_03_pm


AparnaKarve:
> It seems that the uid_ems and name attributes only match for source networks, and not for target networks in the test db

@michaelkro That is interesting. I'll check with the backend folks if that is expected.

michaelkro:
Thanks @AparnaKarve :)

I actually ended up still using the `name` attr for displaying source networks, since it looks like it always matches the `uid_ems`.  So, unless we will have to group target networks, I don't think it matters one way or another...

I figured it was worth mentioning here, just in case

AparnaKarve:
> I actually ended up still using the name attr for displaying source networks, since it looks like it always matches the uid_ems

True -- will double-check with Bill once again.
(IIRC, he mentioned that `uid_ems` is guaranteed to be unique)

I will review/test this sometime today & Monday

Thanks for the PR @michaelkro !

priley86:
great work @michaelkro ! ...  ❤️the enzyme tests started here ;)

AparnaKarve:
@michaelkro I have verified that the PR  -
- deduplicates the source Network list
- Maps all the source networks to the target networks

Still waiting on the information on `uid_ems` Vs. `name`

PR #158, opened 2018-03-20
Refactor `removeNode()`

michaelkro:
* Break down the method and improve readability by extracting pieces into helper functions
* Tests should also help support changes introduced in https://github.com/priley86/miq_v2v_ui_plugin/pull/149, specifically, by ensuring that when we remove a source network node from the TreeView, that it's entire grouping gets removed from the data mapping

# Note
First commit is to fix some tests that were passing due to collision of non-unique ids

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/components/NetworksStepForm/NetworksStepForm.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/components/NetworksStepForm/NetworksStepForm.js
@@ -15,8 +15,8 @@ import {
   targetNetworkWithTreeViewAttrs,
   networkGroupingForRep,
   mappingsForTreeView,
-  removeTargetNetwork,
-  removeSourceNetwork
+  mappingWithTargetNetworkRemoved,
+  mappingWithSourceNetworkRemoved
 } from './helpers';
 
 class NetworksStepForm extends React.Component {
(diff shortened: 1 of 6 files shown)


app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/components/NetworksStepForm/NetworksStepForm.js
```diff
@@ -285,60 +287,46 @@ class NetworksStepForm extends React.Component {
...
+    // --> Source network grouping(s)
+
+    const updatedMappings = isTargetNetwork
+      ? networksStepMappings
+          .map(networksStepMapping =>
+            removeTargetNetwork(networksStepMapping, selectedNode)
+          )
+          .filter(item => item !== null)
+      : networksStepMappings
+          .map(networksStepMapping => {
```
ohadlevy: while it can be premature optimization, it feels like we are looping over the mapping many many times?

michaelkro: Hi @ohadlevy, thanks for taking a look!

Are you referring to the fact that I am `map`ping, and then following that with `filter`?  I will clean this up

priley86:
the nested .reduce() seems cleaner than .map() .filter() ... as it's a re-representation of the network nodes after their removal from the treeview.

this approach works for me 👍

PR #176, opened 2018-04-03
Prep for Sprint6 Overview

michaelkro:
* Stub Aggregate Info Cards
* Fetch all transformation plan requests
  * Will use selectors in future PRs to separate into: pending, in progress, complete
  * Set up `mockMode` and add initial mock data
* Set up polling
  * Polls back end every 15 seconds for all transformation plan requests
* Fetch all transformation plans
# Notes
* Leaves old Overview Card components for reference, to be removed later
* Mock data is based on [example json](https://gist.github.com/bzwei/3213174ee5efb4e377d789df8394cf2b#file-v2v_example-json-L465) and is subject to change
* API endpoint for fetching transformation plans [(example API usage)](https://gist.github.com/bzwei/3213174ee5efb4e377d789df8394cf2b#file-v2v_example-json-L465) currently fails with the following error
  ```json
  {
    "error": {
      "kind": "bad_request",
      "message": "Invalid attributes specified: miq_requests",
      "klass": "Api::BadRequestError"
    }
  }
  ```
  I'm guessing this is still in development, so we can use `mockMode` for the time being


priley86:
can we go ahead and delete react-equalizer w/ this? This is looking good... just have to figure out the UX now ... haha 😆

priley86:
still need to finalize on the UX for aggregate cards - hopefully we can do that today. Bill updates the query to retrieve all plans / tasks yesterday (it looks like you have the right one):
https://gist.github.com/bzwei/3213174ee5efb4e377d789df8394cf2b#file-v2v_example-json-L751

"Query all active requests including tasks"...

I think you can iterate the response and start writing the logic  (reducers) to drive the aggregate counts. Let me know if you need anything from me ;)

Also - the icon selections look rather obtuse...do we have specific ones in mind for aggregate @vconzola @serenamarie125 ?

michaelkro:
Ya, the mockup documents show placeholder icons which don't actually exist in PF, so I just picked some as placeholders for now ¯\_(ツ)_/¯

Sorry, should have specified that in the PR notes :)

vconzola:
@michaelkro After reviewing with Jenny Haines there isn't really a good family of icons to represent the different migration plan states, so we decided we should remove the object icons from the agg status cards altogether. We'll still show the status icons as appropriate, just not the object icons to the left of the count and object type.

michaelkro:
These commits have been included in https://github.com/priley86/miq_v2v_ui_plugin/pull/177

PR #190, opened 2018-04-09
Create new advanced settings flag to enable/disable v2v migration in the Ops UI

AparnaKarve:
The Advanced Settings flag is `transformation` which should be placed directly under the `product` section in the yaml as shown below -

screen shot 2018-04-09 at 1 35 22 pm

----------

To enable v2v UI --
```yaml
:transformation: true
```
--------

To disable v2v UI --
```yaml
:transformation: false
```

or

Do not specify the`:transformation` flag in the yaml

--------

https://bugzilla.redhat.com/show_bug.cgi?id=1565328

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,7 @@
 # MiqV2vUI::Engine.routes.draw do
-Rails.application.routes.draw do
-  match '/migration' => 'migration#index', :via => [:get]
-  match 'migration/*page' => 'migration#index', :via => [:get]
+if ::Settings.product.transformation == true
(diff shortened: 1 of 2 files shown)


priley86:
tested both scenarios locally ... setting `:transformation: true` and `:transformation: false` behaves appropriately. This is good to go.

priley86:
For anyone else testing this in the future...the steps are to go to Administrator (top right) -> Configuration - > Advanced tab. Scroll down to the `:product` section and add the `:transformation: true` value. Then restart the manageiq server.

PR #191, opened 2018-04-10
Upgrade PF react to newest version so we can use it's new goodies

karelhala:
Since there were quite a few improvements in PF react since it's older version we should update it.

### Update package.json
* `pf-react-1.11.0` -> `pf-react-1.16.0`


karelhala:
@himdel ping
@priley86 we need this to be merged, because we are using pf-react for taggint https://github.com/ManageIQ/react-ui-components/pull/3 and this new component is using label remove button, but in current version (`1.11.0`) this feature is missing. And due to settings of our build in `ui-classic` our build proccess takes this old version even if we update it in our package. We are working on fixing this issue, but this is nevertheless easiest and probably fastest solution for us.

himdel:
LGTM :+1: waiting for any input @priley86 might have :)

Yeah, we'll have to define a set of common packages that are shared by default to prevent these kinds of incompatibilities.

AllenBW:
OOooooo so right now v2v is working off the sprint 6 branch, which is already at 16.4 https://github.com/priley86/miq_v2v_ui_plugin/blob/sprint6/package.json#L67

himdel:
Good, so do we merge this so that we don't have to wait for the end of the sprint?
Or is the end of the sprint today-ish? :)

priley86:
So a couple things here... 

1. You will need to make sure yarn.lock references a `1.16.X` version. I recommend changing to `"patternfly-react": "~1.16.4",`. Right now we have a @next dist tag pointed at a demo version of the wizard that is currently incompatible with that older version of PF. If we were to merge this PR as is it would break the wizards in `master`, but if you target `1.16. 5` it works fine (I just tested this). 
```
| => yarn info patternfly-react dist-tags
yarn info v1.0.2
{ latest: '1.16.5',
  next: '1.17.0' }
```
2. We probably should discuss this some more, but going forward, we will continue pulling in new PF React updates into our sprint branches and testing those (like we have done here), so I recommend not updating ManageIQ/react-ui-components `master` to point at a newer version of PF-React until our sprint ends if you can help it (i.e. v2v plugin `master` should point at the same version as ManageIQ/react-ui-components). We should stay in sync there...

Some bigger things I see happening here though (and we can probably discuss it soon)...I need to discuss this w/ some more UXD folks, but this is my initial thoughts:
* PF React is moving to monorepo soon and should be able to support many more components
* Any domain specific component repos downstream (such as ManageIQ/react-ui-components, v2v, etc) should try to stay in sync w/ PF-React (as it's upstream). I am sure there will be cases we would like to create domain specific components that are shareable amongst plugins in MIQ (I am start to think about these as well)... for that I would propose that we move towards a similar monorepo structure to the one @karelhala proposed. Also too - we can compare notes on how PF React is moving to monorepo and see if it helps. I kind of envision ManageIQ/react-ui-components to additionally have some patterns/wrappers that are not quite ready for PF-React and "in the works" upstream, but that we know work downstream in MIQ. Over time they should get more and more similar though...

thoughts?

himdel:
Aaah, so..


> yarn.lock

@karelhala .. v2v is using `yarn.lock`, so we need to update that one too...
(And 1.16.5 sounds better than 1.16.4, so why not :).)


> stay in sync

As for staying in sync, that definitely makes sense, but I don't think we can afford to get blocked by plugins in principle, so the whole "keep pf-react up to date every sprint" thing will probably need to be happening in ui-classic instead of v2v.

(I'm still hoping we can find a solution of ui-classic providing a set of libraries you can use without requiring those in each repo.)


> specific component repos

Yes, among other things, ManageIQ/react-ui-components can be though of both as a kind of staging area for stuff we want to get into patternfly-react, and a place for miq-specific components shared among the plugins. Monorepos do sound good too :).

priley86:
> (And 1.16.5 sounds better than 1.16.4, so why not :).)
1.16.5 is fine. Hoping to release 2.0.0 today or tomorrow after Wizard is merged.

> thing will probably need to be happening in ui-classic instead of v2v.
I guess if the implication is to funnel all pf-react components being used by plugins into ManageIQ/react-ui-components, I feel like this might become a blocker. The idea is to be able to iterate quickly inside our sprint branches w/ new pf-react code. I would rather just coordinate/stay in sync on when we are merging new versions of patternfly-react into the master branch of plugins or ui-classic. If you'd like us to make a PR into ui-classic whenever we are planning to update PF-React versions inside v2v, that's probably fine...We have been merging back into master at the end of every sprint, however this sprint will be special because I believe we would like to keep the existing UI in master for the Summit demo. (The POC UI is what's being demo'd at Summit i've heard and I don't want to interrupt this, but I don't think that stops us from continuing to iterate on latest code/latest flows for sprint6/sprint7 branches). 
@jeff-phillips-18 @ohadlevy do you have any more thoughts on this? @himdel I think this is reasonable... 

> Yes, among other things, ManageIQ/react-ui-components can be though of both as a kind of staging area for stuff we want to get into patternfly-react, and a place for miq-specific components shared among the plugins. Monorepos do sound good too :).
Ya I agree w/ this, if there are domain specific / MIQ specific components that are shared amongst plugins and not consumable from PF React, then we should use ManageIQ/react-ui-components. I would like that repo to go towards a monorepo structure though...

karelhala:
> Ya I agree w/ this, if there are domain specific / MIQ specific components that are shared amongst plugins and not consumable from PF React, then we should use ManageIQ/react-ui-components. I would like that repo to go towards a monorepo structure though...

I am plannig to move all `@ManageIQ/ui-components` and `@ManageIQ/react-ui-components` into monorepo. So we can have multiple packages and don't need to monitor multiple repositories.

We don't have solid solution (yet), but the idea is that we'd provide couple of shared packages we know ui-classic works best. If any new plugin would like to come with newer version of such package (let's say we use react in version 16.2.0 and some plugin wants to use 17.0.0) it has to be as part of his JS pack. If any plugin would like to use package which is not directly supported by ui-classic it has to include it as part of JS pack as well. I know that this might be harder to set up for plugins, but it's much better so we can guarantee that any new plugin will not break other plugin nor ui-classic.

PR #196, opened 2018-04-12
Aggregate Data Cards

michaelkro:
### Migrations Not Started Card
1. Pending transformation plan count is the number of plans returned from the API with no `miq_requests`
2. As soon as a transformation plan is kicked off, it's `miq_requests` array will no longer be empty, and, as a result, this count will be decremented

### Migrations in Progress Card
1. Active transformation plan count is the number of plans returned from the API where its most recent `miq_request` has the `status` of `active`
2. We check each active plan for errors by inspecting its most recent `miq_request`.  If this request has a `task` which has errored, we increment the error count.
3. Once a plan's most recent `miq_request` finishes, it will gain the `status` of `complete` or `failed`, which will result in the in-progress counter to decrement.  Further, if that request was in the error state before finishing, the error count also decrements

### Complete Migrations Card
1. Finished transformation plan count is the number of plans returned from the API where its most recent `miq_request` has the `status` of `complete` or `failed`
2. We check each finished plan for errors by inspecting its most recent `miq_request`.  If this request has a `status` of `failed`, we consider the plan as failed.
3. If a user "retries" a failed plan, a new `miq_request` is generated, and thus, the finished plan count is decremented. Further, since the retried plan is no longer considered finished, the error count will also decrement

## NOTES
1. Not sure what string to match on for a failed `request_task`.  Hence the placeholders like [this](https://github.com/priley86/miq_v2v_ui_plugin/compare/sprint6...michaelkro:fixtures-and-query?expand=1#diff-ad15b724358d9956d9ab108268b7e985R25)
2. Removes the unnecessary query for fetching all `transformation_plan_requests`.  The entire `Overview` will be driven off of fetching `transformation_plans`.  The fixtures are updated to reflect this
3. Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/150
## Screenshots
fullscreen_4_12_18__1_55_pm

agg-cards-1

fullscreen_4_12_18__1_52_pm

fullscreen_4_12_18__1_54_pm

agg-cards-loading


priley86:
🚢

app/javascript/react/screens/App/Overview/components/AggregateCards/ActiveTransformationPlans/ActiveTransformationPlans.js
```diff
@@ -0,0 +1,61 @@
...
+  const countDescription =
+    activePlans.length === 1
+      ? 'Migration Plan In Progress'
+      : 'Migration Plans In Progress';
+
+  const erroredPlans = activePlans.filter(plan => {
+    const [mostRecentRequest] = plan.miq_requests.slice(-1);
+    return mostRecentRequest.miq_request_tasks.some(
+      task =>
+        task.options.progress.current_description ===
```
priley86: for now - and per Bill... the task is failed if it's task.options.progress.current_description !== ''Virtual machine migrated'. This was the last direction I was given on this.

priley86:
This is great... and thanks a lot for spelling out the descriptions in the PR notes - this summarizes it very well.

@AparnaKarve @AllenBW @bzwei please let us know if you have any issue. The main thing to note here is that we are not directly driving status from the plan/request status for In Progress counts. Those counts should immediately reflect the more fine grained task status per @vconzola. This way a user will know immediately if a task fails, even though the migration may still be running. If the backend changes to accommodate this in it's status or we determine this logic is not needed, let us know.

app/javascript/react/screens/App/Overview/components/AggregateCards/ActiveTransformationPlans/ActiveTransformationPlans.js
michaelkro: Ok, I guess we can leave it with a placeholder for now.  We need the exact string match, b/c otherwise this filter will match on all the task descriptions (depending on which `/state` its in) like [here](https://github.com/priley86/miq_v2v_ui_plugin/pull/196/files#diff-9b2adb88683df284c00f835084862fd7R80)

priley86: ok - i agree. we will need to know the exact failure messages to look out for, otherwise it will be shown as error immediately. will ask @bzwei again...

michaelkro: Would `task.status === 'error'` if it was not done successfully?  @bzwei

michaelkro: ah, I guess we could check that `task.status !== 'Ok'` for failure

priley86:
this looks great as a first pass @michaelkro - i'm merging this so we can finish out the Migration section tomorrow. thanks!! 👍

PR #209, opened 2018-04-13
Plans Not Started Section

michaelkro:
* Render each not started transformation plan, displaying its `name`, `description`, and it's `vm_id` count
* Enable the "Migrate" button
  1. Sends a `POST` with the option `{ action: "order" }` to the transformation plan
  2. ~_All_~ Migrate button become disabled as UI waits for the POST response
  3. After the transformation request is successfully created, switch to the `Migrations in Progress` section
  4. Refetch all transformation plans
* Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/159

## NOTES
1.  After a successful POST, the transformation plan should now have a `miq_request`, which will no longer classify it as "not started".  So, in theory, it should be removed from the Not Started List after the refetch

2.  ~Loading state for buttons (disabling all) was just me going rogue.  Happy to remove or change this to whatever is appropriate.  Couldn't think of a way to make the individual buttons have their own loading state.~ 

3.  The mocks show the description and VM count in the reverse order, but our `ListView.Item` component makes this a little awkward.  Maybe another candidate for change to make more flexible upstream?

4.  Not sure if we need to be concerned with the interaction of the refetch here with our polling behavior.  Maybe there will be weird race conditions

## Screenshots
fullscreen_4_13_18__1_35_pm

fullscreen_4_13_18__1_40_pm


priley86:
> After a successful POST, the transformation plan should now have a miq_request, which will no longer classify it as "not started". So, in theory, it should be removed from the Not Started List after the refetch

Agree.

> Loading state for buttons (disabling all) was just me going rogue. Happy to remove or change this to whatever is appropriate. Couldn't think of a way to make the individual buttons have their own loading state.

Ya right now - i'm not sure what the expected behavior is. I'm used to seeing Buttons turn into Spinners in other apps when something happens, but maybe there is more specific PF specification.

> The mocks show the description and VM count in the reverse order, but our ListView.Item component makes this a little awkward. Maybe another candidate for change to make more flexible upstream?

I'm fine w/ flipping those if @serenamarie125 & @vconzola  are. The default description prop renders on the left w/ additional info (icons) on the right.

> Not sure if we need to be concerned with the interaction of the refetch here with our polling behavior. Maybe there will be weird race conditions

hmm yes - if a poll conflicts with an post/reload action, it seems like whichever one finishes first would win...and you'd be waiting for the next poll to come back... maybe we can look at a way to temporarily interrupt polling like we did w/ the Wizards/Overview before, but that should not hold this PR I don't think. Kind of an edge case which can be offset by looking at sockets or reducing the poll time it seems...

michaelkro:
@serenamarie125 good news!  The individual buttons are now being disabled after being clicked.  Re: elasped time, the section I'm working on is the "Not Started Migrations," so there is no elapsed time to worry about :) (I think, right? lol)

serenamarie125:
> Re: elasped time, the section I'm working on is the "Not Started Migrations," so there is no elapsed time to worry about :) (I think, right? lol)

😂you got me 👍

PR #219, opened 2018-04-16
Store migrations filter state in redux

michaelkro:
* Preserves the selected filter so that the user returns to the same
screen when navigating away from and back to the overview
* Converts Migrations to a functional component

## Notes
1. ~There is another bug with our `Card` component family.  Seeing this error when navigating back to `Overview` from `/plan/:plan_id`~ 

    Please see [comment](https://github.com/priley86/miq_v2v_ui_plugin/pull/219#issuecomment-381649021)
    ```
    Uncaught TypeError: Cannot read property 'querySelectorAll' of null
    at CardHeightMatching.js:90
    at Array.forEach (<anonymous>)
    at CardHeightMatching._matchHeights (CardHeightMatching.js:89)
    at CardHeightMatching.js:55
    ```

2. Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/216


priley86:
yep.. i think that bug is fixed by https://github.com/patternfly/patternfly-react/pull/301 ... but if we can get the Wizard in first in PF-R... it might be best to rebase that PR and apply it after? i'm happy to bring in the new `Wizard` in PF-R...then iterate on the `WizardPattern` as we have time...

priley86:
this looks good.. i think we merge the PF-R PR mentioned...potentially after Wizard... hopefully we can get both in very quickly to keep off sprint 7! 👍

priley86:
it's been merged now haha ... can you update pf-react and see if it checks out now?

michaelkro:
Confirmed [comment](https://github.com/priley86/miq_v2v_ui_plugin/pull/219#issuecomment-381649021) locally that updating to the latest version of `patternfly-react` fixes the striked(struck?)through note in the PR description

priley86:
nice improvement - good to go here 👍 .. merging..

PR #221, opened 2018-04-16
Adds parent datacenter to cluster name in infrastructure mapping wizard

### LINKED ISSUE #188: Add datacenter to cluster names in Create Infrastructure Mapping wizard
Wherever we display a cluster name in the mapping wizard we need to append the  associated datacenter where it lives. I looked at a few different options and prefer that we show it as "datacenter_name \ cluster_name". 
4aa-createmapping-clusters w datacentre - option 2

AllenBW:
closes #188

Rather than introducing a new call to get what we needed, an existing call was reconfigured to yield the desired results PLUS MOAR 🌮 💃 🎉 

~@serenamarie125 @vconzola so in the unlikely case a cluster does **NOT** have a parent name.. I need some guidance on preferred message, the ss below proposes "No Parent"~

### when no parent, we display only the clustername
screen shot 2018-04-17 at 12 44 58 pm


vconzola:
@AllenBW How can a cluster not have an associated datacenter?

AllenBW:
@vconzola Don't know, just tellin' yah what I see on my end, if this would ever be the case in the wild (who knows maybe its not possible) would be a good idea to already be ready for it

vconzola:
@AllenBW In that case I recommend a generic "datacenter_name_not_found" unless that exceeds the length limit for the field.

AllenBW:
Oh yah know it might be in the case of a local instance... @bzwei any thoughts on when or why we might not see parent datacenter? `v_parent_datacenter`


`datacenter_name_not_found`  does read a bit verbose

AllenBW:
@agrare gonna say `vmwarews`

serenamarie125:
@AllenBW can you add a screen with super long names 🙈

AllenBW:
screen shot 2018-04-17 at 1 40 24 pm


@serenamarie125 absolutely!  on the left pane, you'll see a super long parent name, right pane a super long cluster name

michaelkro:
@serenamarie125 @vconzola, should the new naming scheme also be reflected in the treeview?

vconzola:
Thanks @AllenBW. Can the complete name (not truncated) be displayed on hover in a tooltip?

@michaelkro Yes, since clusters can have the same name in different datacenters we should show the DC name in the treeview as well.

AllenBW:
screen shot 2018-04-17 at 3 46 32 pm

working tooltip, meanwhile @priley86 mock stuff still is AWESOMEEEEEE

AllenBW:
screen shot 2018-04-17 at 4 06 39 pm


@vconzola @priley86 hi.

Tooltips added to mapping entries!

agrare:
@AllenBW that's strange there should always be a datacenter.

AllenBW:
@agrare was using Dan's latest so nothing tooo real, if ya tell me there is always gonna be one, well then 👍 🙇

priley86:
@vconzola @serenamarie125 any more thoughts? i'm not 100% sure I follow when the ellipsis will take effect and when the tooltip would from the screenshot. Maybe a [giphy](https://giphy.com/apps/giphycapture) would help? ... also - you can just directly edit the HTML in chrome inspector in the "Elements" tab if you want... just edit the html directly to add more text... not sure if this helps.

AllenBW:
Oh I did nothing with the ellipsis, already in place, they show up whenever there is text overflow, editing html produced above overflow ss, heres a gif 👇 
yea

vconzola:
@AllenBW Just to be clear, the tooltip will only appear if the text is truncated, right? Like [this?](https://stackoverflow.com/questions/34721113/truncate-text-show-full-text-in-tooltip-on-mouseover)

AllenBW:
@vconzola incorrect, tooltip will always appear.

vconzola:
@AllenBW are we considering this done as is? in your gif above the tooltip doesn't seem to be working. when you enter the longer cluster name, the tooltip still shows the shorter name? I'm OK with closing this issue (once the tooltip matches the text) and opening a new issue for the enhancement @priley86 talked about in slack of using the [Foreman plugin](https://amirfefer.github.io/react-ellipsis-with-tooltip/?selectedKind=EllipisWithTooltip&selectedStory=in%20a%20table%20cell&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

AllenBW:
@vconzola gif is working, as is tooltip... long name is achieved by editing in html rather than database entry, can't so much edit the tooltip to show it...

So yeah lets call this one done. 

Yean we can look into the foreman plugin that applies tooltip on ellipse only 

cc @priley86

priley86:
it works on my machine @AllenBW 👏 👏 👏 
datacenter

opened #231 for the future...

merging.

PR #222, opened 2018-04-16
Fixed the Plan API parameters

AparnaKarve:
@michaelkro Can you review the changes?

Without these changes, I am seeing a 400 error for "Migration Plans not started"

- Reason for 400 - `created_on` and `updated_on` attributes do not exist. Instead, they should be `created_at` and `updated_at`

- Also, we need the `options` attribute


michaelkro:
@AparnaKarve @priley86 Oops, I just realized that we also need to fix the fixtures and places where we are referencing the old attrs `created_on` and `updated_on`

I'll fix that real quick

priley86:
thanks guys 👍

PR #225, opened 2018-04-17
Make entire migration in progress card clickable

michaelkro:
* Pass route props from react-router to Route components
* Create redirectTo method in Overview
* Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/217

## Notes
1. Normally the `react-router` route props are spread out into the component being rendered in `<Route>`, but in this case, due to `componentRegistry`, had to find a way to pass the data down.
2. Adds styles to make mouse cursor a `pointer` for in progress cards
3. `VMs Migrated` text is no longer a link

## Screens
click

### DIFF
--- a/app/javascript/react/config/Routes.js
+++ b/app/javascript/react/config/Routes.js
@@ -21,7 +21,15 @@ const Routes = ({ store }) =>
           path={`/${path}`}
           render={props => {
             if (props.match.isExact) {
-              return <React.Fragment>{markup}</React.Fragment>;
+              return (
+                <React.Fragment>
+                  {componentRegistry.markup(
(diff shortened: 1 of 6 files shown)


priley86:
wow so fast! ... this looks very good (and nice job with the React Router). 

The only issue I have is I believe setting `onClick` on the `Card` (a div) is not 100% accessible - but that's OK - this is not a v2v issue. In future sprints, maybe we could iterate in PF React on this and make the Card render an `<a>` wrapper ? I think it's probably best to open a Code Improvement issue, but the other Code Improvement issues would be easier to follow up on first as we have time...

That being said, I'm good w/ this and great job ⭐️

PR #226, opened 2018-04-18
Reset pagination when adding a filter

michaelkro:
Filtering the plan tasks returns a new list and thus pagination should
be reset

Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/215


priley86:
i like it 😸 

merging.

PR #232, opened 2018-04-19
Adds ellipsis with tooltip to cluster mapping wizard

AllenBW:
well that was surprisingly easy


priley86:
Red Hat React for the win!!! 🏆 

Beautiful that this was only a few lines to change...

i'm going to go ahead and approve this. Maybe we can contribute this to PF React this sprint since it's such a goodie ;) 🍬

vconzola:
@AllenBW Gorgeous!! Thank you.

PR #245, opened 2018-04-24
Update label and add spinner for pending migration card

### LINKED ISSUE #165: Migration Cards should start with a pending label
Currently, there's a lag once migration starts, and the cards are not updating immediately.  When a migration is initiated but not yet started, show the progress card with a "Pending" label.

mturley:
Closes #165 (p1).

screen shot 2018-04-24 at 3 49 19 pm

### DIFF
--- a/app/javascript/common/API.js
+++ b/app/javascript/common/API.js
@@ -35,4 +35,4 @@ export default {
   }
 };
 
-export const globalMockMode = true;
+export const globalMockMode = false;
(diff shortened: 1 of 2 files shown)


app/javascript/common/API.js
```diff
@@ -35,4 +35,4 @@ export default {
   }
 };
 
-export const globalMockMode = false;
+export const globalMockMode = true;
```
michaelkro: Makes tests unhappy

mturley: oh no! good catch.

mturley:
Passes `yarn test` locally now, sorry for the false start.

priley86:
thanks!

PR #247, opened 2018-04-24
Display a toast notification when a plan execution completes

### LINKED ISSUE #230: Display a toast notification when a plan execution completes
When a plan execution completes we need to display a toast notification. It does not need to persist. For plans where all of the migration tasks complete successfully:
overview - w all copy 3 filter is in progress w toast

For plans where one or more migration tasks fail:
overview - w all copy 3 filter is in progress w toast - failed

In either case selecting the View details link drills down to the details page for the plan.

AllenBW:
fixes #230

### But what does it look like? 
asdfasdf

[The only real nasty here is with the notification actions, itz just a lil hardcoded, would need to be reconfigured to be an action other than a link (but that can be done a lil later)](https://github.com/priley86/miq_v2v_ui_plugin/pull/247/files#diff-2740fa0c4333174ee3f062ae8c2333e9R31)

Oh yeah, so whats the big deal about this, well now anyone/where can create notifications anywhere! (it looks something like this)
```
addNotificationAction({
        message: `your message`,
        notificationType: 'success' || 'error' || 'info' || 'warn',
        data: // if yah wanta add some more details to the notification
        persistent: // if the notification should await user input for dismissing,
        timerdelay: // typically 8000 aka 8 sec,
        actionEnabled: // if yah wanta add an action (but see above)
      });
```

HUGE 🤗 TO @michaelkro FOR THE 💪 💘


AllenBW:
at your earliest free moment, @priley86 or @michaelkro or @mturley  could use some :eyes: on this pr as imma kinda stuck... unsure how to interact with the store via the reducer through the component..  (if that makes sense)

[in particular would like to talk aboutttttt this line
](https://github.com/priley86/miq_v2v_ui_plugin/pull/247/files#diff-2740fa0c4333174ee3f062ae8c2333e9R20)

michaelkro:
Hey Allen, you have to register your new reducer and `mapDispatchToProps`.  Let me dig up some examples :)

michaelkro:
You might need to register your component [here](https://github.com/priley86/miq_v2v_ui_plugin/blob/master/app/javascript/components/index.js)

@priley86, I'm not sure about this one, can you advise?

priley86:
yea i was just trying to chat w/ @AllenBW about using the reducer keys (redux keys) how we have been in the other views...

priley86:
i would create an `index.js` and a `NotificationListActions.js` just to be consistent with the others in this project so far... that was all based on this new folder structure proposed a bit back by the Foreman group:
https://github.com/theforeman/foreman/pull/5240

basically just a common scaffolding for Redux connectors/containers (so you could follow any of the other examples really).

I think after Summit I can probably write a generator to help w/ this.. (i know its more boilerplate)...

priley86:
ironically...there are a lot of other redux middlewares with _much more_ boilerplate...and that's why redux-promise was selected. That being said - apollo-link-state would eliminate most that entirely...

AllenBW:
gotta ask for a lil :hug: again... so close... almost there... buuuuuuuuuuut changes to the state aren't being picked up by the component is anything glaring?
[`NotificationList.js` the component that's playing 💀  :fish:](https://github.com/priley86/miq_v2v_ui_plugin/pull/247/files#diff-2740fa0c4333174ee3f062ae8c2333e9R1)

michaelkro:
@AllenBW, I'm not seeing any instances of the notification list in Overview

AllenBW:
@michaelkro not sure I understand what you're looking at...  with `globalMockMode = true` this is the state tree created..
screen shot 2018-04-26 at 1 30 01 pm

michaelkro:
Wow, love that diagram!  Maybe I misunderstood you...
> changes to the state aren't being picked up by the component

I was just noting that I don't see `<NotificationList>` in the Overview component, so it's not ever being rendered.  I am seeing your new redux key in my store though

fullscreen_4_26_18__1_35_pm

AllenBW:
OH yeah yeah its nottttt..... figured out i was missing an entry here: https://github.com/priley86/miq_v2v_ui_plugin/blob/master/app/javascript/components/index.js

will push that change... still no bueno though :-/

AllenBW:
@michaelkro mmmm here's a thought... maybe it also has something to do with the structure... the `notifications` object holding the `notifications` array...
screen shot 2018-04-26 at 1 48 38 pm

michaelkro:
I think I misunderstood your implementation.  It looks like you want the NotificationList to render globally, which makes sense.  I originally was thinking the 🍞 would only show on the overview route

One thing, are you using the JS array method `filter` [here](https://github.com/priley86/miq_v2v_ui_plugin/pull/247/files#diff-a190e786e9e6bbf051210809a2c83f80R32)

AllenBW:
Yeah global rendering is the goaaaaal, link you posted, which line is it referring to?

michaelkro:
Line 32 in NotificationListReducer.js

AllenBW:
YEP yeah I see it now, clicking from the url instead of pasting produces different results, yesssssss... you understand correctly unsure if it works, because that action is only called from the component that doesn't render..

app/javascript/react/screens/App/common/NotificationList/NotificationList.js
```diff
@@ -0,0 +1,56 @@
...
+        {notifications &&
+          notifications.map((notification, index) => (
+            <React.Fragment>
+              <TimedToastNotification
+                key={index}
+                type={notification.notificationType}
+                persistent={notification.persistent}
+                timerdelay={notification.timerdelay}
+                onDismiss={removeNotificationAction({ key: index })}
+                onMouseLeave={removeNotificationAction({ key: index })}
```
michaelkro: since you're actually calling these two functions above (removeNotificationAction & removeNotificationAction), you need something like

```
onDismiss={() => removeNotificationAction(...)}
```

otherwise, those actions will fire immediately after rendering

michaelkro:
So, good news, I was able to get something on my screen, but it's not in the global way you want, but at least a step in the right direction

fullscreen_4_26_18__2_17_pm

michaelkro:
@AllenBW, I got it rendering globally now.  Let me know if you want to get on a call.  Or, if you prefer, I can leave PR comments

AllenBW:
Back to wipe, need to change a few things, @vconzola looking through review your happiness!!

app/javascript/react/screens/App/Plan/index.js
```diff
@@ -1,17 +1,23 @@
...
 
 const mergeProps = (stateProps, dispatchProps, ownProps) =>
   Object.assign(stateProps, ownProps.data, dispatchProps);
 
-export default connect(mapStateToProps, PlanActions, mergeProps)(Plan);
+export default connect(
+  mapStateToProps,
+  Object.assign(PlanActions, NotificationActions),
+  mergeProps
+)(Plan);
```
michaelkro: In the same vein of thought in my last comment, we don't actually need to inject the `notifications` state nor actions into `Plan`.

We provide `Overview` with the NotificationActions b/c we actually need to call `addNotificationAction()` from within `Overview.js` (so with this in mind, we technically don't need to import all the actions, just the one we need, but a minor detail imo)

But Plan, nor any of its nested components, need access to any of the Notification stuff, yet

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -89,6 +88,32 @@ class Overview extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    const { finishedTransformationPlans, addNotificationAction } = this.props;
+    const freshMigrations = finishedTransformationPlans.filter(
```
michaelkro: We might need to do something else to monitor for `freshMigrations`

Since the `finishedTransformationPlans` are being passed down into `Overview` as props, and they have an initial value of `[]`, a user is going to see a toast for every finished migration when first logging in (which is why we're seeing the two in `mockMode`)

app/javascript/react/screens/App/Plan/index.js
AllenBW: *yet* was my thought as well...

app/javascript/react/screens/App/Overview/Overview.js
AllenBW: Is that not the desired behavior? Might need ux to weigh in on when they want to see them...

michaelkro:
@vconzola, if a migration completes while a user is _not logged in_, do we expect to see a toast when logging in?

vconzola:
@michaelkro Only if the migration fails.

michaelkro:
@vconzola there may not be a way to handle this purely in the front end.  Basically, we need to flag a migration request to indicate whether it's toast has been shown or not, and this can really only be persisted from session to session on the back end

AllenBW:
yeah thats a new ask... not gonna get by todayyyyyyyy 😆 

SO for the demo video, is it using live data? or the mock data? if mock, i suggest we keep this as is... else you'll never see anything, if not... i can move the notification creation to polling and we'll see em, when the move from in process to finished

michaelkro:
@AllenBW I just chatted with @vconzola, and the desired behavior, for now, is just to see notifications when the user is actually using the v2v plugin

1. Show a toast when a plan changes status from "in progress" to "complete"
2. No notifications on login or refresh

michaelkro:
Sorry, I know, this makes it impossible to test now...  since the interaction happens on data changing

AllenBW:
@michaelkro @vconzola updated behavior so only changes observed during polling will throw a notification, leaving the gif on the top cuzzzzzzz yeah it looks pretty

PR #249, opened 2018-04-25
Plans completed details

michaelkro:
* Clicking on a finished migration plan (aka complete) transitions to its details page
* User can filter by Host Name or by Task Status. Since the plan is finished, `Ok` and `Error` are the only two possible statuses
* User can sort by Time Started, Host Name, and Task Status
* Closes https://github.com/priley86/miq_v2v_ui_plugin/issues/218

## Notes
1. Screenshots below will differ slightly from provided mock. Reasoning is to maintain consistency with the `In Progress` details page

## Screens
fullscreen_4_26_18__8_33_am

fullscreen_4_26_18__8_33_am-2

fullscreen_4_25_18__8_44_am

fullscreen_4_25_18__8_46_am

### DIFF
--- a/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList.js
+++ b/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList.js
@@ -24,6 +24,75 @@ import sortFilter from './sortFilter';
 import paginate from './paginate';
 
 class PlanRequestDetailList extends React.Component {
+  static getDerivedStateFromProps(nextProps) {
+    const filterTypes = nextProps.planFinished
+      ? [
+          {
+            id: 'transformation_host_name',
+            title: 'Host Name',
+            placeholder: 'Filter by Host Name',
+            filterType: 'text'
+          },
+          {
+            id: 'status',
+            title: 'Status',
+            placeholder: 'Filter by Status',
+            filterType: 'select',
+            filterValues: [
+              { title: 'Ok', id: 'Ok' },
+              { title: 'Error', id: 'Error' }
+            ]
+          }
+        ]
+      : [
+          {
+            id: 'transformation_host_name',
+            title: 'Host Name',
+            placeholder: 'Filter by Host Name',
+            filterType: 'text'
+          },
+          {
+            id: 'message',
+            title: 'Status',
+            placeholder: 'Filter by Status',
+            filterType: 'select',
+            filterValues: [
+              { title: 'Pending', id: 'Pending' },
+              { title: 'Validating', id: 'Validating' },
+              { title: 'Pre-migration', id: 'Pre-migration' },
+              { title: 'Migrating', id: 'Migrating' },
+              {
+                title: 'VM Transformations completed',
+                id: 'VM Transformations completed'
+              }
+            ]
+          }
+        ];
+    const sortFields = nextProps.planFinished
+      ? [
+          { id: 'delivered_on', title: 'Started', isNumeric: true },
+          {
+            id: 'transformation_host_name',
+            title: 'Host Name',
+            isNumeric: false
+          },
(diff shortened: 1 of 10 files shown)


app/javascript/react/screens/App/Plan/components/PlanRequestDetailList.js
```diff
@@ -24,6 +24,75 @@ import sortFilter from './sortFilter';
 import paginate from './paginate';
 
 class PlanRequestDetailList extends React.Component {
+  static getDerivedStateFromProps(nextProps) {
+    const filterTypes = nextProps.planFinished
+      ? [
```
priley86: do you mind pulling the filter types arrays out into constants? (i.e. maybe into PlanRequestDetailListConstants.js or something?)

app/javascript/react/screens/App/Plan/plan.planRequests.fixtures.js
```diff
@@ -248,6 +248,70 @@ for (let i = 0; i < 30; i += 1) {
...
+  const copy = JSON.parse(JSON.stringify(sampleTask));
+  copy.id = i;
+  copy.miq_request_id = i;
+  copy.options.delivered_on = moment(copy.options.delivered_on)
+    .subtract(Math.floor(Math.random() * 48), 'hours')
+    .format();
+  copy.options.updated_on = moment(copy.options.delivered_on)
+    .add(Math.floor(Math.random() * 100), 'm')
+    .format();
+  copy.message = 'VM Transformations completed';
```
serenamarie125: @vconzola @michaelkro super nit, but wondering if this should be "VM Transformations Completed" OR "VM transformations completed" ?  this would affect both the Infotip title & the value in the table, if we decide to change it

serenamarie125: @vconzola @michaelkro super nit, but wondering if this should be "VM Transformations Completed" OR "VM transformations completed" ? this would affect both the Infotip title & the value in the table, if we decide to change it

michaelkro: I went ahead and went down the capitalization route by using `text-transform: capitalize` to capitalize each word in the string.

I feel like this is better than manually manipulating the message being returned by the API to be sentence-cased (not sure if this is a real term, but going with it)

vconzola:
@michaelkro @serenamarie125 We should not be calling them transformations at all. Transformations is the term being used by the backend because they already have a different meaning for migrations (migrating between hosts on a single provider), but Marco is adamant that we call everything "migrations" in the v2v UI. It sounds like this will be extra work, so we can defer it for now, but we need to change it after Summit. I'll open an issue once we have backend wiring in place and I can do a real UI review.

priley86:
yep - additional work to create our own dictionary of backend statuses, for sure! 

The last correspondence on this w/ Fabien on April 6th he said the following:
> "Update the task message with one of the 5 states we agreed on: Pending, Validating, Pre-migration, Migrating, Migration complete. You can get it with task['message'] (L8). We still have an issue there: the native code overrides the status to "VM Transformations completed" when it's successful. I have to check will Bill if this behavior can be changed."

It seems we'll have to wait until the API is completed to determine the UI solution...

priley86:
@bzwei do you have any update?

vconzola:
@priley86 Sounds good to me.

priley86:
per discussion at standup - going to go ahead and merge this. @michaelkro can you open the additional issue for the status message changes next sprint?

app/javascript/react/screens/App/Plan/plan.planRequests.fixtures.js
serenamarie125: great idea @michaelkro!

PR #253, opened 2018-04-26
Make check_package_json executable

mturley:
Oops. Fixes this error:

```
[GIT-HOOKS WARNING] Non-executable file /Users/mturley/git/manageiq/plugins/miq_v2v_ui_plugin/.githooks/post-merge/check_package_json is skipped
```


mturley:
@michaelkro, this reveals to me a way easier method for you to avoid this `check_package_json` script if you don't want it! you just need to make it non-executable.

PR #263, opened 2018-04-30
Display a toast notification when a plan execution completes

AllenBW:
same as #247 cept into the alpha branch rather than sprint7


priley86:
i decided to name sprint8 branch `gaprindashvili²`

AllenBW:
we're good to go on this one!

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -82,13 +91,51 @@ class Overview extends React.Component {
...
+      const oldMigrationIds = prevProps.finishedTransformationPlans.map(
+        plan => plan.id
+      );
+      const freshMigrations = finishedTransformationPlans.filter(
+        migration => !oldMigrationIds.includes(migration.id)
+      );
+
+      freshMigrations.forEach(plan => {
+        const planStatus =
+          plan.miq_requests[0].status.toLowerCase() === 'complete';
```
michaelkro: oops, missed this my first time through.  A plan can have more than one request, for example, if a plan's first request fails, and then the user retries, the plan will have a total of 2 requests

We need to do something like
```javascript
const [mostRecentRequest] = plan.miq_requests.slice(-1)
```

priley86:
thanks 👍 - is the scrolling issue fixed?

priley86:
so can these existing [inline styles](https://github.com/priley86/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/Overview.js#L80) now be removed?

AllenBW:
Oh if we encapsulate it in a fragment rather than a div, yeah i believe so!

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -182,10 +230,7 @@ class Overview extends React.Component {
     );
 
     const overviewContent = (
-      <div
-        className="row cards-pf"
-        style={{ overflow: 'auto', paddingBottom: 1, height: '100%' }}
-      >
+      <React.Fragement className="row cards-pf">
```
michaelkro: :) `Fragement`

michaelkro: But, it looks like we actually still need those inline styles

priley86:
ok - i'm good this one for now. Filed [265](https://github.com/priley86/miq_v2v_ui_plugin/issues/265) for future integration to the Detail view...

nice job @AllenBW 😄

PR #268, opened 2018-04-30
completed time w/ moment

priley86:
It seems that our i18nProvider is not setting the local UTC time. Can we use browser timezone or use server APIs to get the timezone offset? Today it just displays GMT, so I've reset to using moment in the uxd-summit-booth branch for now...

@AllenBW can you take a peek?


priley86:
if we can get the localized UTC offset working correctly...it would also be great to start using RelativeDatetime in the future to demonstrate our elapsed times (since is a real time seconds counter built-in). Although all of this may just be easier if we use moment.js...

AllenBW:
so so so so so sorry for the 2 day delay in review on this, I might be losing my mind... 😆 

down with using moment, but wondering if we want to maintain the same completed time format? watcha suggest in this pr loooooks like

#### this
screen shot 2018-05-02 at 10 32 22 am

#### but it used to look like this! 
screen shot 2018-05-02 at 10 31 57 am

@vconzola preference of date format? or no preference?

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -94,12 +91,7 @@ const InfrastructureMappingsList = ({
               >
                 <Grid.Col sm={12}>
                   {__('Completed: ')}
-                  <IntlDate
-                    data={{
-                      date: mapping.created_at,
-                      defaultValue: 'Default value'
-                    }}
-                  />
+                  {moment(mapping.created_at).format('YYYY-MM-DD HH:mm:ss A')}
```
AllenBW: `{moment(mapping.created_at).format('MMMM Do YYYY, h:mm a')}`

will maintain existing formatting

vconzola:
@AllenBW  I prefer the "February 7, 2018..." style, but I'm also not sure if there's a "standard" across our products. Good question for @serenamarie125

priley86:
ok thanks @AllenBW 👍 

as long as it's consistent fine w/ me. That was the previous format being used in the migration cards - but it looks like that's changed. It's also needing updates in the detail screen tooltip (`Started`:).

I can update both of those but just wanted to check w/ you first.

priley86:
updated: 
screen shot 2018-05-02 at 11 16 51 am

priley86:
screen shot 2018-05-02 at 11 18 10 am

PR #286, opened 2018-05-02
fix the flicker using a stale prop

### LINKED ISSUE #266: UI flicker on plan details
After integrating the backend API in #260 , i noticed a bit of flicker on the plan detail screen. It would be great to add a loading spinner instead of showing the empty VMs messaging...

flicker

priley86:
Fixes #266 

* ensures Plan Not Started logic on plan details screen is consistent with overview screen
*  move `planRequestPreviouslyFetched` flag to derived state and reset it to false on load. This ensures that when reloading the detail screen, we initialize that state to false. I would like to revisit this issue next sprint w/ Web Sockets.


AparnaKarve:
Works really nice! Thanks @priley86 !!

mturley:
Agh, you beat me to it. I was going to push a solution but it wasn't as nice. Thanks @priley86

PR #287, opened 2018-05-03
fix not started regression

priley86:
The `planNotStarted` in plan detail **should not** have the same logic as plan not started in plan overview (apologies, I just now realized this after fixing the flicker). 

Otherwise, we can't view tasks for Plan Not Started on the Detail.  

cc: @AparnaKarve @michaelkro 

ok to merge?


AparnaKarve:
@priley86 Thanks!
I missed testing this use case yesterday.

AparnaKarve:
looks good -- merging.

PR #300, opened 2018-05-07
Fixes notification removal

### LINKED ISSUE #275: Toast notification disappearing too quickly after successful migration
When a plan completes successfully the toast notification is supposed to appear for 8 seconds or until the user interacts with it.  Currently, the toast seems to disappear as soon as the plan migration card is  closed and the aggregate status card is updated.

AllenBW:
fixes #275  (the last commit in the pr does, ends up timed notification do not enjoy null timerdelays)
**OH** its gotta be said, persistent overrides timerdelay, so if something has a timerdelay but `persistent=true` it'll obey the latter. 

When multiple notifications were fired simultaneously, with persisted=false, the last notification would not be removed from the list.  This adds unique ids to each notification, ensuring everything is removed.

### Also fixes migration div enclosed by a p error
screen shot 2018-05-07 at 12 10 10 pm

### Also fixes failed invalid prop type errors for migrationcompletedlist card
screen shot 2018-05-07 at 12 11 04 pm

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -141,7 +141,6 @@ class Overview extends React.Component {
             id: plan.id
           },
           persistent: !planStatus,
-          timerdelay: planStatus ? 8000 : null,
           actionEnabled: true
         });
       });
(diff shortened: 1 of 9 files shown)


priley86:
reallllllly nice... 👍 . merging.

PR #303, opened 2018-05-07
Use window api methods for correct session timeout

### LINKED ISSUE #283: Timed Out migration screen keeps repeatedly prompting for login
https://bugzilla.redhat.com/show_bug.cgi?id=1573547

@priley86 I believe Brett will reach out to you to prioritize this. Please review as it may be an urgent bug to fix.

AllenBW:
fixes #283
closes https://bugzilla.redhat.com/show_bug.cgi?id=1573547

~So this is the first part of it, but the second part is taking a dip into each reducer and component ensuring each are still getting what they need... Overall, looks that using `window` tools unwraps a layer of abstraction, the `data`.  If feels a little dirty change EACH AND EVERY place that is looking for something along the lines of `actions.payload.data.resources` so gonna keep gnawing on that locally for a bit...~

thanks to the `transformResponse` param done in https://github.com/ManageIQ/manageiq-ui-classic/pull/3662 we're able to make this a REAL short n sweet pr <3

### DIFF
--- a/.jest-setup.js
+++ b/.jest-setup.js
@@ -7,3 +7,4 @@ global.__ = str => str;
 global.n__ = str => str;
 global.sprintf = str => str;
 global.Jed = { sprintf: str => str };
+global.API.get = jest.fn(() => Promise.resolve());
--- a/app/javascript/common/API.js
+++ b/app/javascript/common/API.js
@@ -1,35 +1,31 @@
-import axios from 'axios';
-
-const getAuthToken = () =>
-  localStorage.miq_token ? localStorage.miq_token : '';
-
-axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
-axios.defaults.headers.common['X-Auth-Token'] = getAuthToken();
+const API = window.API; // eslint-disable-line prefer-destructuring
 
 export default {
   get(url, headers = {}, params = {}) {
-    return axios.get(url, {
+    return API.get(url, {
+      transformResponse: e => ({ data: e }),
       headers,
       params
     });
   },
   put(url, data = {}, headers = {}) {
-    return axios.put(url, data, {
+    return API.put(url, data, {
       headers
     });
   },
   post(url, data = {}, headers = {}) {
-    return axios.post(url, data, {
-      headers
+    return API.post(url, data, {
+      headers,
+      transformResponse: e => ({ data: e })
     });
   },
   delete(url, headers = {}) {
-    return axios.delete(url, {
+    return API.delete(url, {
       headers
     });
   },
   patch(url, data = {}, headers = {}) {
-    return axios.patch(url, data, {
+    return API.patch(url, data, {
       headers
     });
   }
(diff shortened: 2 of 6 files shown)


AllenBW:
It should be mentioned... we're waiting on https://github.com/ManageIQ/manageiq-ui-classic/pull/3662 to show up... Once it does, this will be wholly reworked 😁

AllenBW:
Oh yeah, gotta update tests..

AllenBW:
@priley86 @michaelkro looks like [`mockRequest.js`](https://github.com/priley86/miq_v2v_ui_plugin/blob/master/app/javascript/common/mockRequests.js) is tightly coupled with axios being the api interface... with axios gone, unsure how yall wanta proceed here... 🤔

priley86:
thanks for pushing this one to the 🏁 @AllenBW 

i'm fine w/ any API / session modifications that need to be made to resolve the session issue presented. I'm not super knowledgeable about this w/ MIQ/Ruby-on-Rails but I will try to follow along and test w/ you.

in a future architecture in a galaxy far far away... i would love to be able to document such APIs / request headers via something like an api-docgen. I think the documentation would help us a lot in the future, and I'm definitely in the camp that prefers to dockerize the front end (apologies for using the word dockerize so much 😆). We have some folks in UXD exploring this now w/ Dockerized builds, but I think this is pretty far away from our current MIQ/RoR structure...
https://github.com/cdcabrera/apidoc-mock

that being said - let me know if you have any more thoughts on this sort of thing down the road...

AllenBW:
So we gotta do away with axios and use whats advertised on the window under `API`  then apply a response transformation cuz v2v everywhere is expecting to see things in the form `payload.data.foo` (but that's besides the point) 

My question isssssss how do we keep tests happy while doing away with axios 🤔 (also a question for the self, its the issue I'm presently working)

priley86:
sooo i am fine w/ the approach here of using `window.API` since it seems that is now intended for MIQ.

* tested the UI w/ your branch locally and ensured things were kosher. I don't seem to see the session issues but will keep an eye out for this.
* regarding the `*.test.js` changes - i think it is safe to just globally apply this for all `API.get` methods in our unit tests now...soo w/ that being said - do you mind removing those individual `global.API.get = jest.fn(() => Promise.resolve());` statements and just adding it once globally in `.jest-setup.js`?

i.e. this seems to have the same effect:
```
import { configure } from 'enzyme';
import Adapter from 'enzyme-adapter-react-16';
configure({ adapter: new Adapter() });

// Mocking translation function
global.__ = str => str;
global.n__ = str => str;
global.sprintf = str => str;
global.Jed = { sprintf: str => str };
global.API.get = jest.fn(() => Promise.resolve());
```

other than this - i'm very happy 😀 😄 😁and I think you've 🔨 this 🐛 

@AparnaKarve do you have any additional thoughts on this one?

AparnaKarve:
@AllenBW @priley86 Let me test this in the `Download log` context, where we use an `http` call instead of an `API` call.
If that does not work, I can work on a separate PR to make that work, since this PR seems to be working well for `API` calls -- and then we can merge this PR.

priley86:
✈️ ... 📦 ... 🚢

PR #304, opened 2018-05-07
w.i.p. starts download log feature

### LINKED ISSUE #208: When looking at migration plan details, a "download log" link should be available for each VM
When drilling into a Migration Plan, the UI shows each VM.  This issue captures the need to add a "Download Log" link for each VM entry.

There is an API dependency here.

priley86:
Adds the download log link UI elements for In Progress/Completed plan details  (Not Started detail view is unchanged).

cc: @AparnaKarve - can you begin backend integration?

@vconzola @serenamarie125 - we will need to decide interaction. The way I understand this - we will make an async request to download the log file, the backend will gather the log file from the conversion host, then a Rails controller will submit the file to the UI and our download log promise will resolve. B/c of this - I'd be in favor of showing a spinner OR disabling the link during download. Additionally - if log download request fails, we need some way to notify the user.

thoughts on these scenarios?

Initial screenshot (moved disk utilization to list view additional info section and download log to list view action section, ensured responsive spacing):
screen shot 2018-05-07 at 4 39 32 pm

closes #208

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -111,7 +111,8 @@ class Plan extends React.Component {
       isRejectedPlan,
       isFetchingPlan,
       isQueryingVms,
-      isRejectedVms
+      isRejectedVms,
+      downloadLogAction
     } = this.props;
 
     const {
@@ -148,6 +149,7 @@ class Plan extends React.Component {
               <PlanRequestDetailList
                 planFinished={planFinished}
                 planRequestTasks={planRequestTasksMutable}
+                downloadLogAction={downloadLogAction}
               />
             )}
           {!planNotStarted &&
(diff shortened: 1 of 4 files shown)


AparnaKarve:
Thanks @priley86 - will work on this branch to add the backend pieces

priley86:
@AparnaKarve sounds good!

@serenamarie125 i'm OK w/ this! 

The hover/underline is enabled by PF default! ❤️  I will add the toast after @AparnaKarve is ready. 👯 
download-hover

AparnaKarve:
I have made further progress on this PR and have discovered a few things that might need some UX adjustments with the current design.

- The download operation is a queued operation, meaning, it can't be instantaneous.
In light of that, we might want to show a spinner or something equivalent to indicate a Request-pending state. @serenamarie125 @vconzola Do you have any suggestions on how to accommodate a spinner in the current design?
Note that when a Migration is in progress, we are already showing a spinner in the list, adding another spinner while downloading the log, may be a bit too much.

- On the backend side, I have the option to set a timeout for the "wait" operation 
(the "wait" operation is the waiting process where we are waiting for the blob of log contents from the host machine)
Hence I was thinking, maybe we should set a timeout for 20 to 30 seconds instead of waiting indefinitely. @priley86 WDYT about having the timeout?
If at the end of the timeout there is no blob of log, then there will be a TIMEOUT error.

AparnaKarve:
To get a better idea about the log download time, I have set up a demo system here -
https://cf-06f5.rhpds.opentlc.com
(credentials to login are what Miguel gave us the other day)

The download time is about 15 seconds on this system.

PR #308, opened 2018-05-11
Fix the Elapsed time in Completed Migrations list

AparnaKarve:
- Fixed the Elapsed time in Completed Migrations list

Before -
screen shot 2018-05-11 at 11 34 24 am

After -
screen shot 2018-05-11 at 11 33 04 am

The second commit dbe87b6 is about an adjustment to the `mostRecentRequest` calculation based on a technicality (explained in the commit message)

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -118,7 +118,7 @@ class Overview extends React.Component {
       freshTransformationPlans.forEach(plan => {
         const mostRecentRequest = plan.miq_requests.reduce(
           (prev, current) =>
-            prev.updated_on > current.updated_on ? prev : current
+            prev.created_on > current.created_on ? prev : current
         );
 
         let planStatusMessage = sprintf(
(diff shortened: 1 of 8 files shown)


app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -118,7 +118,7 @@ class Overview extends React.Component {
       freshTransformationPlans.forEach(plan => {
         const mostRecentRequest = plan.miq_requests.reduce(
           (prev, current) =>
```
priley86: since we are starting to repeat this logic a lot of places, it seems like this is a good opportunity to make it a function. Do you mind creating a simple helper in `screens/App/common/helper.js` which accepts an array of requests and returns most recent?
```
export const mostRecentRequest = (requests) => {
return requests.reduce(
    (prev, current) => (prev.updated_on > current.updated_on ? prev : current)	+    (prev, current) => (prev.created_on > current.created_on ? prev : current)
   );	   );
}
```

priley86: otherwise..this PR looks good to me @AparnaKarve !

AparnaKarve: Sure! That's a good idea @priley86

AparnaKarve: @priley86 I have extracted the `mostRecentRequest` logic in the `app/javascript/react/screens/App/common/getMostRecentRequest.js` helper.

(It should be easier to review this change visually if you look at 3bde4ef)

priley86:
Looks good! Merging.

PR #309, opened 2018-05-11
Add 'download log' support

AparnaKarve:
Added backend controller to send the log contents to the frontend.

Note: 'download log' is a queued operation, and hence there is a delay of a couple seconds before a log can be downloaded.

Screenshots -

Download in progress 

screen shot 2018-05-11 at 2 36 46 pm

------------------------

Download successful

screen shot 2018-05-11 at 2 35 57 pm

------------------------

Chrome footer shows the downloaded log

screen shot 2018-05-11 at 2 37 48 pm

------------------------

Error during download

screen shot 2018-05-11 at 2 30 04 pm

------------------------

Fixes https://github.com/priley86/miq_v2v_ui_plugin/issues/208


AparnaKarve:
Note that there are currently no spinners when a download is in progress, instead a label will be displayed that says - "Download log is in progress..." -- see screenshot above.

We can replace this with something better if needed.

vconzola:
@AparnaKarve @priley86 I think the way you implemented it with changing the label and not having a spinner is perfect. Thanks for getting this done so quickly.

priley86:
the demo environment worked for me as well, and I just tested it locally. It shows the error alert when in `simulate_queue_worker` mode, but that is expected. LGTM

app/javascript/react/screens/App/Plan/PlanActions.js
```diff
@@ -71,3 +73,49 @@ export const fetchPlanAction = (url, id) => {
 export const resetPlanStateAction = () => ({
   type: RESET_PLAN_STATE
 });
+
+export const downloadLogAction = (task, addNotificationAction) => dispatch =>
```
michaelkro: An alternative to way to get access to `addNotificationAction` could be to directly import the action into this file from `NotificationListActions.js.

That would eliminate the need to pass the action down through the component tree as a prop

AparnaKarve: @michaelkro Sure, will give that a try. Thanks.

AparnaKarve: @michaelkro So directly importing `addNotificationAction` and calling it directly did not work. So I am using `dispatch` on `V2V_NOTIFICATION_ADD ` instead.
Is that OK?

michaelkro: Absolutely!  In fact, I think this makes much more sense than my suggestion (sorry about that by the way 😬, I should have double checked before commenting)

AparnaKarve: No problem! Thanks for confirming!!

AparnaKarve:
@priley86 @michaelkro Should I get rid of this line from the reducer?
`.set('migrationTaskLog', action.payload.data)`

The payload in this case is the log file (contents of the log file), the size can be anywhere around 5 to 6 MB.
Since I am not using `migrationTaskLog` anywhere, no point in keeping it in the redux state.
Do you agree?

priley86:
yes - i most definitely agree w/ you. That will reduce memory footprint a lot it sounds. If those variables can be constrained to the download function, they'd be released after I assume (unlike redux state). Very good catch on this.

AparnaKarve:
@priley86 / @michaelkro Can you merge please? So that this gets included in tonight's nightly.
Thanks.

PR #320, opened 2018-05-16
Display "Download in progress" only for the tasks that requested the download

AparnaKarve:
The Reducer now maintains an array of task id's for the tasks that requested a log download.

Only those tasks that clicked on the "Download Log" link will now display the text "Download in progress".

Fixes https://github.com/priley86/miq_v2v_ui_plugin/issues/318

screen shot 2018-05-16 at 3 48 28 pm


priley86:
thanks @AparnaKarve !

PR #325, opened 2018-05-20
plan error reporting popup

### LINKED ISSUE #299: Provide migration error reporting at the plan level
Migration Plans with any aggregated failures currently show an error status.  We should provide a popover, accessible by clicking this error icon, including a high level message noting how many vms have failed at that point

@priley86 do we need an API for this, or do we already have support?
FYI @vconzola

priley86:
Closes #299 

Adds the plan error reporting popup:
* do not proceed to detail if the error icon is clicked (ignore the default card click behavior for the icon)
* show the popup with "{x} of {y} VM migrations failed." message. Use the most recent execution request to determine the number of failed VM tasks.

screen shot 2018-05-20 at 10 25 40 am


AparnaKarve:
Tested in UI. Looks good!

PR #326, opened 2018-05-20
fix spacing on detail view

priley86:
While adding the error reporting popup, I noticed some spacing issues w/ the plan detail list view. This adds a min-width to the description so that spacing is consistent.

Before:
before

After:
after


AparnaKarve:
Tested. Looks good!

PR #330, opened 2018-05-21
Remove unused MiqV2vUI::ApplicationController

himdel:
Looks like this may have been originally intended to be the plugin base controller, but:

a) such a thing needs to live in ui-classic, to include all the necessary helpers etc.
b) it is not currently being used

Removing


AparnaKarve:
LGTM

PR #331, opened 2018-05-21
Use standard application layout

himdel:
v2v currently bakes its own application layout, which is problematic because it means v2v is currently managing all the assets for itself.

Meaning any changes in manageiq application layout can currently break v2v.

Since v2v does not need to override the menu, navbars etc, it should be using the standard manageiq layout, and only affecting `center_div`.

This PR does just that.
Depends on https://github.com/ManageIQ/manageiq-ui-classic/pull/3965  for `<title>` to work right (merged, not backported yet)

https://github.com/ManageIQ/manageiq-ui-classic/issues/3963: needed to be compatible with both master and gaprindashvili


himdel:
Note that this may or may not be 100% there: looks like we would need to move the content one layer up to get rid of extra margins.

If that's a concern, we can wait for an ops UI PR to add a layout without margins.

priley86:
thanks! will test this shortly...

priley86:
@himdel yes, we will need the added margins it seems...would like to maintain this UI design if possible.

himdel:
Makes sense, on it today :)

---

Before (master, master):

before



After (cpr, #331):

TODO

himdel:
@priley86 Ok, with https://github.com/ManageIQ/manageiq-ui-classic/pull/3970 this should be working again :)

Bonus: we're using `<body>` now ;)

priley86:
ok testing this now...

priley86:
this looks good to me - tested the layout is applying the same way as before.

@Fryguy - you may want to incorporate this into your branch...

priley86:
@AparnaKarve can you confirm?

AparnaKarve:
@himdel @priley86 The Migration screen does not seem to be rendering well for me.

screen shot 2018-05-22 at 12 48 45 pm

Also seeing an error in the browser console -
screen shot 2018-05-22 at 12 55 09 pm

@himdel Can you try again at your end?

priley86:
hmm - i reupdated core miq, ui-classic and targeted this branch...can you try again?

AparnaKarve:
@himdel @priley86 Ignore my above comment... :) and sorry for the noise

I rebased ui-classic and things are back to normal now.

Layout looks good! (like it was before)

PR #334, opened 2018-05-21
Move componentRegistry, mount, mount_react_component to v2v

himdel:
Not being used in ui-classic, there's a new registry being rolled out now,
less interdependencies if we just keep it here.

https://github.com/ManageIQ/manageiq-ui-classic/issues/3963: needed because no registry in gaprindashvili ui-classic

(This PR is based on #331, merge that one first please)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
   globals: {
     $: true,
     __: true,
-    sprintf: true,
+    sprintf: true
   },
   extends: [
     'standard',
--- a/app/controllers/migration_controller.rb
+++ b/app/controllers/migration_controller.rb
@@ -14,6 +14,12 @@ class MigrationController < ApplicationController
     def layout_full_center
       "layouts/full_center_v2v"
     end
+
+    def mount_react_component(name, selector, data = [])
+      javascript_tag(:defer => 'defer') do
+        "$(v2v.mount('#{name}', '#{selector}', #{data.to_json}));".html_safe
+      end
+    end
   end
 
   menu_section :migration
--- a/app/javascript/components/componentRegistry.js
+++ b/app/javascript/components/componentRegistry.js
@@ -1,33 +1,64 @@
 import React from 'react';
 import { i18nProviderWrapperFactory } from '../common/i18nProviderWrapperFactory';
+import forEach from 'lodash/forEach';
+import map from 'lodash/map';
 
-const { componentRegistry } = window.ManageIQ.react;
+const componentRegistry = {
+  registry: {},
 
-// extends current MIQ componentRegistry with i18nProviderWrapper
-componentRegistry.markup = (name, data, store) => {
-  const currentComponent = componentRegistry.getComponent(name);
+  register({ name = null, type = null, store = true, data = true }) {
+    if (!name || !type) {
+      throw new Error('Component name or type is missing');
+    }
+    if (this.registry[name]) {
+      throw new Error(`Component name already taken: ${name}`);
+    }
 
-  if (!currentComponent) {
-    throw new Error(
-      `Component not found:  ${name} among ${this.registeredComponents()}`
+    this.registry[name] = { type, store, data };
+    return this.registry;
+  },
+
+  registerMultiple(componentObjs) {
+    return forEach(componentObjs, obj => this.register(obj));
+  },
+
+  getComponent(name) {
+    return this.registry[name];
+  },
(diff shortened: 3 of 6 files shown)


priley86:
will await #331 resolution

package.json
```diff
@@ -64,6 +64,8 @@
     "file-saver": "^1.3.8",
     "intl": "~1.2.5",
     "jquery": "~3.2.1",
+    "lodash.foreach": "^4.5.0",
+    "lodash.map": "^4.6.0",
```
priley86: is there any reason why to continue using lodash map and lodash foreach? I believe there are es6 compatible versions of those methods.. 

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_each
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_map

priley86: orderby is not so simple though and used by our sortabular/table sorting component ;(

himdel:
Fixed the lodash imports, but rebased on top of #331 so I'll have to rebase again once that one is in :).

priley86:
merged #331  - just needs some minor eslint fixes it appears...otherwise LGTM

priley86:
ok... now #331 has been merged... rebase and some minor eslint fixes it looks like...

AparnaKarve:
@himdel So the idea behind this PR to avoid backporting a ton of other PRs in ui-classic to gaprindashvili?

priley86:
since this creates a local component registry (and no longer references `window.ManageIQ.react` - is it OK to assume that is going away or you are just doing this temporarily? If it's local to v2v, I assume it could also be removed later on.

I would like to revisit this pattern in the future if we can. I think we may be able to achieve module separation using webpack code splits / ES6 JS modules instead of the global cache. I'm still not accustom to this sort of pattern and not sure what other teams are going w/ here.

priley86:
for the short term (i.e. for upcoming releases) - i have no issue w/ this at all. Just let us know if you'd like this in `gaprindashvili` and `master`?

priley86:
tested this locally.. it works fine for me.

himdel:
So.. basically, last week, we finished setting up the new registry in ops UI - that one is now documented in [guides](https://github.com/ManageIQ/guides/blob/master/ui/register_react_component.md).

The idea is to use the new registry instead - but mostly only for the root component - essentially any time you need to render a react component in non-react context.

The new registry can work with any react component directly - dropping the need for components to take only `data` and `store` as props. (And we have a `= react 'Name', {(props)}` helper to render it anywhere.)

So... I'm moving the old registry here, as right now, v2v completely depends on it right now.
Also, it is not available in gaprindashvili, so moving this here solves it.
And it won't be available in master for much longer (pretty much waiting for this PR for that, and a similar graphql change), so moving here makes sure we don't break it.


> So the idea behind this PR to avoid backporting a ton of other PRs in ui-classic to gaprindashvili?

Yup, a ton :). And not to break it when we remove it.


> is it OK to assume that component registry is going away or you are just doing this temporarily? If it's local to v2v, I assume it could also be removed later on.

Yup, the idea behind it being local is that you're free to change it to something more standard :). (And you'll know you can't break any other plugin by doing that.)


> Just let us know if you'd like this in gaprindashvili and master?

Both gaprindashvili and master please :).

priley86:
> The idea is to use the new registry instead - but mostly only for the root component - essentially any time you need to render a react component in non-react context.

Ok, this is making more sense now. From an MIQ standpoint supporting multiple frontend frameworks, I can see the case for this on adding React support to the legacy Ng pages. This is not really necessary in a React only architecture I suppose.

> Yup, the idea behind it being local is that you're free to change it to something more standard (or not, up to you) :). (And you'll know you can't break any other plugin by doing that.)

With a lot of product teams moving to consuming PatternFly React and those components becoming standardized and more scalable directly from PF, I hope this need diminishes, but I see your point here. I tend to prefer consuming directly from PF for new React pages unless we are extending a component for a specific need across MIQ. In that case we have the [ManageIQ/react-ui-components](https://github.com/ManageIQ/react-ui-components)  repo. I.m.o. we should try to make the upstream PF React components as flexible as possible first though.

> Both gaprindashvili and master please :).
Not a problem. I would like to reconsider this topic post release. Thanks for your patience.

priley86:
@AparnaKarve do you have any more concerns before merge?

AparnaKarve:
Verified locally that these changes work.
Merging...

PR #335, opened 2018-05-21
BootstrapSelect - use global jQuery

himdel:
to make sure selectpicker is there, even when global jQuery comes from bower
this means we now depend on global jquery + bootstrap-select, but works in master and gaprindashvili

---

Fixes

	miq_debug.self-a72dc0b9f690f63b4688fbca3774709918c11fdfb7b5fc454690d11d007fc88b.js?body=1:29 TypeError: m(...).selectpicker is not a function
		at t.value (BootstrapSelect.js:29)
		at commitLifeCycles (react-dom.production.min.js:159)

https://github.com/ManageIQ/manageiq-ui-classic/issues/3963: needed because otherwise, we need to backport all the changes to support moving ui-classic dependencies from bower to npm. If global jQuery comes from bower, any version of jQuery from npm will not see bootstrap-select.

(This PR *could* be gaprindashvili-only if so desired.)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,8 +4,9 @@ module.exports = {
     'jest/globals': true
   },
   globals: {
+    $: true,
     __: true,
-    sprintf: true
+    sprintf: true,
   },
   extends: [
     'standard',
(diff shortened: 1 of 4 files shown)


priley86:
@himdel if you can rebase this w/ latest `master` and update `yarn.lock` and add `$` to globals, this PR is fine to merge...

just make this change to `.eslintrc.js` and tests will pass:
```
module.exports = {
  env: {
    browser: true,
    'jest/globals': true
  },
  globals: {
    __: true,
    sprintf: true,
    $: true
  },
```

himdel:
@priley86 Thanks, should be ready now :)

priley86:
thanks!

PR #336, opened 2018-05-21
Remove unused npm deps

himdel:
Axios used to be used for API, now only used in specs (moved to devDependencies).

Cx is unused (but some places import classnames as cx, possibly the source of confusion)

Intl is unused (not react-intl)

patternfly and react-bootstrap have no direct dependencies here - unless they were to specify a version constraint?


priley86:
re axios: agree, that was missed

re Cx: agree, we are only using `classnames`

re: Intl, agree - it is no longer used.

patternfly-react now depends on react-bootstrap as a direct dependency so i think it ok to remove. patternfly is coming from ui-classic.

I am planning on updating patternfly-react to the latest soon...i am guessing we should update that in ui-classic as well.

AparnaKarve:
Thanks for testing and confirming @priley86 

Thanks for the PR @himdel

himdel:
> patternfly-react now depends on react-bootstrap as a direct dependency so i think it ok to remove. patternfly is coming from ui-classic.

> I am planning on updating patternfly-react to the latest soon...i am guessing we should update that in ui-classic as well.

Agreed, we should try to keep both on the same version.

@priley86 Do you happen to know, does v2v actually depend on that version of patternfly?

Gaprindashvili ui-classic is using 3.23.1, so if we don't *have* to backport newer versions, would be nice :)

If we need it, it should be doable, but there's been at least 4 PRs fixing breakages from new versions which were not backported in gaprindashvili (https://github.com/ManageIQ/manageiq-ui-classic/pull/3041, https://github.com/ManageIQ/manageiq-ui-classic/pull/3571, https://github.com/ManageIQ/manageiq-ui-classic/pull/3780, https://github.com/ManageIQ/manageiq-ui-classic/pull/3940, though one of them may be related to npm more than patternfly versions).

PR #339, opened 2018-05-21
Start polling on the plan page based on more reliable status data

### LINKED ISSUE #311: Migration Details Page Should Auto-Refresh
https://bugzilla.redhat.com/show_bug.cgi?id=1573625
Since we have a spinner on the page which gives false impression of page will be automatically refreshed, but it does not.

mturley:
There appears to be some condition where the polling does not start on this page. I loosened the logic here in case we have a new status string, or race condition, in Kedar's environment that is preventing the polling from starting.

I'm hoping this will fix #311 but I need help figuring out how to test it against @kedark3's environment.

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -65,11 +65,11 @@ class Plan extends React.Component {
         const mostRecentRequest = getMostRecentRequest(miq_requests);
         const planRequestId = mostRecentRequest.id;
         fetchPlanRequestAction(fetchPlanRequestUrl, planRequestId);
-        if (mostRecentRequest.status === 'active') {
-          this.startPolling(planRequestId);
-        } else if (
-          mostRecentRequest.request_state === 'finished' ||
-          mostRecentRequest.status === 'Error'
+        this.startPolling(planRequestId);
+        if (
+          mostRecentRequest.status !== 'active' &&
+          (mostRecentRequest.request_state === 'finished' ||
+            mostRecentRequest.status === 'Error')
         ) {
           this.setState(() => ({
             planFinished: true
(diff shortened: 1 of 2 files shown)


michaelkro:
Seeing polling for a "completed migration"

app/javascript/react/screens/App/Plan/Plan.js
```diff
@@ -65,11 +65,11 @@ class Plan extends React.Component {
         const mostRecentRequest = getMostRecentRequest(miq_requests);
         const planRequestId = mostRecentRequest.id;
         fetchPlanRequestAction(fetchPlanRequestUrl, planRequestId);
-        if (mostRecentRequest.status === 'active') {
-          this.startPolling(planRequestId);
-        } else if (
-          mostRecentRequest.request_state === 'finished' ||
-          mostRecentRequest.status === 'Error'
+        this.startPolling(planRequestId);
```
AparnaKarve: @mturley Calling `this.startPolling(planRequestId)` was a good call, because it does prove that if polling is called regardless of the value in `mostRecentRequest.status`, then the page does refresh -- I confirmed this on an appliance.

I think this whole if/else block can be replaced and simplified by this --
```js
if (mostRecentRequest.fulfilled_on === null) {
  this.startPolling(planRequestId);
} else {
  this.setState(() => ({
    planFinished: true
  }));
}

```

`fulfilled_on` being `null` is a reliable indicator of the request being in it's in-progress state.

AparnaKarve: Will test this some more tomorrow. Thanks!

mturley: Scratch that, if we have `mostRecentRequest.id` we will have `mostRecentRequest.fulfilled_on`. I'll go with your version @AparnaKarve

michaelkro:
@AparnaKarve Looks good to go!

JPrause:
@miq-bot add_label blocker

JPrause:
@priley86 @AparnaKarve it appears this needs to be baclported to gaprindashvili to resolve: https://bugzilla.redhat.com/show_bug.cgi?id=1588577
Can you add the gaprindashvili/yes label if this can be backported.

AllenBW:
@JPrause it already exists in Gaprindashvili https://github.com/ManageIQ/miq_v2v_ui_plugin/commits/gaprindashvili?after=3d713ff775f1688e87af32260d6dbef6b76eb32e+69

JPrause:
Thanks Allen,..my bad.

PR #340, opened 2018-05-22
Use http, not API for querying rails controller endpoints

himdel:
`API.get` does not send the authentication cookie, only the api token.
So MigationLogController can't see the session.

We now have `http` for that (from https://github.com/ManageIQ/manageiq-ui-classic/pull/3662)

Fixes https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/333#issuecomment-391052384

https://github.com/ManageIQ/manageiq-ui-classic/issues/3963: need to backport the http change too for gapri

### DIFF
--- a/app/javascript/common/http.js
+++ b/app/javascript/common/http.js
@@ -1,0 +1,17 @@
+const { http } = window;
+
+export default {
+  get(url, headers = {}, params = {}) {
+    return http.get(url, {
+      transformResponse: e => ({ data: e }),
+      headers,
+      params
+    });
+  },
+  post(url, data = {}, headers = {}) {
+    return http.post(url, data, {
+      headers,
+      transformResponse: e => ({ data: e })
+    });
+  }
+};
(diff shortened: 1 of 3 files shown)


himdel:
WIP - looks like `http` is not sending the cookie right

AparnaKarve:
Looks great! 
GTG when the classic UI PR is merged.

app/javascript/common/http.js
```diff
@@ -0,0 +1,17 @@
+const http = window.http; // eslint-disable-line prefer-destructuring
```
priley86: this can be written as `const { http } = window;` for es6 destructuring (does the same thing)

priley86:
other than this...lgtm

app/javascript/common/http.js
himdel: ok, fixing in API.js as well :)

himdel:
Removed wip, fixed the destructuring, ops UI PR is merged :)

PR #344, opened 2018-05-23
Rename to manageiq-v2v

Fryguy:
@priley86 Opened my branch as a WIP PR just so we can have a discussion location.


Fryguy:
Migration is a non-starter because ActiveRecord already owns ActiveRecord::Migration .  So, you will run into a ton of complications when creating database migrations, not to mention the terminology conflict and confusion that will ensure with all of the Rails developers on the team.

Fryguy:
@priley86 I roughly saw this discussion the other day 😄 

A: Migrations broke something.
B: What?  I just ran it and it works for me.
A: No, if you start the app, you get this weird error.
B: Did you run the migration?
A: Yeah, and it's not working.
B: What do you mean it's not working?
A: I start the app, and it gives me an error about some javascript files not available
B: What are you talking about? There are no javascript files with a database migration!
A: What?! Nooo... not a database migration!  I mean the migrations tab in the UI.
B: Ohhhhhh....that's not confusing at all </sarcasm>

(Of course it didn't go exactly this way, but it demonstrates my point 😜 )

priley86:
i feel your pain...

NickLaMuro:
@Fryguy Just an FYI, I have opened up https://github.com/ManageIQ/manageiq-ui-classic/pull/3986 as a way to solve the inflector issue you were having when compiling assets.  Should allow the javascript file to be generated as `manageiq-v2v.js` now with the inflector loaded in from this repo.

app/javascript/components/index.js
```diff
@@ -144,7 +144,7 @@ export const coreComponents = [
     data: true,
     store: false
   },
-  { name: 'v2v_ui_plugin', type: MiqV2vUi }
+  { name: 'manageiq-v2v', type: ManageIQV2V }
```
himdel: This bit should just be the name of the react component, and the `name` only needs to match the `mount_react_component` invocation arg.

Isn't it confusing to call the root component the same as the repo? To me, something like `v2v-root` would sound a bit saner..

(But no strong feelings about it, just saying :).)

Fryguy: 🤷‍♂️  I don't understand any of this, tbh.  I'll pick whatever string you tell me to :smile:

himdel: It's fine as it is, we can always rename later :)

priley86:
rebase? I have no issues w/ this.

AllenBW:
This is now g2g @Fryguy @Fryguy @himdel, jumped in, we're workin'

Fryguy:
Thank you!!

Fryguy:
@AllenBW Can you squash?  I don't like that `git pause` commit in there :)

AllenBW:
Yah big 👶 😏

carbonin:
@Fryguy is there a manageiq PR to rename this in the Gemfile? Looks like this is causing bundler failures:

```
Could not find gem 'miq_v2v_ui' in https://github.com/ManageIQ/miq_v2v_ui_plugin.git (at master@33b0c42).
The source does not contain any versions of 'miq_v2v_ui'
```

carbonin:
Or are you waiting to also rename the whole repo?

himdel:
Yes, can we also rename the repo?

It's always been annoying that the gem name is not the same as the repo name, so maaybe it doesn't have to anymore?

mturley:
+1 for the repo name and the gem name matching, whatever name that is i don't really have an opinion :)

Issue #310, opened 2018-05-14
Enhance infrastructure migration info


bthurber:
Add storage and network mapping info into infrastructure mapping details post creation.

bthurber:
@vconzola since the clusters and datastores have the same info could we combine them for a less complex view?  Maybe just call it Source/Target Clusters and Datastores.

vconzola:
@bthurber Not sure what's complex about this view, but we can get rid of the cluster-cluster mapping section since cluster mappings are contained within the datastore mappings. Is that what you're suggesting?

priley86:
I'm currently not getting back the datastore names/network names in the same API call, but I do get back the associated plan names.

http://0.0.0.0:8080/api/transformation_mappings?expand=resources&attributes=transformation_mapping_items,service_templates
```
{
  "name": "transformation_mappings",
  "count": 2,
  "subcount": 2,
  "pages": 1,
  "resources": [
    {
      "href": "http://0.0.0.0:8080/api/transformation_mappings/1",
      "id": "1",
      "name": "singleVM",
      "description": null,
      "comments": null,
      "state": "draft",
      "options": {
        
      },
      "tenant_id": null,
      "created_at": "2018-05-04T19:00:24Z",
      "updated_at": "2018-05-04T19:00:24Z",
      "transformation_mapping_items": [
        {
          "id": "1",
          "source_id": "3",
          "source_type": "EmsCluster",
          "destination_id": "2",
          "destination_type": "EmsCluster",
          "transformation_mapping_id": "1",
          "options": {
            
          },
          "created_at": "2018-05-04T19:00:24Z",
          "updated_at": "2018-05-04T19:00:24Z"
        },
        {
          "id": "2",
          "source_id": "10",
          "source_type": "Storage",
          "destination_id": "9",
          "destination_type": "Storage",
          "transformation_mapping_id": "1",
          "options": {
            
          },
          "created_at": "2018-05-04T19:00:24Z",
          "updated_at": "2018-05-04T19:00:24Z"
        },
        {
          "id": "3",
          "source_id": "29",
          "source_type": "Lan",
          "destination_id": "4",
          "destination_type": "Lan",
          "transformation_mapping_id": "1",
          "options": {
            
          },
          "created_at": "2018-05-04T19:00:24Z",
          "updated_at": "2018-05-04T19:00:24Z"
        },
        {
          "id": "4",
          "source_id": "42",
          "source_type": "Lan",
[...]

PR #346, opened 2018-05-24
associated plans

### LINKED ISSUE #321: Show associated plans in a infrastructure mapping
@priley86 @vconzola can you add more detail please :)

priley86:
Closes #321 

* adds the associated plans attribute to the transformation mapping query so they are also returned in that query
* displays the associated plans count in the list view item and the associated plan names in the list view detail. Display nothing if there are no associated plans.
* correctly pluralizes "Source Cluster(s)", "Target Cluster(s)", and "Associated Plan(s)" in the list item count - previously that was not done...

Would like to iterate on #310 next after we have API direction.

screen shot 2018-05-24 at 10 49 12 am

### DIFF
--- a/app/javascript/components/index.js
+++ b/app/javascript/components/index.js
@@ -90,7 +90,7 @@ export const coreComponents = [
     data: {
       fetchTransformationMappingsUrl:
         'api/transformation_mappings?expand=resources' +
-        '&attributes=transformation_mapping_items',
+        '&attributes=transformation_mapping_items,service_templates',
       fetchTransformationPlansUrl:
         '/api/service_templates/?' +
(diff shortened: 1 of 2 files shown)


app/javascript/components/index.js
```diff
@@ -90,7 +90,7 @@ export const coreComponents = [
     data: {
       fetchTransformationMappingsUrl:
         'api/transformation_mappings?expand=resources' +
-        '&attributes=transformation_mapping_items',
+        '&attributes=transformation_mapping_items,service_templates',
```
AparnaKarve: Nice! I needed this change for deleting transformation mappings as well!

AparnaKarve:
Tested and looks great!

PR #350, opened 2018-05-25
Pass DualPaneMapperCount as a prop

michaelkro:
Production builds seem to have an issue filtering the children of
DualPaneMapperList to separate out the counter. To avoid this all
together, just pass in the counter component as a prop and let all the
children be DualPaneMapperListItem(s)

## Notes
1. Closes https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/348
2. As this does not manifest in development, but only in production builds, we will have to wait until this change is incorporated into the nightlies to verify.  To test this locally, go through the Mapping Wizard and verify that nothing has actually changed
3. In the following screenshot you can see that the count component is being rendered alongside the list items (spun up a summit demo to see this)
    developer_tools_-_https___1cloudforms-rhpdsrhmicroredhat-aceruhmj_srv_ravcloud_com_migration
  which leads me to believe that this [variable](https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/DualPaneMapper/DualPaneMapperList.js#L9) is always returning null for some reason

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/DualPaneMapper/DualPaneMapperList.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/DualPaneMapper/DualPaneMapperList.js
@@ -3,17 +3,7 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { Spinner } from 'patternfly-react';
 
-const DualPaneMapperList = ({ children, listTitle, loading, id }) => {
-  const childrenArray = React.Children.toArray(children);
-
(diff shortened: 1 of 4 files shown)


priley86:
@michaelkro were you able to confirm this in the QA environment?

michaelkro:
We'll have to wait until the next nightly build to verify the fix.  The current implementation works fine locally, but for some reason, things get weird in prod and some filtering isn't working

priley86:
ok sounds good @michaelkro

Issue #343, opened 2018-05-23
Sort value changing after clicking header of discovered vms table


kedark3:
https://bugzilla.redhat.com/show_bug.cgi?id=1573963

priley86:
i believe i have a fix for this - planning to update PF React dependency and we'll see what that does...

JPrause:
@AllenBW can you add the blocker label for me.

AllenBW:
@JPrause...

Never gonna give you up
Never gonna let you down
Never gonna run around and desert you
Never gonna make you cry
Never gonna say goodbye
Never gonna tell a lie and hurt you

unlike... say... @miq-bot 😏 🔪

priley86:
yes - pf-react dependency update went in. also waiting on nightly feedback on this one. 

@kedark3 if you are still seeing this issue, can you please report back which browsers (user agent string would be great).

@AllenBW @JPrause not sure I understand what this is blocking, but hopefully it is fixed soon.

bthurber:
Appears to be fixed in latest nightly.

PR #351, opened 2018-05-25
update patternfly-react, remove bindMethods

priley86:
I believe this should address #343 but will wait for QA/nightly to confirm this.

There was a previous issue w/ tables not rendering in production webpack builds addressed in PF React recently. See:
https://github.com/patternfly/patternfly-react/pull/357

* removed `bindMethods` - this is no longer suggested upstream and we are suggesting the use of Class Properties instead.


priley86:
rebased #350

app/javascript/react/screens/App/Overview/components/ModalWizard/ModalWizardStateProvider.js
```diff
@@ -1,31 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { bindMethods } from 'patternfly-react';
 
 class ModalWizardStateProvider extends React.Component {
   constructor() {
     super();
     this.state = { activeStepIndex: 0 };
```
michaelkro: should we get rid of the constructor here and use the class properties syntax for the state?

priley86: sure...happy to move to the `state = {}` syntax too... just a few!

priley86: updated... i think this syntax is cleaner...except when you are using props to initialize state.

PR #356, opened 2018-05-30
Delete Transformation Mappings feature

AparnaKarve:
Added support for deleting transformation mappings

----------------------------

`pf delete` icon added in the `actions` area in the list 
(shows disabled and enabled delete icons)
screen shot 2018-05-29 at 7 13 04 pm

----------------------------
Simple delete confirmation dialog for a mapping -
screen shot 2018-05-29 at 7 13 35 pm

----------------------------

Delete confirmation dialog that shows associated Plans that are not started and/or
associated plans in error -
screen shot 2018-05-29 at 7 13 54 pm

----------------------------


AparnaKarve:
@vconzola The second Delete Confirmation modal above shows the affected Plans.

Do you think we should also delete these Plans --  since there is no point in keeping the plans around without their transformation mapping?

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -277,7 +357,51 @@ class Overview extends React.Component {
             clusters={clusters}
             transformationMappings={transformationMappings}
             createInfraMappingClick={showMappingWizardAction}
+            inProgressRequestsTransformationMappings={inProgressRequestsTransformationMappings()}
+            showDeleteConfirmationModalAction={
+              showDeleteConfirmationModalAction
+            }
+            setMappingToDeleteAction={setMappingToDeleteAction}
           />
+          <Modal
```
priley86: would you mind moving this Modal to another stateless functional component? i.e. `components/DeleteMappingConfirmationModal` ? I think extracting this Modal JSX and relevant logic there in another component would be nice...

Will try to test this out on my side today...

vconzola:
@AparnaKarve We need to keep the plans so users can go back and see which VMs have been migrated. If users want to then archive the plans they can do that separately. That's the direction I've gotten from Brett and Marco. Your question does bring up an interesting point, however. We need to decide what to display for plans that no longer have an associated mapping. I'll create an issue for this.

himdel:
(Depends on https://github.com/ManageIQ/manageiq-api/pull/383 (in case this goes to gaprindashvili))

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -27,151 +28,206 @@ function clusterName(clusters, clusterId) {
...
-              href="#"
-              onClick={e => {
-                e.preventDefault();
-                createInfraMappingClick();
-              }}
-            >
-              <Icon type="pf" name="add-circle-o" />
-              &nbsp;{__('Create Infrastructure Mapping')}
-            </a>
+  <div>
```
priley86: i think we can just use `React.Fragment` instead of introducing the additional <div> ? i.e. just wrap <Grid.Col> and <DeleteInfrastructureMappingConfirmationModal> w/ <React.Fragment>

priley86:
thanks for the updates @AparnaKarve ! looks good...just one other minor nit...

@himdel looks like this API was merged... i will update locally and test this first thing tomorrow.

priley86:
confirmed this works locally. merging.

PR #358, opened 2018-05-30
Prevent opening migration wizards when user has insufficient providers

### LINKED ISSUE #163: Minor UX edge case when no objects in database
I haven't set up a provider yet, thus I don't have any clusters, networks, nor datastores.  However, I'm able to go to the Migration screen and start the wizard, only to get to the cluster page with nothing there.  We should probably prevent access to the wizard, or perhaps the entire migration tab when there are no providers defined.  Or perhaps, more specifically, if there are no providers defined for the migration types we care about.

Does it make sense to even show v2v if there is only 1 provider?  Can the source and destination be on the same provider?  If not, then substitute "no providers" with "less than 2 providers" up above.

mturley:
Fixes #163 🎊


app/javascript/react/screens/App/Overview/OverviewConstants.js
```diff
@@ -10,6 +10,10 @@ export const CONTINUE_TO_PLAN = 'CONTINUE_TO_PLAN';
 export const V2V_SET_MIGRATIONS_FILTER = 'V2V_SET_MIGRATIONS_FILTER';
 export const V2V_RETRY_MIGRATION = 'V2V_RETRY_MIGRATION';
 
+export const FETCH_PROVIDERS = 'FETCH_PROVIDERS';
+export const FETCH_PROVIDERS_URL =
+  '/api/providers?expand=resources&attributes=type&filter[]=or%20type=ManageIQ::Providers::Redhat::InfraManager&filter[]=type=ManageIQ::Providers::Vmware::InfraManager';
```
priley86: are we sure we want to filter providers by these two types? Just double checking ;)

It seems like for now this is the way to go...

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -273,11 +279,20 @@ class Overview extends React.Component {
...
-          />
+          {hasSufficientProviders ? (
+            <InfrastructureMappingsList
+              clusters={clusters}
+              transformationMappings={transformationMappings}
+              createInfraMappingClick={showMappingWizardAction}
+            />
+          ) : (
+            <OverviewEmptyState
+              description="The VMWare and Red Hat Virtualization providers must be configured before attempting a migration."
```
priley86: for i18n , you can add `{__("The Vmware...")}`

priley86:
this looks mostly good ! Just a few travis things and minor additions... can you add a screenshot for us? 

great job!

app/javascript/react/screens/App/Overview/OverviewConstants.js
mturley: That's what I got from https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/163 and slack discussions. Right now the only migration path we have is VMware <--> RHV, so those two specific providers are required. If you remove the `&filter` part of this URL, the rest of the code will work fine, it'll just be looping over a larger array when it checks for those types. So we can make this more generic if we need more provider data later.

app/javascript/react/screens/App/Overview/Overview.js
mturley: Good catch, it was because I copied and pasted another OverviewEmptyState that wasn't i18n-ized.. i'll fix those ones real quick

mturley:
Whoops, took this screenshot last night but forgot to post it here!

screen s hot 2018-05-30 at 4 28 30 am

mturley:
@priley86 95e1f15 fixes the failing unit tests. I can add some new ones if you like, too, or we can go ahead and merge this.

app/javascript/react/screens/App/Overview/OverviewConstants.js
mturley: @priley86, also, i should point out that this filter URL came directly from @AparnaKarve, so that's why I did it this way.

mturley:
0b46497 fixes the button, which I had linking to the wrong page. This PR should be all good to go now.

mturley:
0b46497 fixes the button, which I had linking to the wrong page. This PR should be all good to go now, unless @vconzola you think the button needs to do anything else besides just take you to "Compute > Infrastructure > Providers".

mturley:
This is ready to merge, IMO, if the button needs to be changed we can do it in another PR

app/javascript/react/screens/App/Overview/OverviewConstants.js
AparnaKarve: The above API gives us an output that is focussed on a very specific section of the `providers` API - the `type` of the provider. Also it filters out only the RHV and VMWare providers.

Like @mturley said, we can always modify this API to give us a more generic Provider API output (which will result in a larger json), but for now the above works.

AparnaKarve:
@mturley This looks good for the most part.

I noticed that the polling still happens even if we do not have the required providers - would be great if we polled *only* if we had the required set of providers.

Feel free to address this in a follow-up PR if necessary. Thanks.

app/javascript/react/screens/App/Overview/OverviewConstants.js
priley86: great! thanks.

app/javascript/react/screens/App/Overview/OverviewActions.js
```diff
@@ -27,6 +29,19 @@ export const showPlanWizardAction = id => dispatch => {
   });
 };
 
+export const fetchProvidersAction = () => dispatch => {
+  dispatch({
+    type: FETCH_PROVIDERS,
+    payload: new Promise((resolve, reject) => {
```
priley86: nit: i think we can simplify this to:
```
export const fetchProvidersAction = () => dispatch => {
  dispatch({
    type: FETCH_PROVIDERS,
    payload: API.get(FETCH_PROVIDERS_URL)
  });
};
```

priley86:
tested this locally...working great!

Just needs a rebase and one nit if you are OK w/ that one.

app/javascript/react/screens/App/Overview/OverviewActions.js
priley86: (does the same thing w/ redux-promise-middleware as it will auto resolve/reject

mturley:
Shoot, I forgot to rebase this earlier. Should be ready to go @priley86 !

PR #359, opened 2018-05-30
Fixes the drilldown issue on NotStarted Plans

AparnaKarve:
A drill-down in Not Started Plans is currently throwing the below error -

TypeError: Cannot read property '0' of undefined ==> which seems to be coming from PlanVmsList.js:23

The PR fixes the above error

/cc @priley86


priley86:
ahh thanks! yes, i missed this in the bindMethods move...i agree @AparnaKarve ... I think you can also use the keyword `static` to declare static class properties since we are using this transform-class-properties. This works the same way though! Merging...

https://babeljs.io/docs/plugins/transform-class-properties/

PR #364, opened 2018-05-31
Depend on transform-es2015-modules-commonjs, mentioned in .babelrc

himdel:
Our `.babelrc` uses:

```
  "env": {
    "test": {
      "plugins": ["transform-es2015-modules-commonjs"]
    }
```

But `transform-es2015-modules-commonjs` is never mentioned in `package.json`.

Making sure it is :).


himdel:
Uh, yeah, sorry about that, my defaults are not sui or v2v defaults :)

Updating to `^` :)

Issue #316, opened 2018-05-16
Drill into an Archived Migration Plan


serenamarie125:
Once archived migration plans are shown in the main Migrations page, allow the user to drill in and see the plan details.

Depends on #315

michaelkro:
Closed by #367

Issue #315, opened 2018-05-16
View Archived Migration Plans from the main Migration page 


serenamarie125:
User should be able to view archived migration plans.  This means that we need to allow for a filter for the "Archived" state which will show these plans in a list view.

serenamarie125:
@AparnaKarve can you check if there's an API to get all Migration Plans that are "archived"?  If available, please remove the label.  Thanks!

AparnaKarve:
@serenamarie125 This will need both backend and API changes.
So the `needs-api` label is indeed applicable.

serenamarie125:
@blomquisg re: get all Migration Plans that are "archived"
Aparna mentioned that we definitely need support, just touching base to see if this is in the works, or if we need to meet to make sure we are in agreement on the requirements/dependencies.  lmk!

blomquisg:
@serenamarie125 it's not yet in progress on the back end.  Let's meet about this and #314 to make sure we have the right requirements on the back end.

serenamarie125:
@vconzola can you add mocks?

vconzola:
@serenamarie125 Here's a [link](https://docs.google.com/presentation/d/1cvVq0dKL_rXarz-HnD9mQAdcfMJ438ivjPhYMPpOFM0/edit?usp=sharing) to the mockups.

serenamarie125:
Looks like these are the associated API PRs:
https://github.com/ManageIQ/manageiq-schema/issues/207
https://github.com/ManageIQ/manageiq/issues/17480

According to Adam, you can run with those PRs applied locally (he doesn't expect to have to make major changes)

michaelkro:
Closed by #367

PR #367, opened 2018-06-01
Archive Plans

### LINKED ISSUE #314: User should be able to archive a Migration Plan
User should have access to archive a migration plan.  If a migration plan is running, the option should not be available ( should be disabled ).

We should provide a confirmation for archive - thus, the user clicks on the Archive action, but is presented with an confirmation dialog with "Cancel" and "Archive" buttons

michaelkro:
Enable archiving of completed transformation plans (successful & errored)
* Add archive button to Completed Migrations ListView
* Set up fetching so that archived and active (unarchived) plans are
  returned separately.
* Set up POST to archive a plan
* Modify MigrationsCompletedList to also be used for display of
  archived plans

ADDITIONAL CHANGES
* Refactor to remove usage of Object.values() due to node support
* Extract migrations filters strings to constants

Closes #314, #315, #316

## Notes
* To quickly unarchive all your plans, use the following command in the rails console
  ```ruby
  ServiceTemplateTransformationPlan.update_all(deleted_on: nil)
  ```
  Alternatively, use the `unarchive!` method on individual records

* Big thank you to @AparnaKarve for showing me how to set up filters when fetching plans!

### Flow
archive

### Archive Button
archive-button

### Archive Modal (Plan with unmigrated VMs)
archive-modal

### Archive List
fullscreen_6_6_18__7_57_am

### Archive Details
archived-plan-details

### Archive Empty
archived-empty-state


app/javascript/components/index.js
```diff
@@ -94,6 +94,15 @@ export const coreComponents = [
       fetchTransformationPlansUrl:
         '/api/service_templates/?' +
         "filter[]=type='ServiceTemplateTransformationPlan'" +
+        '&filter[]=deleted_on=nil' +
```
priley86: If I read @agrage's email correctly... "You shouldn't need to use the deleted on attribute directly, there is an active and an archived scope that you should be able to use."... this doesn't make sense?

@agrare can you confirm these two query filters for archived/unarchived are correctly constructed?

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -176,8 +186,11 @@ class Overview extends React.Component {
 
     createTransformationPlanRequestAction(url).then(() => {
       retryMigrationAction(planId);
-      setMigrationsFilterAction('Migration Plans in Progress');
-      fetchTransformationPlansAction(fetchTransformationPlansUrl);
+      setMigrationsFilterAction('In Progress Plans');
```
priley86: Looking at these again... it would be nice to break the "In Progress Plans", "Migration Plans in Progress", "Migration Plans Not Started" strings out into constants across the board. It seems like "In Progress Plans" is the same as "Migration Plans in Progress" ?

app/javascript/react/screens/App/Overview/components/Migrations/Migrations.js
```diff
@@ -18,20 +18,37 @@ const Migrations = ({
...
+  fetchTransformationPlansAction,
+  fetchArchivedTransformationPlansUrl,
+  archivedTransformationPlans,
+  allArchivedPlanRequestsWithTasks
 }) => {
   const filterOptions = [
-    'Migration Plans Not Started',
-    'Migration Plans in Progress',
-    'Migration Plans Completed'
+    'Not Started Plans',
```
priley86: ahh you are renaming them in an array. Can we also reference the strings as constants?

app/javascript/react/screens/App/Overview/components/Migrations/Migrations.js
```diff
@@ -81,15 +96,15 @@ const Migrations = ({
       </Grid.Col>
       {transformationPlans.length > 0 && (
         <React.Fragment>
-          {activeFilter === 'Migration Plans Not Started' && (
+          {activeFilter === 'Not Started Plans' && (
```
priley86: same here...constants preferred.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -47,6 +50,43 @@ const MigrationsCompletedList = ({
...
+            const archiveMigrationWarningText = (
+              <React.Fragment>
+                <p>
+                  {__('Are you sure you want to archive migration plan ')}
+                  <strong>{plan.name}</strong>?
+                </p>
+                {failed && (
+                  <p>
+                    {__(
+                      'This plan includes VMs that failed to migrate. If you archive the plan, you will not be able to retry the failed migrations'
```
priley86: Missing period at the end

"If you archive the plan, you will not be able to retry the failed migrations."

app/javascript/components/index.js
agrare: @priley86 I can confirm that the following works:
```
>> ServiceTemplate.active
=> #<ActiveRecord::Relation [#<ServiceTemplate id: 1, name: nil, description: nil, guid: "ffd1a950-a007-49d8-891f-9c4741c5dbb9", type: nil, service_template_id: nil, options: {}, created_at: "2018-06-01 13:07:25", updated_at: "2018-06-01 13:07:25", display: nil, evm_owner_id: nil, miq_group_id: 1, service_type: "unknown", prov_type: nil, provision_cost: nil, service_template_catalog_id: nil, long_description: nil, tenant_id: 1, generic_subtype: nil, deleted_on: nil>]>
>> ServiceTemplate.archived
=> #<ActiveRecord::Relation []>
```
I can't confirm how that should be accessed through the API

AparnaKarve: @agrare To be able to use `ServiceTemplate.archived` in an API, we would need `archived` to be a virtual column I think.
Accessing `deleted_on` directly is working well in the API, and should be OK to use it directly since there are no other complexities involved with the `archived`/ `unarchived` logic.

`deleted_on` == `nil` is `unarchived`
`deleted_on` != `nil` is `archived`

agrare: Oh really having it as a scope isn't enough?  Well if you can get it working through the API and put a PR to core in I'll merge it.

AparnaKarve: @agrare I just created this core PR - https://github.com/ManageIQ/manageiq/pull/17509

michaelkro:
@priley86, the CI failures are due to the use of `Object.values()` [here](https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js#L39)

Seems like this method isn't supported in node 6.x

I can remove my unit tests to keep CI from failing, but wasn't sure how we want to handle this

priley86:
yes i'm fine w/ this... `Object.keys` it is i guess?

michaelkro:
Our CI script runs builds using node 6 and 6.10.  `Object.values` is an ES2017 feature that is not supported by node < 6.14

Please see [node.green](https://node.green/#ES2017-features-Object-static-methods-Object-values)

priley86:
I'm not sure how important Node6 support is for v2v at this point. It's nice having the support, but doesn't seem required anymore. My guess is we should just stay in sync w/ ui-classic which appears to be `>=6.9.1` right now. 
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/package.json#L108

@AparnaKarve @AllenBW can you confirm?

AparnaKarve:
@priley86 Agree, let's stay in sync with ui-classic

app/javascript/components/index.js
AparnaKarve: @agrare Looks like we will need a PR to amend the virtual column to `archived`, 
since `archived?` works for this case -

`/api/service_templates/<id>?attributes=archived?`

but not for this case -

`/api/service_templates?filter[]=archived?=false`, what @michaelkro is using above

(There seems to be a bug with `filter` in the API... not sure how easy it is to resolve. From what I can tell it's related to `MiqExpression`)

priley86:
can you rebase #391 when you get a chance? there was a request to merge that one first just now...

michaelkro:
Yup, no worries!

priley86:
this looks good to me! will go ahead and rebase my branch locally...

@AparnaKarve are we OK to merge this now or do we need to resolve the filter issue above on the backend first?

vconzola:
@serenamarie125 I wasn't listed as a reviewer, but it looks good to me too.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -84,32 +169,48 @@ const MigrationsCompletedList = ({
...
           iconType="pf"
           iconName="info"
           description={
             <span>
-              {__(
-                'There are no existing migration plans in a Completed state.'
+              {sprintf(
+                __('There are no existing migration plans in %s state.'),
+                archived ? 'an Archived' : 'a Completed'
               )}
```
AparnaKarve: The `archived` conditional needs to decide what message to display...

So if `archived` is true, then 
```js
sprintf(__('There are no existing migration plans in an Archived state.'))
```
and likewise the other message for `archived` = false

The only reason why I'm suggesting this change is because not all languages work like English.
For e.g. in a different language, the sentence construction rules or the grammar may not be the same as English and the `%s` placeholder in that particular spot may not work.

AparnaKarve:
Other than the above comment, everything else looks just great!!

Thanks @michaelkro !!

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
michaelkro: Updated :)

I squashed my commit to help those working on the G backport

AparnaKarve: Cool! :) 
That looks good!

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -412,6 +461,16 @@ Overview.propTypes = {
...
-  hasSufficientProviders: PropTypes.bool
+  hasSufficientProviders: PropTypes.bool,
+  confirmModalVisible: PropTypes.bool,
+  confirmModalOptions: PropTypes.object,
+  showConfirmModalAction: PropTypes.func,
+  hideConfirmModalAction: PropTypes.func,
+  fetchArchivedTransformationPlansUrl: PropTypes.string,
+  archivedTransformationPlans: PropTypes.array,
+  allArchivedPlanRequestsWithTasks: PropTypes.array,
+  isFetchingArchivedTransformationPlans: PropTypes.string,
```
himdel: Was the string intentional here?

I'm seeing...

```
miq_debug.self-1351f771c2cd2fab0d83069fd3f62aadbb7effc11a01ed4942f52308793453be.js?body=1:30 Warning: Failed prop type: Invalid prop `loading` of type `string` supplied to `Spinner`, expected `boolean`.
    in Spinner (created by Overview)
    in Overview (created by Connect(Overview))
```

and given the condition...

```
        <Spinner
          loading={
            !requestsWithTasksPreviouslyFetched &&
            !this.hasMadeInitialPlansFetch &&
            (isFetchingAllRequestsWithTasks ||
              isFetchingProviders ||
              isFetchingTransformationPlans ||
              isFetchingArchivedTransformationPlans ||
              isFetchingTransformationMappings)
          }
```

When `isFetchingArchivedTransformationPlans` is the first truthy value, it gets passed to `loading`.

Not sure if you want an explicit `!! isFetchingArchivedTransformationPlans` to convert to bool, or if this should never have been a string.

Can you take a look please @michaelkro ?

michaelkro: Hi @himdel, I've been noticing that error as well but just haven't had the time to fully investigate why that particular value is a string.

In some cases, we have used IDs as flags to isolate loading state, e.g., disable the migrate button for a specific migration plan ListItem after it is clicked.

However, that being said, I will take a look and at the very least put up a PR coercing it to a bool, since that is really all we need in this case.

Thanks for taking the time to report the issue!

PR #371, opened 2018-06-01
[WIP] i18n fixes and cleanups

himdel:
```
app/javascript/react/config/config.js
4:export const links = [

app/javascript/react/config/Routes.js
4:import { links } from './config';
9:  links.map(({ path, component }) => {
```

So, we only use `links` in Routes, where we only read `path` and `component`.

So, the `text` attribute is dead now, removing.

(If `text` should not be dead, we should add gettext and use it somewhere.)

---

Plus various other i18n fixes from comments

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -176,7 +176,7 @@ class Overview extends React.Component {
 
     createTransformationPlanRequestAction(url).then(() => {
       retryMigrationAction(planId);
-      setMigrationsFilterAction('Migration Plans in Progress');
+      setMigrationsFilterAction(__('Migration Plans in Progress'));
       fetchTransformationPlansAction(fetchTransformationPlansUrl);
     });
   };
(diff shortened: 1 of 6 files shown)


priley86:
this is fine as we aren't using the `text` attributes... and it looks like we missed one of the i18n references in the Breadcrumbs. Can you also update this for us?

```
    const toolbarContent = (
      <Toolbar>
        <Breadcrumb.Item href="/dashboard/maintab?tab=compute">
          Compute
        </Breadcrumb.Item>
        <Breadcrumb.Item active>Migration</Breadcrumb.Item>
      </Toolbar>
    );
```
?

himdel:
Sure, added :) and also added ActiveTransformationPlans for good measure :)

himdel:
...and fixed the rest of setMigrationsFilterAction :)

app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -176,7 +176,7 @@ class Overview extends React.Component {
 
     createTransformationPlanRequestAction(url).then(() => {
       retryMigrationAction(planId);
-      setMigrationsFilterAction('Migration Plans in Progress');
+      setMigrationsFilterAction(__('Migration Plans in Progress'));
```
priley86: can we hold this until #367 is done? we are already using `%sprintf` for those i believe...

priley86: just the breadcrumbs for now would be great...

himdel:
Sorry.. this was meant to be a 5 minute thing
Feel free to pick commits, or ping me when I should rebase.

I'm back to the build stuff.

himdel:
But agreed this should **not** be merged in this form:

https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/components/Migrations/Migrations.js#L84 is ....horrible

adding WIP

priley86:
yes let's clean that up in #367 - thanks...

priley86:
repurposed in #381 ... closing.

PR #379, opened 2018-06-01
Emptystate.title is a required field, cannot be undefied, ie-empty

### LINKED ISSUE #378: Migration Plan Wizard import csv empty title error
screen shot 2018-06-01 at 12 44 09 pm

Ends up `<EmptyState.Title />` (a title sans children) makes for a warning that is very unbecoming.

AllenBW:
closes #378

No visual changes, just rids ourselves of that nasty warning... 
screen shot 2018-06-01 at 2 25 01 pm

But also... maybe... this should be propagated back to pfr? In the form of consistent spacing of both the an empty state body and title from the icon...


priley86:
@vconzola - my vote would be to add a real title (like the empty state in PF). Do you have any suggestions for us? Otherwise, this one-off fix will have to do...

thanks @AllenBW !!

vconzola:
@priley86 @AllenBW  How about "Import File". Short and sweet.

PR #391, opened 2018-06-05
Add missing gettext

mzazrivec:
`gaprindashvili/yes`


priley86:
@mzazrivec - thanks... i think we are still addressing a few of these in #367  - will follow up in this PR after that one...

martinpovolny:
What's blocking this one from being merged?

martinpovolny:
Thank you!

PR #393, opened 2018-06-06
Add status icon to breadcrumb bar in Plan page

### LINKED ISSUE #347: Add the migration plan status icon in the breadcrumbs in Migration Plan Details page
Currently when you drill into a migration plan, you don't know if the plan is in progress/completed/migrated. 

Add a icon before the migration plan name in the breadcrumb to indicate it's status

@vconzola can you include the actual font icon names for each of the migration plan states?

mturley:
I need help testing this one with an "archived" plan. I took a guess that the `request_state` would equal "archived" instead of "finished", but I need someone to help me find the real value there. @AparnaKarve maybe you can help me there?  (i promise i'll stop bugging you soon 😂 )

Also, as implemented, there will be no icon in the breadcrumb bar if the plan is not started or in progress, only if it is completed or archived. Not sure if that's what we wanted, and if not, what should we use for those other icons @vconzola @serenamarie125 ?

Once the above is addressed, this closes #347 .

Note also: I hate that I have both `planRequestState` and `planRequestStatus` passed down.. they are named after where they come from in the API, but seen next to each other they are a little confusing. @priley86 i'm open to redux property name suggestions, haha. We may not need planRequestState anyway, if the archived state is inferred from elsewhere.

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Immutable from 'seamless-immutable';
 import { Link } from 'react-router-dom';
-import { Breadcrumb, Spinner } from 'patternfly-react';
+import { Breadcrumb, Spinner, Icon } from 'patternfly-react';
 import Toolbar from '../../../config/Toolbar';
 import PlanRequestDetailList from './components/PlanRequestDetailList/PlanRequestDetailList';
 import PlanVmsList from './components/PlanVmsList';
@@ -99,6 +99,8 @@ class Plan extends React.Component {
   render() {
     const {
       planName,
+      planArchived,
+      planRequestFailed,
       isRejectedPlanRequest,
       isFetchingPlanRequest,
       isRejectedPlan,
@@ -117,6 +119,19 @@ class Plan extends React.Component {
       planRequestPreviouslyFetched
     } = this.state;
 
+    const icons = {
+      inProgress: <Icon type="pf" name="spinner2" />,
+      success: <Icon type="pf" name="ok" />,
+      failed: <Icon type="pf" name="error-circle-o" />,
+      archived: <Icon type="fa" name="archive" />
+    };
+    const breadcrumbIcon =
+      (planRequestFailed && icons.failed) ||
+      (planArchived && icons.archived) ||
+      (planFinished && icons.success) ||
+      (!planNotStarted && icons.inProgress) ||
+      null;
+
     return (
       <React.Fragment>
         <Toolbar>
@@ -127,7 +142,12 @@ class Plan extends React.Component {
             <Link to="/migration">{__('Migration')}</Link>
           </li>
           {!isRejectedPlan &&
-            planName && <Breadcrumb.Item active>{planName}</Breadcrumb.Item>}
+            planName && (
+              <Breadcrumb.Item active>
+                {breadcrumbIcon}
+                {breadcrumbIcon ? ` ${planName}` : planName}
+              </Breadcrumb.Item>
+            )}
         </Toolbar>
 
         <Spinner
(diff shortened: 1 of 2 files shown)


michaelkro:
An archived plan will have the attribute `deleted_on: <some timestamp>`

Active plans (unarchived) will have the attribute `deleted_on: nil`

priley86:
@mturley @michaelkro - i would not be opposed to seeing that `deleted_on` attribute transformed to an `archived` boolean attribute in the reducer if you guys are not opposed. Since we are not displaying the `deleted_on` datetime anywhere, it seems a boolean in the UI model would suffice. Keeping these attributes in the reducer seems preferable so that they can be referenced anywhere in the UI then...

michaelkro:
Makes sense to me @priley86 👍

mturley:
@priley86, agreed. I'm all for abstracting away the API's model in favor of a simpler model of what the UI needs. Will keep our propTypes cleaner.

I'm on it as soon as I get in today (by 11)

mturley:
@michaelkro thanks for the API info. Did you find that by inspecting the objects yourself, or is there documentation I'm not finding?

michaelkro:
I was working on the `archive` feature, so just so happened to be fresh on the mind

AparnaKarve:
@michaelkro Agreed. There is no way to know this stuff unless one has been following along on that thread.

@priley86 I really like the idea of abstracting the model's details in the Reducer... we have been doing that in other places too, it's just great.
That said, I'm (anyway) planning to work on the backend side to make sure that the `archived` and the `active` attributes are supported in the API -- but for now, let's go with `deleted_on`.

app/javascript/react/screens/App/Plan/Plan.js
```diff
@@ -127,7 +141,13 @@ class Plan extends React.Component {
             <Link to="/migration">{__('Migration')}</Link>
           </li>
           {!isRejectedPlan &&
-            planName && <Breadcrumb.Item active>{planName}</Breadcrumb.Item>}
+            planName && (
+              <Breadcrumb.Item active>
+                {breadcrumbIcon}
```
AllenBW: golly `&nbsp;` makes me cringe... what about doing a template literal here instead?  🤔 (or not and I'll keep cringing)

mturley: `Yeah,&nbsp;they&nbsp;are&nbsp;ugly&nbsp;aren't&nbsp;they.&nbsp;&nbsp;`

as much as i want to keep making you cringe, i'll use a literal. good point. I figured this was better than special margin CSS, but I forgot i could specify spaces more cleanly, so thanks for that tip @AllenBW

serenamarie125:
@mturley I think we need to be consistent, and have icons with all.

i think pficon-off would be not started
I think pficon-migration would be in progress ... which Is really migrating right @vconzola ?

vconzola:
@mturley We don't use a status icon for Not Started plans on the Overview page, so I recommend we don't include one on the Details page. And using the pficon-spinner2 for In Progress would be consistent with what we're showing on the Overview and for individual VM in-progress indicators. @serenamarie125 We don't use pficon-migration anywhere else in the UI so I wouldn't use it here.

app/javascript/react/screens/App/Plan/Plan.js
priley86: `&nbsp;  4 L1F3 &trade;`

priley86: to be honest i wish spaces looked more like this: `gl.createBuffer(); gl.bufferData([1.0, -1.0,  1.0, 1.0, -1.0,-1.0, -1.0,]` - but hey, who's being picky here? 🤣 

+1 for the template literal if that's easier to see in the JSX... i just thought you guys liked to kick it old-school... 🎶 📻 🎶

mturley:
Got it, thanks @vconzola .

mturley:
And thanks @AparnaKarve for the rest of the details. Pushing some final commits here in a moment

mturley:
This should be good, but I would love help test with real Plans in each of the 5 states!

Not started (no icon):
screen shot 2018-06-06 at 2 19 54 pm

In progress (@vconzola this is how it looks statically, is that `spinner2` supposed to spin? i can animate it...):
screen shot 2018-06-06 at 2 17 24 pm

Completed successfully:
screen shot 2018-06-06 at 2 16 43 pm

Completed with errors:
screen shot 2018-06-06 at 1 47 27 pm

mturley:
Hm. One last commit to fix the extra space when there's no icon. just a sec.

mturley:
All ready to go when you guys are @priley86 @AparnaKarve @michaelkro @AllenBW

mturley:
oops! Thanks @michaelkro I forgot to post the archived screenshot!

mturley:
Thanks guys!

Issue #394, opened 2018-06-06
[Test Automation Requirement] Please add an id or a name to div surrounding the Infrastructure Mappings List


kedark3:
See this screen shot for proposed id and the div where it should be
![screenshot from 2018-06-06 10-01-11](https://user-images.githubusercontent.com/8180469/41042857-abe1f716-6970-11e8-8759-77c9df127697.png)

kedark3:
@priley86 we may need some more identifiers in here: 
![screenshot from 2018-06-06 12-39-32](https://user-images.githubusercontent.com/8180469/41052432-be50eb26-6986-11e8-8fed-53307b924ef4.png)
Most of the items have only class "row-sm-4" or "col-sm-4" which makes it harder to tell which is which in automation.

priley86:
the original issue should now be resolved. I would suggest a new issue for any other related automation in the expansion below (since that UI was changed recently). hope this makes sense!

PR #395, opened 2018-06-07
adds compound expansion infra mappings UI

### LINKED ISSUE #310: Enhance infrastructure migration info

priley86:
Closes #310 #394 

* Adds new compound expansion UI for Infrastructure Mappings.
* Queries networks and datastores for displaying additional attributes (such as the network and disk names) on the infrastructure mapping list view. Adds cluster's provider to the clusters query. Add all API actions/reducers associated.
* Create new helper method for bucketing the unique target clusters/datastores/lans across a mapping into a project data structure (used to construct these new visualizations)
* Add custom list view styling and expanded display styles. @serenamarie125 @vconzola do you have any feedback on this? I just used CSS to construct the borders and arrow. I couldn't tell from the UX if you'd like a specific border style or border gradient applied. Also - do we have preferred shade of blue for the target border?

Screenshots...

## Networks
screen shot 2018-06-07 at 9 19 37 am

## Datastores
screen shot 2018-06-07 at 9 19 46 am

## Clusters
screen shot 2018-06-07 at 9 19 52 am

## Associated Plans
screen shot 2018-06-07 at 9 20 41 am

## Error handling
If we receive any errors loading mappings/clusters/datastores/networks - just display an error. I don't want to overcomplicate this error handling logic if we don't have to. You need these data collections to construct the models. You'd need the mappings and clusters to display all of the visualizations.
screen shot 2018-06-06 at 5 53 30 pm

GIF of UX
compound-expand-list-view


app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.scss
```diff
@@ -0,0 +1,101 @@
...
+    flex-direction: row;
+  }
+  .infra-mapping-item-source,
+  .infra-mapping-item-target {
+    height: 100%;
+    padding: 10px;
+    width: 75%;
+  }
+  .infra-mapping-item-source {
+    border: 1px solid black;
```
AllenBW: Though I wasn't asked, we'll likely wanta use something from https://github.com/patternfly/patternfly/blob/master/src/less/color-variables.less#L76 for this color

priley86: yep... colors t.b.d...thanks!!

AllenBW: Something like... 

``` @import './node_modules/patternfly/dist/sass/patternfly/color-variables';```

then `$color-pf-blue`

I can't for the life of me understand why its so difficult to get to the colors though 🤔 😢

**Comment by michaelkro** [comment]
Are we not concerned with deduping networks for these visualizations?

**Comment by priley86** [comment]
@michaelkro i was asking @AparnaKarve the same question... From what it sounds, the dedupe only needs to occur on the creation (in the Mapping Wizard). After creation, they should not be duplicated.

@AparnaKarve can you confirm?

File: app/javascript/react/screens/App/Overview/OverviewReducer.js [hunk 2]
```diff
@@ -227,8 +237,47 @@ export default (state = initialState, action) => {
...
+        .set('clusters', action.payload.data.resources)
+        .set('isFetchingClusters', false)
+        .set('isRejectedClusters', false)
+        .set('errorClusters', null);
+    case `${V2V_FETCH_CLUSTERS}_REJECTED`:
+      return state
+        .set('errorClusters', action.payload)
+        .set('isRejectedClusters', true)
+        .set('isFetchingClusters', false);
+
```
AllenBW: tiny nothing, inconsistent spacing between reducer cases 🔍

AparnaKarve:
@priley86 Looking at the Network screenshot above, it does look like a dedupe might be necessary.

I think we have 2 options here -
- Implement the deduplication similar to the deduplication in the Networks step

or

- Add more info about each Network, such as -
```
VM Network (on switch 1)                                                   
                            ------------------>     VM_Tagged (on switch 1)
VM Network (on switch 2)
```

Although, from an end user's standpoint, I don't know if we should disclose that there are multiple switch id's involved here

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,551 @@
...
+
+                    return (
+                      <ListView.Item
+                        key={mapping.id}
+                        heading={mapping.name}
+                        description={
+                          <small>
+                            {__('Completed: ')}
+                            {moment(mapping.created_at).format(
+                              'MMMM Do YYYY, h:mm a'
```
AllenBW: we do this in infastructuremappinglist.js and planrequestdetailslist... might be time to make a constant 🤔

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,551 @@
...
+                                mapping.expanded && mapping.expandType === 0
+                              }
+                              toggleExpanded={() => {
+                                this.toggleExpand(mapping, 0);
+                              }}
+                            >
+                              <Icon type="pf" name="network" />
+                              <div className="mappings-expand-label-group">
+                                <div className="mappings-expand-label">
+                                  <strong>{sourceLanCount}</strong>&nbsp;
```
AllenBW: if we gotta have em, we gotta have em, but if there's a better way than sprinkling `&nbsp;` :-/ :trollface:

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,551 @@
...
+                              }
+                              toggleExpanded={() => {
+                                this.toggleExpand(mapping, 1);
+                              }}
+                            >
+                              <Icon type="fa" name="database" />
+                              <div className="mappings-expand-label-group">
+                                <div className="mappings-expand-label">
+                                  <strong>{sourceDatastoreCount}</strong>&nbsp;
+                                  {sourceDatastoreCount === 1
```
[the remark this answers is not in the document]
AllenBW: oh same comment for where this is done elsewhere

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,551 @@
...
-                                ))
-                              : null}
+                          <Grid.Col sm={12}>
+                            {mapping.expandType === 0 && (
+                              <React.Fragment>
+                                <Grid.Row>
+                                  <Grid.Col xs={6}>
+                                    <b>{__('Source Networks')}</b>
+                                  </Grid.Col>
+                                  <Grid.Col xs={6} style={{ paddingLeft: 0 }}>
```
AllenBW: [le sigh... one day... ](https://v4-alpha.getbootstrap.com/layout/grid/#no-gutters)

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,551 @@
...
-                                    <span>{plan.name}</span>&nbsp;&nbsp;
-                                  </div>
-                                ))
-                              : null}
+                          <Grid.Col sm={12}>
+                            {mapping.expandType === 0 && (
+                              <React.Fragment>
+                                <Grid.Row>
+                                  <Grid.Col xs={6}>
+                                    <b>{__('Source Networks')}</b>
```
AllenBW: Should this and line 342 be pluralized 🤔  (cuz sometimes there is a single source network - target network)

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: sure - can add a common helper...

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: i still think those are more explicit and we don't really have a standard on this... i.e. "how many spaces am I looking at..."

priley86: what do you do elsewhere in MIQ?

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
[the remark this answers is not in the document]
priley86: sure - maybe a helper which renders that internally?

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: those are unique to these only. I guess I can put them in classes if you really want ;)

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: I don't think a label should be pluralized.

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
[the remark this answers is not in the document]
michaelkro: I know we have to be careful about using `sprintf` to modify words...  not sure about using it for pluralization though

@AparnaKarve?

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
AllenBW: yeah it happens everywhere in miq... doesn't mean I like it 😏

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
AllenBW: I thought we did this somewhere else... but not finding it.. might not be a bad idea to put em in a class for use elsewhere if/when necessary 🤷‍♂️

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.scss
priley86: i used `$color-pf-blue-400` after eyeballing this a few times... it was the darkest pf-blue color I could denote any distinction. The other colors do not differentiate enough from the the left-hand black border w/ these 👀 . Please note that... we will need to keep a patternfly or patternfly-react dependency here in v2v plugin to get those imported sass variables in this plugin (from the npm installed package). So my vote is to just keep patternfly-react since it depends on patternfly and we going to continue consuming es6 helpers from PF React in the future)...

HOWEVER - patternfly-react (and other ui-classic common packages) should also still be updated in ui-classic when doing upgrades from now on...

cc: @himdel 
https://github.com/ManageIQ/guides/pull/314

another twist on this would be... w/ PF Next (and PF React Next components coming) - we'll be exporting the CSS directly in the ES6 module alongside the React component (so that problem kinda dissipates). More on this in a PR coming upstream soon....but it shouldn't affect us until after GA (when we could even consider PF next components).

lots of tangents but figured it would be good to note this here 🍰 

📚

app/javascript/react/screens/App/Overview/OverviewReducer.js
priley86: removed empty lines in that reducer

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: ok man - you win... my <3 of JSX supersedes my <3 of explicit spaces... 

JSX is really HTML6 ... a lotta folks just don't really admit it yet ;)

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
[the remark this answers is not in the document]
priley86: you talked me into creating a simple pluralize helper... i think the shorthand notation will save us a lot of keystrokes. Hopefully the translators can manage the additional translation.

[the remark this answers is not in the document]
priley86: otherwise our pluralize formula gets very complex...

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: Sass and friends are happy ;)

app/javascript/common/helpers.js
```diff
@@ -1,3 +1,6 @@
 // todo: expose this from patternfly-react instead?
 export const getDisplayName = Component =>
   Component.displayName || Component.name || 'Component';
+
+export const pluralize = (count, string) =>
```
himdel: Please don't ever use this `pluralize` :)

Won't work with i18n at all.

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,561 @@
...
+                              toggleExpanded={() => {
+                                this.toggleExpand(mapping, 0);
+                              }}
+                            >
+                              <Icon type="pf" name="network" />
+                              <div className="mappings-expand-label-group">
+                                <div className="mappings-expand-label">
+                                  <strong>{sourceLanCount}</strong>
+                                  {` `}
+                                  {pluralize(sourceLanCount, 'Source Network')}
```
himdel: The canonical gettext way of doing this would be `n_('Source Network', 'Source Networks', sourceLanCount)` ... except it's still not quite right, because the order is not necessarily  "# thingies" in all languages.

So... this would end up being `sprintf(n_('%d Source Network', '%d Source Networks', sourceLanCount), sourceLanCount)`.

(That, or

```
n_(`${sourceLanCount} Source Network`, `${sourceLanCount} Source Networks`)
```

once our i18n tooling supports template strings properly. (Unless it already does @mzazrivec ?)

app/javascript/common/helpers.js
AparnaKarve: @priley86 Grammar rules vary from language to language, so what defines 'pluralization' in English (adding an `s` or `es` in most cases) does not apply to all languages.

To keep it simple, lets just go with - 
`__('Source Datastore')` for one and 
`__('Source Datastores')` for many

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
AllenBW: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

app/javascript/common/helpers.js
AllenBW: this is a SUPER neat idea though! Like this idea... if complete sentences make the world an easier place... we really should just use the react-intl helper thing... (especially since we already have the dep)


[we can use the react-intl formatted plural thing](https://github.com/yahoo/react-intl/wiki/Components#formattedplural)

```
<FormattedPlural value={sourceDatastoreCount} one=__('Source Datastore') other=__('Source Datastores') />
```

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: I don’t think the other 3rd party libraries are any more reliable than gettext (I’ve seen others mentioned too besides these). My vote is to stay consistent with gettext and if that’s the way this is achieved that’s fine, but it’s not reducing the original solution much. I still think on-Prem translation solutions end up on the translator. There is SaaS offerings for i18n (real time translation APIs) that are very reliable, but it doesn’t look like we’ll be investing in those.

priley86: I don’t think the other 3rd party libraries are any more reliable than gettext (I’ve seen others mentioned too besides these). My vote is to stay consistent with gettext and if that’s the way this is achieved that’s fine, but it’s not reducing the original solution much. I still think on-Prem translation solutions end up on the translator. There is SaaS offerings for i18n (real time translation APIs) that are very reliable, but it doesn’t look like we’ll be investing in those.

priley86: Was just looking for the one off solution for these cases (not pluralizing all)

priley86: I don't want to render the count within the same block (because styling is different around the <strong> tags). So for now, I just renamed the method (I think my original intent was only for these simple scenarios), however I think it could be made more flexible this way...

Happy to revisit this in the future, but I think this is good enough for now.

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,574 @@
...
+                              toggleExpanded={() => {
+                                this.toggleExpand(mapping, 0);
+                              }}
+                            >
+                              <Icon type="pf" name="network" />
+                              <div className="mappings-expand-label-group">
+                                <div className="mappings-expand-label">
+                                  <strong>{sourceLanCount}</strong>
+                                  {` `}
+                                  {simplePluralize(
```
AllenBW: effin elegant 🤗 🏩

AparnaKarve:
The Network dedupe can be addressed separately in a smaller PR.

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
```diff
@@ -1,226 +1,574 @@
...
+                                    sourceLanCount,
+                                    __('Source Network'),
+                                    __('Source Networks')
+                                  )}
+                                </div>
+                                <div className="mappings-expand-label">
+                                  <strong>{targetLanCount}</strong>
+                                  {` `}
+                                  {simplePluralize(
```
himdel: Still problematic, for mostly the same reasons .. (or see the list of plural rules in https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals )

michaelkro:
I'm still seeing duplicated source networks in the Infrastructure Mapping networks expansion
* v2v master
* nightly db

manageiq__migration

priley86:
ok - thanks, i'll look at this again shortly...

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
priley86: Ok - so this is becoming a PF design issue as well I'm afraid. If we need to style the count separately from the label, then we'd need to extract two strings. I think your original suggestion breaks this design. I will follow up on this...

priley86: Another option may be to render that count alongside the text and just hide the count in the case we are displaying the label, however I think this would also be confusing to a translator. Has this not come up before? Displaying counts with differing styles?

priley86: Will just remove the count styling for now.

AparnaKarve: @himdel From a Gaprindashvili standpoint, I don't think we can do this (what you suggested above) -
```js
n_(`${sourceLanCount} Source Network`, `${sourceLanCount} Source Networks`)
```
If the support for enhanced pluralization has been added in ui-classic, it must be pretty new.

@priley86 We can look into enhanced pluralization all throughout the app after the release is out the door.

AparnaKarve:
@michaelkro Yes, we agreed that the Network dedupe could be fixed in a separate PR.
https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/395#issuecomment-395897414

michaelkro:
whoops, my bad guys

app/javascript/react/screens/App/App.scss
```diff
@@ -1,2 +1,3 @@
+@import 'patternfly/dist/sass/patternfly/color-variables';
```
himdel: This breaks gaprindashvili ui-classic travis:

```
     # --- Caused by: ---
     # Sass::SyntaxError:
     #   File to import not found or unreadable: patternfly/dist/sass/patternfly/color-variables.
```

(https://travis-ci.org/ManageIQ/manageiq-ui-classic/builds/391256750)

There is no `patternfly`, since neither v2v nor ui-classic depend on any such package now.

This should be `patternfly-sass/dist/...` - fixed in #410.

app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
himdel: @AparnaKarve Not really, it was introduced together with the rest, back in December 2015.

simaishi:
@priley86 Is there a BZ that goes with this PR? I can't bacakport unless this change is approved for the next release.

bthurber:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1600635

simaishi:
Looks like this is already in Gaprindashvili branch (rebased from master before I took over the backport), removing the `yes` label.

PR #405, opened 2018-06-11
Fix duplicate network on Infrastructure Mapping

priley86:
Deduplicate infra mappings networks on the Infra Mapping List and update counts correctly. Change the `Completed` label to`Created`.

Also incorporates #403 

Also fixes i18n issue... cc: @himdel 

screen shot 2018-06-11 at 10 45 05 am

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
+++ b/app/javascript/react/screens/App/Overview/components/InfrastructureMappingsList/InfrastructureMappingsList.js
@@ -37,26 +37,6 @@ class InfrastructureMappingsList extends React.Component {
     };
   }
 
-  sourceCount = (mapping, sourceType) =>
-    [
-      ...new Set(
-        mapping
-          .filter(item => item.source_type.toLowerCase() === sourceType)
-          .map(item => item.source_id)
-      )
-    ].length;
-
-  destinationCount = (mapping, destinationType) =>
-    [
-      ...new Set(
-        mapping
-          .filter(
-            item => item.destination_type.toLowerCase() === destinationType
-          )
-          .map(item => item.destination_id)
-      )
-    ].length;
-
   clusterName = cluster =>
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   globals: {
     $: true,
     __: true,
+    n__: true,
```
AparnaKarve: @priley86 Thanks, this looks good.
I have one concern though -- this won't work with Gaprindashvili
@himdel Can you confirm?

I was thinking that we could add the `n` implementation after G is released, just to keep things easier right now with our `gaprindashvili` branch.

Once G is released, we could apply the `n` changes all over the app.

priley86: @himdel @AparnaKarve can you please discuss it and give me the suggested solution for now? I would like to get past this one :)

priley86: we can revisit this in a future PR. Since `gaprindashvili` does not currently support, I am reverting this change.

AparnaKarve: @priley86 Thanks!

himdel: `n__` has been supported since 2015 in our JS.

Why would it not work in gaprindasvhili?

himdel:
Not seeing any i18n changes, did this move to a different PR?

.eslintrc.js
mzazrivec: Where does the thing about `n__` not working in gaprindashvili come from? We have had it for quite a while actually.

priley86: @AparnaKarve - i will add the `n__` syntax back in a separate PR later today and let you confirm it once more...

AparnaKarve: `n__` has never been used before in our ui-classic JS sources yet, hence wanted to be doubly sure that it was OK for gaprindashvili.

@priley86  apologies for these extra i18n iterations.

With the release impending for Gaprindashvili, just being extra cautious that a wrong commit does not get in 

Thanks everyone!

bthurber:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1572169

PR #410, opened 2018-06-13
Look for color-variables at correct place

ZitaNemeckova:
Fixes huge Travis failure of Gaprindashvili branch (manageiq-ui-classic)

Error:
```
ActionView::Template::Error:
        File to import not found or unreadable: patternfly/dist/sass/patternfly/color-variables.
```
Must go to Gaprindashvili asap :)

### DIFF
--- a/app/javascript/react/screens/App/App.scss
+++ b/app/javascript/react/screens/App/App.scss
@@ -1,3 +1,3 @@
-@import 'patternfly/dist/sass/patternfly/color-variables';
+@import 'patternfly/color-variables';
 @import './Overview/Overview.scss';
 @import './Plan/Plan.scss';


himdel:
This now references `patternfly-sass-$VERSION/dist/sass/patternfly/_color-variables.scss`, coming from the patternfly-sass gem.

(The path is this short because this is the asset pipeline, and `patternfly-sass-$VERSION/dist/sass` is what is in the search path.)

This is relevant both for master and for gaprindashvili. (In master, patternfly is installed in node_modules, but it is not a direct dependency, so still not a good idea to rely on it.)

himdel:
@priley86 Merging to get gaprindashvili green.

Please take a look once you're up, just to make sure :)

priley86:
ok this looks fine to me ;)  - as long as we can consume the Sass variables. thanks @himdel @ZitaNemeckova 

In the distant future w/ PF Next components, I would like to ensure we can consume the variables from JS (currently looking including them in the JS modules w/ the pf-next style components), but I think this is a ways out still...

PR #416, opened 2018-06-15
move mapping description to list item

priley86:
* flip the mapping description and the created at label locations
* use ellipsis utility if the description is too long

screen shot 2018-06-15 at 11 30 02 am


AparnaKarve:
LGTM2! :)

@priley86 just one quick question on `Created` ...
Are we showing `Created` only when there are associated plans?

PR #418, opened 2018-06-15
Make the elapsed time clocks start ticking!

### LINKED ISSUE #338: Plans In Progress: Elapsed time should tick every second regardless of polling interval
Currently, the elapsed time only changes every 15 seconds (or whenever the polling returns). It should tick once per second to stay in sync with the actual elapsed time.

mturley:
I added a reusable `TickingIsoElapsedTime` component, which takes as props a `startTime` (which can be either a string or a `Date()`), and an optional `endTime`. It'll tick once per second if the endTime is null, or show the total elapsed time if endTime is non-null.

Also it renders as a `React.Fragment`, so you can stick it as a child anywhere and it will add only the text node to your DOM. So we can use this inside any element.

Thanks for bearing with me on this, I was caught up working on http://what-is-react.surge.sh/ today.

Fixes #338.

pexxovqa3a 1

### DIFF
--- a/app/javascript/components/dates/IsoElapsedTime.js
+++ b/app/javascript/components/dates/IsoElapsedTime.js
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-export const IsoElpasedTime = (startTime, endTime) => {
+export const IsoElapsedTime = (startTime, endTime) => {
   let elapsedHours = moment(endTime).diff(startTime, 'hours');
   const elapsedMinutes = moment(endTime).diff(startTime, 'minutes') % 60;
   const elapsedSeconds = moment(endTime).diff(startTime, 'seconds') % 60;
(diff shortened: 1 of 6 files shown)


AparnaKarve:
@mturley I briefly tested this and your continuously ticking clock looks perfect so far!!
More testing on Monday...

Truly appreciate your effort here! Thanks!!

mturley:
https://rhux.slack.com/files/U80S1C3S8/FB8LMARPC/pexxovqa3a.gif

mturley:
My pleasure! These simple little React components are a delight to code up, so it was a nice break.

app/javascript/components/dates/IsoElapsedTime.js
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-export const IsoElpasedTime = (startTime, endTime) => {
```
mturley: 😏

priley86:
Looks great! Thanks @mturley  👍

mturley:
Let's merge this as-is for v2v, but if/when we abstract this out for patternfly-react (it will work fine in both pf3 and pf4, it's just text!), we should add support for formatting the date/time differently. Maybe an optional prop, with ISO as the default. Something we could just pass along to momentjs.

Anyway, thanks for the review @priley86

PR #419, opened 2018-06-18
Use request_state attribute to control polling

AparnaKarve:
`fulfilled_on === null` works well for the master branch, but seems to be an undefined in gaprindashvili until the request is actually finished.
This is affecting polling and hence the auto-refresh on the page.

The `request_state` attribute should work for both master and gaprindashvili

This is a possible fix for
https://bugzilla.redhat.com/show_bug.cgi?id=1588577

-- will be confirmed after some testing

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -63,7 +63,7 @@ class Plan extends React.Component {
         const mostRecentRequest = getMostRecentRequest(miq_requests);
         const planRequestId = mostRecentRequest.id;
         fetchPlanRequestAction(fetchPlanRequestUrl, planRequestId);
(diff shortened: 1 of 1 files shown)


priley86:
I confirmed that `request_state` is indeed coming back in the `gaprindashvili` stream for me...
```
{
  "href": "http://localhost:3000/api/service_requests/4",
  "id": "4",
  "description": "PlanB",
  "approval_state": "approved",
  "type": "ServiceTemplateTransformationPlanRequest",
  "created_on": "2018-05-16T00:19:38Z",
  "updated_on": "2018-05-25T13:41:30Z",
  "requester_id": "1",
  "requester_name": "Administrator",
  "request_type": "transformation_plan",
  "request_state": "active",
```

AllenBW:
@AparnaKarve wondering if this should be g specific though?

PR #426, opened 2018-06-21
move patternfly dependency to devDependencies

priley86:
Since patternfly & patternfly-react are now sourced from ui-classic, we are safe to remove this dependency from v2v. Updates to patternfly should now take place there, instead.

This is not ready for `gaprindashvili` since patternfly-react is not referenced there in ui-classic.


priley86:
hmm it breaks tests unfortunately... I will have to move it to devDependencies instead I suppose...

priley86:
A couple prettier updates after the semantic update in yarn lock. Also removing the `sort-comp` lint rule since it's been removed in PF React. Looking to adopt [eslint-plugin-patternfly-react](https://github.com/patternfly/patternfly-react/tree/master/packages/eslint-plugin-patternfly-react) more fully in a future PR.

Issue #422, opened 2018-06-20
Add v2v commit hash to the UI


priley86:
Per discussion w/ @AparnaKarve - i would like to look into adding the last v2v commit hash somewhere into the MIQ UI. I think it would probably be easiest to place this in the existing "About Modal" like we do with the MIQ commit hash and add a section for "V2V Version".

![screen shot 2018-06-20 at 8 43 41 am](https://user-images.githubusercontent.com/4237045/41659112-603349f4-7466-11e8-95b7-7009073d18fb.png)

thoughts @vconzola ?

Otherwise, we can probably obscure this version and write it out somewhere in the HTML or in the Console logs...

AllenBW:
OH I LIKE THIS IDEA!!! 
maybe something like... 
<img width="604" alt="screen shot 2018-06-20 at 8 55 52 am" src="https://user-images.githubusercontent.com/6640236/41659625-cbe1184c-7467-11e8-96fc-dac606392925.png">

vconzola:
@priley86 This looks good to me. Let's bring it up in the sync meeting later today.

priley86:
Yes - exactly @AllenBW

AparnaKarve:
Since the About box is a ui-classic component, we would have to review this idea with some of the ui-classic folks first.
@priley86 In case if adding the hash in the  About box box does not work out for whatever reason, could we stick the hash somewhere in the Migration page html, like you originally suggested?

AllenBW:
Good news everyone!!

I was talking to @Fryguy and he said... (👈 this is the joke part)

TL;DR - If we're gonna list one plugin, we have to list em all. About Modal is fine, but we might want to have a scrolling boxy thing.  

### This is what "em all" really means
(from rails console `Vmdb::Plugins.instance.vmdb_plugins.map { |p| p.class.name }` produces this)

```
["ManageIQ::Providers::AnsibleTower::Engine",
 "ManageIQ::Providers::Amazon::Engine",
 "ManageIQ::Providers::Azure::Engine",
 "ManageIQ::Providers::Foreman::Engine",
 "ManageIQ::Providers::Google::Engine",
 "ManageIQ::Providers::Kubernetes::Engine",
 "ManageIQ::Providers::Kubevirt::Engine",
 "ManageIQ::Providers::Lenovo::Engine",
 "ManageIQ::Providers::Nuage::Engine",
 "ManageIQ::Providers::Redfish::Engine",
[...]

PR #436, opened 2018-06-22
Add v2v plugin branch and commit sha in the main html [#422]

AparnaKarve:
Add v2v branch and commit sha in the main html

What was suggested here https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/422#issuecomment-398808617 is applicable to all plugins, and may have to be implemented on a global level, possibly from miq core. While it is a complete solution, it may take a while to implement considering the extra styling required in the About Modal for the scrolling.
So for now what we could do is -  tackle this from our end.
That said, let's go with @priley86 's suggestion to put the hash in the html. 

See the screenshot below and checkout the commit sha in the highlighted line -

screen shot 2018-06-22 at 3 04 51 pm

Fixes https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/422
https://bugzilla.redhat.com/show_bug.cgi?id=1594348

### DIFF
--- a/app/controllers/migration_controller.rb
+++ b/app/controllers/migration_controller.rb
@@ -20,6 +20,11 @@ class MigrationController < ApplicationController
         "$(v2v.mount('#{name}', '#{selector}', #{data.to_json}));".html_safe
       end
     end
+
+    def migration_sha
+      version = vmdb_build_info(:version)
+      "#{version}.#{`git ls-remote https://github.com/ManageIQ/miq_v2v_ui_plugin.git #{version}`.split('refs')[0].try(:strip!) || 'undefined'}"
+    end
   end
 
   menu_section :migration
--- a/app/views/migration/index.html.haml
+++ b/app/views/migration/index.html.haml
@@ -1,2 +1,2 @@
-%div#reactRoot{:style => 'height: 100%;'}
+%div#reactRoot{:style => 'height: 100%;', :v2v_sha => migration_sha}
   = mount_react_component('manageiq-v2v', '#reactRoot', { user: @current_user })


AllenBW:
I LOVE this idea!! But... there's a catch... when you pull away the magic rug of internet (meaning when yah don't have an internet connection) this breaks v2v 😭 
screen shot 2018-06-25 at 8 43 04 am

maybe the def could fail more gracefully when it gets nothing back?  Also we'll have to update this url when the repo name changes... 🤔

priley86:
@AparnaKarve - i appreciate this one a lot!

I think this is a great feature as you've implemented it, but one concern i'd have is that this would give the commit hash at the time the MIQ server is started (since we are requerying that url each time), correct? Ideally we could persist the hash at the time the appliance image is built some place...

I'm fine w/ merging this as is for the preliminary support it would give. Would you mind if we keep the original issue/bugzilla open though?

AparnaKarve:
> maybe the def could fail more gracefully when it gets nothing back? Also we'll have to update this url when the repo name changes

@AllenBW Sure, that should be doable.
And yes we will need a repo name update in the url as well.

> I think this is a great feature as you've implemented it, but one concern i'd have is that this would give the commit hash at the time the MIQ server is started (since we are requerying that url each time), correct? Ideally we could persist the hash at the time the appliance image is built some place...

@priley86 The commit hash method would be executed every time you access Migration from the menu. Persisting on the Rails side, while doable is a more involved process and would be a good idea to get into that once we implement this for all plugins, not just v2v.

We could keep the BZ open, although I would need the BZ tagged to this PR in order to backport this to Gaprindashvili.
Also, since the BZ is not applicable to v2v alone, we should probably not categorize it as a v2v BZ.

AparnaKarve:
@AllenBW I have addressed the sans-internet case you mentioned above :)
In an event when there is no connectivity, you will see either `master.undefined` or `gaprindashvili.undefined` in the html.

Since there is no rush for this, please review/merge at your convenience.
Thanks.

priley86:
@AparnaKarve agree... OK - just let me know if you end up wanting to backport this to G. I don't think that's urgent (and we could just keep the issue and BZ open for now until the long term solution is put in place). Otherwise, will be happy to open another BZ later.

I'm fine w/ merging as-is if others are...

AparnaKarve:
Thanks folks!
I'm keeping the gaprindashvii/yes label ON because I think the hash will be useful to be in-sync with the QE folks especially in situations where they see an issue and we can't reproduce it because we are on the latest commit hash.

simaishi:
@AparnaKarve Doesn't this always return 'undefined' on downstream appliances?

AparnaKarve:
@simaishi Not sure, would have to check that.
Will let you know what I find.

AparnaKarve:
@simaishi For downstream appliances, it does not return `undefined` ...

Although, what I noticed was it just returns the commit sha for the current `master` head or the `gaprindashvili` head and not what the appliance is using.

Let's hold off on the `gaprindashvili` backport for now.
Thanks.

simaishi:
That's probably because downstream `version` tag (e.g. 5.10.0.0) won't exist on upstream repo. Not sure what the best way is, but hard-coding upstream repo url might not be a good idea... cc @Fryguy

AparnaKarve:
The `/var/www/miq/vmdb/BUILD` seems to contain the commit hash info for miq core

Currently I am looking at a 5.10 nightly appliance, which contains this in the BUILD file -
```
5.10.0.0-pre-nightly-20180604133801_a49e48a
```

So I would think that we need to add the commit hashes for other plugins in the BUILD file as well.

simaishi:
I'll leave that to other people to decide, as adding SHA for other plugins to BUILD file will require other changes and need to be discussed 😄

AparnaKarve:
True.

I guess I was overly optimistic when I thought we could resolve this from our side (the v2v side) :)

Issue #430, opened 2018-06-21
Consume PatternFly React eslint rules


priley86:
We are exposing a standard eslint rule plugin now from PatternFly React. It should be feasible to adopt those lint rules across v2v. This is definitely a background/code improvement task.

vconzola:
Close as wontfix - maybe someday.

PR #443, opened 2018-06-27
Increase printwidth in support of internalization 

AllenBW:
This pr is for master, not gonna try and backport this one to G if we are down with this work, will make one specific for Gaprindashvili

In support of #440

@mzazrivec would yah mind seeing if this branch captures all the missing strings?  There might be a few singletons we gotta make special exceptions for...

Also, if your wondering why we consolidated the prettier options... https://github.com/prettier/eslint-plugin-prettier#options


AparnaKarve:
@AllenBW Thanks for looking into this.

Wow, that's a lot of changes ... unfortunately it may not have caught all of them, I think :(

How about we just ` /* prettier-ignore */` or `// prettier-ignore` ignore them?
```js
<span>
   {__('There was an error loading Infrastructure Mappings.') /* prettier-ignore */}
</span>
```
@priley86 kinda suggested that in his EDIT comment here https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/440#issuecomment-400421252
and it seems to work.

The only catch is to grep the code for all instances of this - `__(`
 and then apply the `prettier-ignore` comment for every case manually.

AllenBW:
Oh yeah well we can do that to every single one, this wasn't intended to be a catch all, more like a catch most, minimizing the number of  strings we had to apply individual ignores to..

AllenBW:
In fact, this only leaves 4 strings that need to be fixed 😁 one more commit inc...

AparnaKarve:
Oh nice! I didn't realize it was that close :)

priley86:
@AllenBW i'm ok w/ extending the printWidth to 120 characters, but unfortunately this is not gonna jive w/ #430 ... so once we move to the PatternFly eslint rule plugin, we'd need to extend it to 120. 

Obviously the preference is to have as few of the prettier-ignoreline's sprinkled in the code as possible... Many of us have "formatOnSave" specified in VS Code so trying to make this as consistent as possible w/ PatternFly React settings.

priley86:
did we not validate the backend rake change would alleviate?

.eslintrc.js
```diff
@@ -20,10 +20,6 @@ module.exports = {
   ],
   plugins: ['prettier', 'jest', 'react'],
   rules: {
-    'prettier/prettier': [
```
priley86: prefer to keep them here to match PF React... any reason you prefer them in package.json?

priley86: we could extend them here later on as well..

package.json
```diff
@@ -5,7 +5,7 @@
     "test": "npm run lint && jest",
     "test:watch": "jest --watchAll",
     "test:current": "jest --watch",
-    "prettier": "prettier --single-quote --trailing-comma=none --write 'app/javascript/**/*.js' '*.{js,json}'",
+    "prettier": "prettier --write 'app/javascript/**/*.js' '*.{js,json}'",
```
priley86: ok i still think these are a little more readable up here, but whatever floats your boat ;) 

It does not seem that you get "single source of truth" (eslint & prettier configs) unless you are using the default prettier settings and [estlint-config-prettier](https://prettier.io/docs/en/eslint.html#turn-off-eslint-s-formatting-rules).  Since we are planning to override here, and then subsequently consume from an eslint plugin upstream, I think we'll just have to add back the eslint rule overrides later on (depending on if the upstream config goes to 120 or not).

Clear as mud 😆

.eslintrc.js
priley86: ok - i'm fine w/ removing them for now (and potentially reintroducing later...or not...hopefully). ;)

https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/443#discussion_r198525644

package.json
AllenBW: right 🤦‍♂️ 🏩 💘

AparnaKarve:
@AllenBW I do not see this line addressed - https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/components/DeleteInfrastructureMappingConfirmationModal/DeleteInfrastructureMappingConfirmationModal.js#L33

AllenBW:
@AparnaKarve just keepin' yah on yer toes 😏:heart: (GOOD CATCH, funny thing is, totally had that file open and modified, just didn't stage the file for commit)

AparnaKarve:
> just keepin' yah on yer toes

Haha.. ok :) -- hope I'm doing a good job

Just noticed the gaprindashvili/no label... this would need to be gaprindashvili/yes, right?
because #440 is gaprindashvili/yes?

AparnaKarve:
gaprindashvili/yes would mean all the other processes involved ...
like, we need a blocker BZ for 5.9.3 so @simaishi can backport this PR

AparnaKarve:
@AllenBW That works! Thanks.

This PR looks good ... Thanks again for taking the initiative on this!

Issue #424, opened 2018-06-20
UI/API schema checks 


priley86:
In this issue, we will look at API schema validation once the JSON responses have been parsed by the UI. If schema checks fail and attributes are missing that are expected from the current UI build, we will log a schema failure to the console. This should help UI developers analyzing QE issues.

priley86:
Going to start investigating this...

priley86:
going to continue moving here and try to have some things ready for test mid next week.

AllenBW:
Pr merged, we can now close this (remember to use yer github hooks 😏 )

priley86:
thanks - yea... i will probably open up some more issues as I have time to work on adding more validation, but I'm fine w/ closing this initial issue now!

AllenBW:
Future issue, can be schema *enhancement* 😏 😋

PR #453, opened 2018-06-28
[#424] schema validators for Overview screen

priley86:
Starts the UI schema validation for #424 

Adds validators for all of the Overview APIs. Would <3 some feedback on this. My current thinking is - these schemas will be screen/component specific (since we sometimes specify different attributes in the requests per page/component), so these schemas probably belong alongside the reducers. 

When a validation error is found, we'll get a notification in the UI like this:
screen shot 2018-06-26 at 4 03 56 pm

The goal of this is to quickly isolate UI/API integration issues in the QE test environments. QE should be able to quickly identify these and report them back to us (indicating a missed build integration or a missed API integration in the UI).

@AparnaKarve  - I was not able to dig through every field we currently use on the `Request` and `Plan` objects. This was an initial attempt. Will continue reviewing this but posting this for early 👀

### DIFF
--- a/app/javascript/data/models/clusters.js
+++ b/app/javascript/data/models/clusters.js
@@ -1,0 +1,19 @@
+import { array, string, object } from 'yup';
+
+export const clustersSchema = array()
+  .of(
+    object().shape({
+      href: string().required(),
+      id: string().required(),
+      ems_id: string().required(),
+      uid_ems: string().required(),
+      ext_management_system: object()
+        .shape({
+          emstype: string().required(),
+          name: string().required()
+        })
+        .required(),
+      v_parent_datacenter: string().required()
+    })
+  )
+  .nullable();
--- a/app/javascript/data/models/clusters.js
+++ b/app/javascript/data/models/clusters.js
@@ -13,7 +13,7 @@ export const clustersSchema = array()
           name: string().required()
         })
         .required(),
-      v_parent_datacenter: string().required()
+      v_parent_datacenter: string().nullable()
     })
   )
   .nullable();
--- a/app/javascript/react/screens/App/Overview/OverviewValidators.js
+++ b/app/javascript/react/screens/App/Overview/OverviewValidators.js
@@ -1,222 +1,36 @@
-import { array, string, object } from 'yup';
-import { logSchemaError } from '../../../../common/schemaHelpers';
+import { validateSchema } from '../../../../data/schemaHelpers';
+import { clustersSchema } from '../../../../data/models/clusters';
+import { datastoresSchema } from '../../../../data/models/datastores';
+import { networksSchema } from '../../../../data/models/networks';
+import { plansSchema } from '../../../../data/models/plans';
+import { providersSchema } from '../../../../data/models/provider';
+import { requestsSchema } from '../../../../data/models/requests';
+import { mappingsSchema } from '../../../../data/models/transformationMappings';
 
-const providersSchema = array()
-  .of(
-    object().shape({
-      href: string().required(),
-      type: string().required(),
-      id: string().required()
-    })
-  )
-  .nullable();
-
-export const validateProviders = providers => {
-  providersSchema
-    .validate(providers, { abortEarly: false, strict: true })
-    .catch(err => {
-      logSchemaError(err);
-    });
+export const validateOverviewClusters = clusters => {
+  validateSchema(clustersSchema, clusters);
 };
 
-const clustersSchema = array()
-  .of(
-    object().shape({
-      href: string().required(),
-      id: string().required(),
-      ems_id: string().required(),
-      uid_ems: string().required(),
-      ext_management_system: object()
-        .shape({
-          emstype: string().required(),
-          name: string().required()
-        })
-        .required(),
-      v_parent_datacenter: string().required()
-    })
-  )
-  .nullable();
-
-export const validateClusters = clusters => {
-  clustersSchema
-    .validate(clusters, { abortEarly: false, strict: true })
-    .catch(err => {
-      logSchemaError(err);
-    });
+export const validateOverviewDatastores = datastores => {
+  validateSchema(datastoresSchema, datastores);
 };
 
-const networksSchema = array()
-  .of(
-    object().shape({
-      href: string().required(),
-      id: string().required(),
-      switch_id: string().required(),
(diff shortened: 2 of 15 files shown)


AparnaKarve:
@priley86 This is a very interesting concept of having our own frontend schema validator.
So down the line if the backend pulls the rug under us (delete/update certain attributes in the API), we will instantly know what is missing.

I have recently seen a case first hand, where an attribute (`deleted_on`) was not present in the G branch, because of which archiving plans was failing -- so this tool would have been very handy then.

One question I have is about the extra call in the reducers to validate the schema -- is it lightweight enough to keep it in there even in production mode?

priley86:
yea - i think the overhead is worth it given current conditions/architecture (fairly lightweight here). The idea is to notify the user if the schema is invalid and continue moving the reducer regardless of failure...we could obviously disable that in production builds, but if we are doing our homework, this should *never* happen in a build that we ship. I think this is something we should target for post 5.9.4 since we are so closing to releasing that branch (not comfortable adding this in so late in the game for that).

thoughts?

app/javascript/react/screens/App/Overview/OverviewValidators.js
```diff
@@ -0,0 +1,222 @@
...
+  .of(
+    object().shape({
+      href: string().required(),
+      type: string().required(),
+      id: string().required()
+    })
+  )
+  .nullable();
+
+export const validateProviders = providers => {
```
AllenBW: Looks like we do this `validate` function a few times and in large its the same each time, would be nice to abstract this from each schema to have one generic validation tool to export

app/javascript/react/screens/App/Overview/OverviewValidators.js
```diff
@@ -0,0 +1,222 @@
+import { array, string, object } from 'yup';
+import { logSchemaError } from '../../../../common/schemaHelpers';
+
+const providersSchema = array()
```
AllenBW: for some reason I got this itch... that all schemas should be in their own file... kinda like how we have constants in their own file... 

(BTW writing this comment after the first one, skip down to the one on line 14 then continue reading)

And if there exists a generic validator function stored in some `common` folder ☝️ becomes a real real reality

package.json
```diff
@@ -59,6 +59,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "core-js": "^2.5.7",
```
AllenBW: but aren't these dev dependencies? 🤔

app/javascript/react/screens/App/Overview/OverviewValidators.js
priley86: i started it this way... but then i realized we query different attributes in different locations. I can look into extending a base schema object then?

package.json
priley86: No... they would be runtime. That is the main idea i guess - would like this to run on the physical machine and validate responses on the fly...

app/javascript/react/screens/App/Overview/OverviewValidators.js
priley86: this can be done just calling validate on a passed in schema in a generic method, will update.

app/javascript/react/screens/App/Overview/OverviewValidators.js
priley86: So I did some more research on this - it looks it's possible to achieve what we want using `schema.concat()` (so we can extend arrays/objects/etc.), but we just have to keep in mind that once a base type has been defined for a field, it must be compatible once extended. Here is a simple example that demonstrates this:

```
const contactSchema = object({
  name: string()
    .required(),
  email: string()
    .email(),
  address: object({
    line1: string().required(),
    zipcode: string().required()
  })
})
```
can be extended like this (say you want to extend address and add line2 after the fact):
```
let contactAddress = object({
    address: object({
      line2: string().required()
  })
});

let contactSchemaEnhanced = contactSchema.concat(contactAddress);
let contact = {
  name: 'jimmy',
  email: 'jdog@cool.biz',
  address: {
    zipcode: '12345'
  }
}
await contactSchemaEnhanced.validate(contact, { abortEarly: false, strict: true })
```
and this would validate and work like you would expect (giving line1 and line2 validation errors)...

*However*, two schemas are considered incompatible if the overrides for the same field are not the same type... so extending like this:
```
let contactAddress = object({
    address: object({
     line2: string().required(),
     zipcode: number().required()
  })
});
let contactSchemaEnhanced = contactSchema.concat(contactAddress);
```
would give an error: 
`You cannot `concat()` schema's of different types: string and number`

This is pretty much consistent with our data model (you'd never expect the same field to return w/ two different types to my knowledge). In the cases that does happen, we can look at `when()` and `if()` for dynamic type validation (but those should be rare).

https://github.com/jquense/yup#mixedwhenkeys-string--arraystring-builder-object--value-schema-schema-schema

So this being said, I will update these current schemas to use a base provider, cluster, datastore, lan, plan, request, etc. and we can adjust these as needed as we find different variations in the models coming back... (I think alongside the reducer in the `{Reducer}Validations.js` file is still probably a good place to add specific overrides.

priley86:
refactored this based on comments above, I think this will work quite well for us...when one needs to extend the base schema, it can be done in the specific validator using the `{schema}.concat` method like I mention [above](https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/453#discussion_r199327411). 

Also included a base validation method which can optionally be disabled (however I think we should strive to keep this enabled until the next release approaches).

Let me know if there is any further suggestions here - I'd like to get this in as the "base" for other API schema validations and continue adding those as we have time... 😄

app/javascript/data/models/clusters.js
```diff
@@ -0,0 +1,19 @@
...
+      id: string().required(),
+      ems_id: string().required(),
+      uid_ems: string().required(),
+      ext_management_system: object()
+        .shape({
+          emstype: string().required(),
+          name: string().required()
+        })
+        .required(),
+      v_parent_datacenter: string().required()
```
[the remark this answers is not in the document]
priley86: updated!

priley86:
much appreciated @AllenBW ❤️ ... it's just another form of UH MERICA patriotic service 🇺🇸 🎆 🇺🇸

PR #454, opened 2018-06-29
Only iterate through mostRecentRequests if it exists 😋

### LINKED ISSUE #452: Viewing archived plans blows up when mostRecentRequests is undefined 😭
wompwomp

AllenBW:
fixes #452

Now yah can see archived plans, and the whole of v2v doesn't crash 😬

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
@@ -33,7 +33,7 @@ const MigrationsCompletedList = ({
             const failed = mostRecentRequest && mostRecentRequest.status === 'Error';
 
             const tasks = {};
-            mostRecentRequest.miq_request_tasks.forEach(task => {
+            if(mostRecentRequest) mostRecentRequest.miq_request_tasks.forEach(task => {
               tasks[task.source_id] = task.status === 'Ok';
             });
 


app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -33,7 +33,7 @@ const MigrationsCompletedList = ({
             const failed = mostRecentRequest && mostRecentRequest.status === 'Error';
 
             const tasks = {};
-            mostRecentRequest.miq_request_tasks.forEach(task => {
+            if(mostRecentRequest) mostRecentRequest.miq_request_tasks.forEach(task => {
```
AparnaKarve: how about -
```js
mostRecentRequest && mostRecentRequest.miq_request_tasks.forEach(task => {
```

like it's done above

AllenBW: intentionally and for a very good reason
screen shot 2018-06-28 at 8 37 13 pm


eslint is not a fan

AparnaKarve: Ok, fair.

AparnaKarve:
@AllenBW Changing this to gaprindashvili/yes, since the bug was introduced by https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/444, which is also gaprindashvili/yes

PR #459, opened 2018-06-29
[GAPRINDASHVILI] adjust linting with `prettier`

AparnaKarve:
Adjust linting with `prettier` linting rules set for the gaprindashvili branch to fix travis

Related PR https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/450


AparnaKarve:
@AllenBW Right, makes sense. Thanks!

PR #461, opened 2018-07-02
[GAPRINDASHVILI] Check for valid `vms` before calculating the in active VMs length

AparnaKarve:
Gaprindashvili equivalent of https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/460


AparnaKarve:
@priley86 @AllenBW Please review.

Had to create a separate G PR to address linting issues on G

PR #462, opened 2018-07-02
Ignore Mike's todolist json file...

mturley:
I started using a command line todo list (check it out, `brew install todolist`) and it stores its data as a file in each repo (separate todo lists for each repo I work on).

By no means do you guys have to merge this. But if you do I won't have to keep seeing it red in my git status...

Alternatively, does anyone know a good solution for clone-specific .gitignore overrides, without committing the changes to .gitignore?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ tmp/
 
 # IDE - VSCode
 .vscode/*
+
+# You guys should check out http://todolist.site/  (brew install todolist)
+.todos.json


.gitignore
```diff
@@ -23,3 +23,6 @@ tmp/
 
 # IDE - VSCode
 .vscode/*
+
+# You guys should check out http://todolist.site/  (brew install todolist)
```
AllenBW: Not opposed to merging, but the shameless plug could be more subtle 😏

priley86:
yea.... this is kinda funny 😆 ...

you might look at [global gitignores](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) rather than your local MIQ friends ;)

AllenBW:
> you might look at global gitignores rather than your local MIQ friends ;)

@mturley you ok if we close this pr? that's the one thing I am opposed to.. orphan prssssss

Issue #437, opened 2018-06-22
All VMs don't show on Migration Plan Request Details Page


kedark3:
See https://bugzilla.redhat.com/show_bug.cgi?id=1594428 for more description

priley86:
@AparnaKarve is fixing this in #374 this sprint.

AllenBW:
@JPrause a dupe of which bz?

JPrause:
@AllenBW see @priley86 comment above.

AllenBW:
@JPrause yah, not seeing another bz # 🤷‍♂️

JPrause:
Any update on this PR. We have an upcoming build on July 10

AparnaKarve:
@JPrause Work is in progress, a PR should be out soon, sometime this week.

vconzola:
Duplicate of #374.

PR #468, opened 2018-07-03
[#437] Resolve all Retry Migration issues

AparnaKarve:
- Updates Progress bar in the main Migration area
- Displays correct number of VMs in the Completed List
- Displays the VMs that were successfully migrated in the previous requests in Plan Detail screen

Fixes https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/437, #469


AparnaKarve:
With @kedark3 's assistance, an appliance was set up where a Plan had 2 out of 3 VMs migrated successfully.
A Retry operation was tried on this appliance for this Plan to validate the PR - results were successful.

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
```diff
@@ -0,0 +1,18 @@
...
+  groupBy: (items, key) =>
+    items.reduce(
+      (result, item) => ({
+        ...result,
+        [item[key]]: [...(result[item[key]] || []), item]
+      }),
+      {}
+    ),
+
+  flattenArray: (arrayItems, flattenedArray) =>
```
AllenBW: wondering why not use https://lodash.com/docs/4.17.4#flatten?  We have already use lodash in the app 🤔

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
```diff
@@ -0,0 +1,18 @@
+const commonUtilitiesHelper = {
+  getMostRecentEntityByCreationDate: entities =>
+    entities.reduce((prev, current) => (prev.created_on > current.created_on ? prev : current)),
+
+  groupBy: (items, key) =>
```
AllenBW: So this function exists in lodash, a tool we use elsewhere in v2v, https://lodash.com/docs/4.17.4#groupBy

Wonder if using ☝️ might be a lil less code

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
priley86: I'm going to have to respectfully disagree here. I think it's preferred to use es6 methods when the functions are trivial and there is not much more to implement using a native function. Individual Lodash imports will soon be stripped by Webpack tree shaking techniques and loaded individually, however I still don't think that's a reason to prefer them (since they often times add more code on the loading end). In these cases I prefer to extend the native functions which removes the need for those imports. I have seen cases where lodash functions do provide a lot of value (`sortBy` for example is non trivial), but these are specific to the use case...

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
priley86: same as above...

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
priley86: I will believe this was mentioned in several Foreman PRs as well...

AllenBW: Ooooo! We disagree indeed. I don't have performance metrics to compare, so I cannot assert one method is "better than another"  🤷‍♂️  *BUT* the less code we have to maintain, watch something go wrong with, the better. Both of these functions have been written before, I see no strong case for us to rewrite them.

☝️ being said, this is a blocker, my personal preference to not rewrite commonly functions will not hinder approval of this

priley86: Yea, obviously we are still loading Lodash and not stripping it...so it's already included and would be minimal impact on this project, but long term I still think it's considered bad practice now by most communities (kind of like using jQuery at this point, we avoid it when we can but obviously we still rely on it in some places). Reference:
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_flatten

If we end up converting things to lodash, it becomes preferential to do what Foreman did w/ the lodash babel plugin. Went ahead and referenced that in [3776](https://github.com/ManageIQ/manageiq-ui-classic/pull/3776#issuecomment-402750123)
but happy to open a separate issue if we care at all about the performance of lodash.

AparnaKarve:
On an unrelated topic to the above discussion, there might be 2 open items to the Retry issue (depending on how we look at it)

Item 1 -
Consider this scenario -
In the first migration attempt, 2 out of 3 VMs  migrated successfully.
When I Retry the same plan, the Migration card in the main screen shows
`0 of 1 VMs migrated`

-- should the Migration card  say `2 of 3 VMs migrated` instead?

Item 2 -
When the Retry Migration completes, what should we show for Elapsed Time in the main Overview page?
Note that, in this scenario the Plan has completed in several attempts (> 1 Request), and the Requests themselves could have been ordered over a span of several minutes/hours/days.

app/javascript/react/screens/App/common/commonUtilitiesHelper.js
AllenBW: Love love the link, thanks for sharing!

AllenBW: Ok mmm I would make one humble request, that we keep the naming consistent, renaming this to `flatten` would be nice 🤔

AparnaKarve: sure, not a problem!

priley86:
Does this also include the fix for the recently opened #469 @AparnaKarve ?

AllenBW:
> -- should the Migration card say 2 of 3 VMs migrated instead?
Sonds very reasonable to me! Migrated implies success, only two of them succeeded,unless we wanta negate the verbiage... 🤔 

Item 2 - yeah this is tricky, the rub is that not all the VMs will be migraine at the same time some will have already been completed while others will still be needing to be completed... Maybe it should be total elapsed time.. Including the retry time..

AparnaKarve:
@priley86 Yes, it includes a fix for #469 -- I will add it to the above comment.

app/javascript/react/screens/App/Plan/PlanReducer.js
```diff
@@ -35,102 +36,88 @@ const excludeDownloadDoneTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
...
-        taskDetails.diskSpaceCompletedGb = '0';
-      }
+  vmTasks.forEach(task => {
+    const taskDetails = {
+      id: task.id,
+      message: task.message,
+      transformation_host_name: task.options && task.options.transformation_host_name,
+      delivered_on: new Date(task.options.delivered_on),
+      updated_on: new Date(task.updated_on),
+      completed: task.message === 'VM Transformations completed' || task.message === 'VM Transformations failed',
```
AllenBW: Is this a string that might change with translation?

priley86: Somewhat related: we actually already have an issue about those... if we want to tackle that in a future PR (or in this one): https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/255

"VM Transformations completed" vs "Migrations completed"

Ideally, we can create some sort of dictionary to map them to the UI translation on the Redux end... I don't know if we tackle this now though...

AparnaKarve: Those strings are coming from the backend -- hopefully the backend has gettexted them.

But you are right, this should be -
```js
completed: task.message === __('VM Transformations completed') || task.message === __('VM Transformations failed'),
```

We can't control the backend, but we can at least do our part.

app/javascript/react/screens/App/Plan/PlanReducer.js
```diff
@@ -35,102 +36,88 @@ const excludeDownloadDoneTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
...
-        options: {}
-      };
-
-      taskDetails.options.progress = task.options.progress;
-      taskDetails.options.virtv2v_wrapper = task.options.virtv2v_wrapper;
-
-      if (!task.diskSpaceCompletedGb) {
-        taskDetails.diskSpaceCompletedGb = '0';
-      }
+  vmTasks.forEach(task => {
```
AllenBW: any chance vmTasks could ever be null/undefined?

AparnaKarve: No, not in PlanReducer - i.e., we are always expected to have tasks in the Plan detail screen.
(PlanReducer is the reducer for Plan detail screen)

app/javascript/react/screens/App/Plan/PlanReducer.js
AparnaKarve: I have also found a couple of incorrect `sprintf` usages - i.e. incorrect in the context of i18n.

I'm thinking of handling this string and a bunch of other strings in a separate PR for addressing all known i18n issues.

AparnaKarve:
Per our discussion, I have kept the elapsed time in the Overview screen to reflect the elapsed time of the most recent request.

PlanB in the below screenshot in it's most recent request had 1 VM to migrate (Webserver-01). The migration for this VM took 4min:38sec

screen shot 2018-07-05 at 2 25 06 pm

The Plan detail page shows the VM, Webserver-01 that took  4min:38sec -

screen shot 2018-07-05 at 2 24 40 pm

/cc @vconzola

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
```diff
@@ -52,12 +53,16 @@ const MigrationsInProgressCard = ({ plan, allRequestsWithTasks, reloadCard, hand
...
-    });
+  let tasksOfPlan = {};
+  if (requestsOfAssociatedPlan.length > 0) {
+    tasksOfPlan = getMostRecentVMTasksFromRequests(requestsOfAssociatedPlan, plan.options.config_info.vm_ids);
+  } else if (mostRecentRequest) {
+    tasksOfPlan = mostRecentRequest.miq_request_tasks;
+  }
+  tasksOfPlan.forEach(task => {
+    tasks[task.source_id] = tasks[task.source_id] || {};
+    tasks[task.source_id].completed = task.status === 'Ok' && task.state === 'finished';
```
priley86: we still haven't extracted these task statuses to constants... since the number of places these are used is increasing, and we have pre/post playbook support upcoming, i think it would be nice to move those to constants in a future PR

priley86:
pulled this locally earlier today...works for me!

AparnaKarve:
@priley86 @AllenBW In the last commit, I have made the necessary adjustments for https://github.com/ManageIQ/manageiq/pull/17674

Should be GTG now.

priley86:
thanks! Will be nice to get this one to QE sooner than later...merging...

AparnaKarve:
@AllenBW Yeah the new structure `options.config_info.actions.vm_id` caught us all by surprise on Friday...since this PR was getting impacted with that change, I had to do some adjustments.

BTW, your old plans won't work anymore because of this new structure - so you would have to delete them all and create new ones.

simaishi:
@AparnaKarve Please create a separate PR for Gaprindashvili due to conflicts...

```
$ git status
On branch gaprindashvili
Your branch is ahead of 'origin/gaprindashvili' by 3 commits.
  (use "git push" to publish your local commits)

You are currently cherry-picking commit 2e91e28.
  (fix conflicts and run "git cherry-pick --continue")
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)

Changes to be committed:

	modified:   app/javascript/components/index.js
	new file:   app/javascript/react/screens/App/Overview/components/Migrations/helpers/getMostRecentVMTasksFromRequests.js
	modified:   app/javascript/react/screens/App/Overview/overview.requestWithTasks.fixtures.js
	modified:   app/javascript/react/screens/App/Overview/overview.transformationPlans.fixtures.js
	modified:   app/javascript/react/screens/App/Plan/Plan.js
	modified:   app/javascript/react/screens/App/Plan/PlanActions.js
	modified:   app/javascript/react/screens/App/Plan/PlanConstants.js
	new file:   app/javascript/react/screens/App/common/commonUtilitiesHelper.js

Unmerged paths:
  (use "git add <file>..." to mark resolution)

	both modified:   app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
	both modified:   app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
	both modified:   app/javascript/react/screens/App/Plan/PlanReducer.js
```

AparnaKarve:
@simaishi Would it be possible to merge https://github.com/ManageIQ/manageiq/pull/17627 and https://github.com/ManageIQ/manageiq/pull/17674 first before I create a G specific PR for this?
Since the last commit in this PR (e6f6846) has adjustments that were made because the above PRs were merged.

If it's not possible to merge the above PRs at this time, I might have to exclude the last commit from this PR and then add it after those PRs are merged.

simaishi:
@AparnaKarve I was holding off backporting those, as PRs/BZs aren't properly linked together and I can't tell which PRs are needed for pre/post feature...  I hope to get them cleared out in the next few days, so let's hold off creating G-specific PR for this until next week.

simaishi:
@AparnaKarve FYI, https://github.com/ManageIQ/manageiq/pull/17627 and https://github.com/ManageIQ/manageiq/pull/17674 are backported to Gaprindashvili branch now.

JPrause:
@AparnaKarve was there further work needed to resolve the conflict.

AparnaKarve:
Gaprindashvili specific PR - https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/521

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/521

PR #479, opened 2018-07-10
Fix the Archived Plan view

AparnaKarve:
Fixes the Archived Plan view broken since #468.

@priley86 Thanks for spotting the bug!

Clubbing this PR under - https://bugzilla.redhat.com/show_bug.cgi?id=1592956


priley86:
confirmed this fixes the issue I was having locally, thanks @AparnaKarve

AparnaKarve:
@simaishi This will be backported via #521, hence setting `gaprindashvili/no` here.

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/521

PR #487, opened 2018-07-11
Undo the code that puts the master commit hash in the html

AparnaKarve:
Removing the changes in #436 since it really does not serve any purpose.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1597249


priley86:
ok - looks like the plan is revisit #422 more formally later? No problem...

PR #401, opened 2018-06-08
[#369] Inactive migration plans can be scheduled to run 

AllenBW:
fixes #369

### Where we're at... 
yes

this one's dedicated to @bdunne @gtanzillo @abellotti

PR #488, opened 2018-07-13
add schedules schema validation to overview

priley86:
piggybacking on #401. Adds the schedules schema validation. 🚥

### DIFF
--- a/app/javascript/data/models/schedules.js
+++ b/app/javascript/data/models/schedules.js
@@ -1,0 +1,11 @@
+import { string, object } from 'yup';
+
+export const schedulesSchema = object().shape({
+  href: string().required(),
+  id: string().required(),
+  run_at: object()
+    .shape({
+      start_time: string().required()
+    })
+    .required()
+});
--- a/app/javascript/data/models/plans.js
+++ b/app/javascript/data/models/plans.js
@@ -1,4 +1,5 @@
 import { array, string, object } from 'yup';
+import { schedulesSchema } from './schedules';
 
 export const plansSchema = array()
   .of(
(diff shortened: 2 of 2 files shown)


app/javascript/data/models/schedules.js
```diff
@@ -0,0 +1,11 @@
+import { string, object } from 'yup';
+
+export const schedulesSchema = object().shape({
+  href: string().required(),
+  id: string().required(),
+  run_at: object()
```
AllenBW: So the full object looks kinda like:

{
href:'foo'
schedules:[
{run_at:{
start_time: 'foo'
}}]}

priley86: i am seeing:
```
      "schedules": [
        {
          "href": "http://0.0.0.0:8080/api/service_templates/5/schedules/31",
          "id": "31",
          "name": "Order ServiceTemplateTransformationPlan 5",
          "description": "Order ServiceTemplateTransformationPlan 5",
          "sched_action": {
            "args": [
              1,
              {
                
              },
              {
                "submit_workflow": true
              }
            ],
            "method": "queue_order"
          },
          "filter": null,
          "towhat": "ServiceTemplate",
          "run_at": {
            "interval": {
              "unit": "once"
            },
            "start_time": "2018-07-21T14:40:00Z",
            "tz": "UTC"
          },
```

Only attempting to get the fields the UI is using, but if you'd like to add more, fine w/ me ;)

AllenBW: OH gotcha ok, yeah that's right, apologies, i thought this was reaching from an encapsulating service_template object 👍

PR #417, opened 2018-06-15
Use a new rbac feature id for the menu item.

martinpovolny:
To be used together with https://github.com/ManageIQ/manageiq/pull/17596

PR #490, opened 2018-07-16
Remove Settings.product.transformation - no longer conditional on a product feature

### LINKED ISSUE #432: Remove product->transformation setting
The product-> transformation setting in settings.yml is a toggle for disabling the feature while still dev-incomplete.  On master, this is no longer the case, so we should remove this flag, as we don't keep these extra product settings for officially released features.  We can wait until backporting dies down a bit in order to avoid backport issues.

cc @himdel @martinpovolny

himdel:
We're using rbac now, from #417

Fixes #432

@Fryguy , @priley86 I have no idea about the gaprindashvili status of this? Should the new builds have it enabled too, or is the product feature removal hammer only?

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,11 +1,9 @@
 # ManageIQ::V2V::Engine.routes.draw do
-if ::Settings.product.transformation == true
-  Rails.application.routes.draw do
-    match '/migration' => 'migration#index', :via => [:get]
-    match 'migration/*page' => 'migration#index', :via => [:get]
+Rails.application.routes.draw do
+  match '/migration' => 'migration#index', :via => [:get]
+  match 'migration/*page' => 'migration#index', :via => [:get]
 
-    get "migration_log/download_migration_log(/:id)",
-      :action     => 'download_migration_log',
-      :controller => 'migration_log'
-  end
+  get "migration_log/download_migration_log(/:id)",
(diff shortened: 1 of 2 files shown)


priley86:
i believe this was also slighted for G... ping @bthurber ...

AllenBW:
We should hold off on merging this one till https://github.com/ManageIQ/manageiq/pull/17711 gets in.

himdel:
@AllenBW No no, it needs to be the other way around.

https://github.com/ManageIQ/manageiq/pull/17711 can't be merged until this one is, otherwise v2v will just disappear :).

AllenBW:
Merging, looks to be working just peachy! (user missing product feature, doesn't see Migration nav element)

screen shot 2018-07-16 at 10 00 55 am

AparnaKarve:
@simaishi The `gaprindashvili/yes` label here and in https://github.com/ManageIQ/manageiq/pull/17711 is questionable. 
(not sure who makes the decision on that)

I think the changes in this PR are meant for master only.

Before backporting this PR and https://github.com/ManageIQ/manageiq/pull/17711, let us first find out what exactly has been planned for the `Settings.product.transformation` flag.
Thanks.

himdel:
Answered in https://github.com/ManageIQ/manageiq/pull/17711#issuecomment-405612660, the label is correct.

PR #514, opened 2018-07-25
[#515]Adds migration scheduling to complete mapping list

### LINKED ISSUE #515: Can not schedule migrations after failure (retry page)
**User case**

1. Create infra map `map1`
2. Attach mapping to `plan1`
3. Schedule `plan1` for X time
4. Delete plan `map1` before X time

Here when migration get triggered, it fails (as obvious, since it does not find `map1`)
If you check Completed Plan section, you will find only two options `Retry` and `Archive` are there (Not `Schedule migration`). 

So overall, we can not schedule failed (retry) migration

AllenBW:
Fixes #515
### What we look like
screen shot 2018-07-27 at 10 41 41 am

OK OK so there's an issue with the using the kebab [(described here)](https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/514#issuecomment-408263817) so we're using buttons for now. We now hide "stale migrations" from the completed list, those migrations that happened in the past, and allow the user to schedule anew **but** there's an issue with the api serializing the keys (?) or something like that, so rescheduling an already failed migration doesn't so much work @bdunne is looking into da issue... 

### ~its a long video, but worth a watch!~ Its now an old video, but still worth watching
asdfassssss
~Edit: if this seems clunky, having to unschedule to re schedule, s' something that can be improved upon, but I don't want to make any promises before that it could be done by friday, so here seems like a good stopping spot... 🤞~

### DIFF
--- a/app/javascript/react/screens/App/Overview/OverviewActions.js
+++ b/app/javascript/react/screens/App/Overview/OverviewActions.js
@@ -245,31 +245,41 @@ export const archiveTransformationPlanAction = (url, id) => {
   return _archiveTransformationPlanActionCreator(uri.toString());
 };
 
-export const toggleScheduleMigrationModal = planId => ({
+export const toggleScheduleMigrationModal = plan => ({
   type: V2V_TOGGLE_SCHEDULE_MIGRATION_MODAL,
-  payload: planId
+  payload: plan
 });
 
 export const scheduleMigration = payload => dispatch =>
   dispatch({
     type: V2V_SCHEDULE_MIGRATION,
     payload: new Promise((resolve, reject) => {
-      let url = `/api/service_templates/${payload.planId}`;
+      const {
+        scheduleTime,
+        plan: { id: planId }
+      } = payload;
+      const scheduleId = (payload.plan.schedules && payload.plan.schedules[0].id) || null;
+      let url = `/api/service_templates/${planId}`;
       let body = {
         action: 'order',
         resource: {
-          schedule_time: payload.scheduleTime
+          schedule_time: scheduleTime
         }
       };
-      if (payload.scheduleId) {
-        url = `${url}/schedules/${payload.scheduleId}`;
-        body = { action: 'delete' };
+      if (scheduleId) {
+        url = `${url}/schedules/${scheduleId}`;
+        body = scheduleTime
+          ? {
+              action: 'edit',
+              resource: { run_at: { ...payload.plan.schedules[0].run_at, start_time: scheduleTime } }
+            }
+          : { action: 'delete' };
       }
       return API.post(url, body)
         .then(response => {
           resolve(response);
           let msg = __('Migration successfully unscheduled');
-          if (payload.scheduleTime) {
+          if (scheduleTime) {
             msg = sprintf(
               __('Migration successfully scheduled for %s'),
               formatDateTime(response.data.run_at.start_time)
--- a/app/javascript/react/screens/App/Overview/OverviewReducer.js
+++ b/app/javascript/react/screens/App/Overview/OverviewReducer.js
@@ -100,7 +100,7 @@ export const initialState = Immutable({
   isFetchingNetworks: false,
   isRejectedNetworks: false,
   scheduleMigrationModal: false,
-  scheduleMigrationPlanId: null,
+  scheduleMigrationPlan: {},
   isSchedulingMigration: false,
   isRejectedSchedulingMigration: false,
   errorSchedulingMigration: false,
@@ -382,10 +382,10 @@ export default (state = initialState, action) => {
     case V2V_TOGGLE_SCHEDULE_MIGRATION_MODAL:
       if (action.payload !== undefined) {
         return state
-          .set('scheduleMigrationPlanId', action.payload.planId)
+          .set('scheduleMigrationPlan', action.payload.plan)
           .set('scheduleMigrationModal', !state.scheduleMigrationModal);
       }
-      return state.set('scheduleMigrationPlanId', null).set('scheduleMigrationModal', !state.scheduleMigrationModal);
+      return state.set('scheduleMigrationPlan', null).set('scheduleMigrationModal', !state.scheduleMigrationModal);
     case `${V2V_SCHEDULE_MIGRATION}_PENDING`:
       return state.set('isSchedulingMigration', true).set('isRejectedSchedulingMigration', false);
     case `${V2V_SCHEDULE_MIGRATION}_FULFILLED`:
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsNotStartedList.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsNotStartedList.js
@@ -7,6 +7,7 @@ import ScheduleMigrationModal from '../ScheduleMigrationModal/ScheduleMigrationM
 import { formatDateTime } from '../../../../../../components/dates/MomentDate';
 import { MIGRATIONS_NOT_STARTED_SORT_FIELDS } from './MigrationsConstants';
 import sortFilter from '../../../Plan/components/sortFilter';
+import ScheduleMigrationButton from './ScheduleMigrationButton';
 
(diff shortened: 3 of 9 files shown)


AllenBW:
@Yadnyawalkya should this be gaprindashvili/yes? piggy back off that same bz?

vconzola:
@AllenBW I'm a little confused as to why Unschedule is required before the plan can be scheduled again. Does running the plan not get rid of its schedule? I think this will be very confusing to the user. If an Unschedule button needs to appear for every completed plan, then I don't want to merge this before GA. Let's take the time to do it right.

AllenBW:
@vconzola no need for confusion,  When we have a schedule on the record, something that was used to kick off a migration, it stays on the record until removed. But that makes sense, we might want to know when the last migration was scheduled to run, or maybe not?

UX guidance on how to "do this right" would be most appreciated.  Do we clear the past schedule upon migration completion, losing that record? Provide an edit button to modify the existing schedule? Do we ignore all past schedules and treat it as a new one?

vconzola:
@AllenBW I think the best UX is to clear the past schedule upon migration completion so that the options on a Completed plan become Schedule, Retry, and Archive with Archive being hidden under a kebab. I'll create a mockup.

vconzola:
@AllenBW Thinking about this does make me think of another edge case, however. Any idea what happens today if a scheduled migration fails to run for some reason, e.g., network problem? Do we indicate that in the UI at all? Leads me to think that we need an error in the plan list view and also to prioritize rescheduling migrations (vs. unschedule and schedule again like we have today).

AllenBW:
@vconzola 👍 ok can clear upon completion, will have that change up in a few shakes. To your other question I believe this would fail just as a normal migration, to my knowledge we aren't provided any indication as to why a schedule failed to run.

vconzola:
@AllenBW I'm talking about the case where the request is never sent so the migration never starts at the scheduled time. Is this possible and a case we need to consider?

AllenBW:
@vconzola Not sure I follow, the request is never sent? All api errors are handled by miq, so if theres an issue with setting the schedule we'll know then.  If your saying what happens in the case the schedule fails to run? Not sure we get any feedback to that effect.

vconzola:
@AllenBW That's the case I was thinking of - the schedule fails to run. Do we need to address this possibility?

AllenBW:
@vconzola how abouts we chat this over in the standup 👍

vconzola:
@AllenBW Here's a mockup with a schedule button added for failed Completed plans and the Archive feature added under a kebab.
overview - w all copy 4 filter is completed with archive in kebab

mturley:
Late to the party, but this was a red flag for me:

> Do we clear the past schedule upon migration completion, losing that record?

I agree with @vconzola that the remaining options should be Schedule, Retry, and Archive, and we shouldn't have to unschedule. But I don't know how I feel about losing the history of what happened. Maybe a new feature for down the line should be Migration History?  I can just imagine a user realizing something was wrong with a completed migration and wanting to go look at how it was configured.

AllenBW:
@vconzola 

Archive as a kebab is a problem because we have a whole list item wrapped in a click event.  So when you click on the kebab, rather than seeing the dropdown you are redirected to the completed migration. Working on getting the kebab to behave like the rest of the buttons, independent of the row click event, but just in case thats not immediately possible... we can either,

1. change the completed migration name into the redirect to the page
2. swap a kebab for the tried and true button till this gets resolved

mturley:
@AllenBW have you tried stopping propagation on the kebab's event handler?  `event.stopPropagation()` is your friend when you have event handlers on nested elements. https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

AllenBW:
@mturley yes. and about a billon other things.

AllenBW:
https://github.com/ManageIQ/manageiq-api/pull/436 is the pr we need for this to work...

Is ready for review from other parties whenever they can spare the cycles

app/javascript/react/screens/App/Overview/OverviewReducer.js
```diff
@@ -100,7 +100,7 @@ export const initialState = Immutable({
   isFetchingNetworks: false,
   isRejectedNetworks: false,
   scheduleMigrationModal: false,
-  scheduleMigrationPlanId: null,
+  scheduleMigrationPlan: {},
```
michaelkro: I think you want to set the default as an empty string here due to [this line](https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/514/files#diff-a64dbfef583d0adfd0466639743beb9cR194)

AllenBW: eeeeek nope, good catch, that needs to be an object, cuz the plan is an object

app/javascript/react/screens/App/Overview/OverviewActions.js
```diff
@@ -245,31 +245,41 @@ export const archiveTransformationPlanAction = (url, id) => {
...
 export const scheduleMigration = payload => dispatch =>
   dispatch({
     type: V2V_SCHEDULE_MIGRATION,
     payload: new Promise((resolve, reject) => {
-      let url = `/api/service_templates/${payload.planId}`;
+      const {
+        scheduleTime,
+        plan: { id: planId }
+      } = payload;
+      const scheduleId = (payload.plan.schedules && payload.plan.schedules[0].id) || null;
```
mturley: This will be a `TypeError` if `payload.plan.schedules` is an empty array (truthy but `[0]` is undefined). Not sure if we'll ever hit that case, but we should maybe do:

```
const firstSchedule = payload.plan.schedules && payload.plan.schedules[0];
const scheduleId = firstSchedule ? firstSchedule.id : null;
```

app/javascript/react/screens/App/Overview/OverviewActions.js
```diff
@@ -245,31 +245,41 @@ export const archiveTransformationPlanAction = (url, id) => {
...
       };
-      if (payload.scheduleId) {
-        url = `${url}/schedules/${payload.scheduleId}`;
-        body = { action: 'delete' };
+      if (scheduleId) {
+        url = `${url}/schedules/${scheduleId}`;
+        body = scheduleTime
+          ? {
+              action: 'edit',
+              resource: { run_at: { ...payload.plan.schedules[0].run_at, start_time: scheduleTime } }
```
mturley: A nice DRY `firstSchedule` could be used here as well

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsNotStartedList.js
```diff
@@ -130,30 +94,16 @@ class MigrationsNotStartedList extends React.Component {
...
-                                onClick={e => {
-                                  e.stopPropagation();
-                                  showConfirmModalAction(confirmModalOptions);
-                                }}
-                                disabled={loading === plan.href}
-                              >
-                                {__('Unschedule')}
-                              </Button>
-                            )}
+                            <ScheduleMigrationButton
```
mturley: Nice!

app/javascript/react/screens/App/Overview/OverviewActions.js
AllenBW: if it exists, as in if ` payload.plan.schedules` passes, there will be something inside 👍

michaelkro:
Looks like https://github.com/ManageIQ/manageiq-api/pull/436 has been merged.  I've looked at the new kebab code, and it looks good to me.

In the interest of time, I've gone ahead and taken a look at the CI failure.  We can replace [this line](https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/components/Migrations/__test__/MigrationsCompletedList.test.js#L62) with something like

```js
const archiveButton = item.find('a').filterWhere(link => link.text() === 'Archive');
```

to fix the failing test case

AllenBW:
Updated! @michaelkro

michaelkro:
🎉

PR #517, opened 2018-07-26
[#370] Cancel a migration's in-progress tasks

### LINKED ISSUE #370: User can cancel a migration in progress
User can cancel a migration.  Need to finalize the requirements here:
- is cancel action only enabled/available for migrations in progress
- cancel confirmation should be shown

AparnaKarve:
Fixes #370

screen shot 2018-07-25 at 6 49 17 pm

---------------------

screen shot 2018-07-27 at 7 53 44 am

-----------------------

screen shot 2018-07-25 at 6 56 49 pm

This is a continuation to @priley86 's and @mturley 's work...
Includes -

- Canceling Migration Tasks
- Displays the `fa-ban` icon for canceled tasks
- Notification about the the Cancellation
- Managing the checkboxes and `Cancel Migration` button state appropriately

https://bugzilla.redhat.com/show_bug.cgi?id=1564257

To be merged after https://github.com/ManageIQ/manageiq-api/pull/421


app/javascript/react/screens/App/Plan/PlanActions.js
```diff
@@ -206,3 +207,36 @@ export const downloadLogAction = task => dispatch => {
...
+      API.post(url, {
+        action: 'cancel',
+        resources
+      })
+        .then(response => {
+          resolve(response);
+
+          const cancelSuccessMsg = __('Migration canceled successfully for VMs: ');
+          const vmNames = [];
+          tasks.map(task => vmNames.push(task.vmName));
```
michaelkro: I think here, we can just do

```js
const vmNames = tasks.map(task => task.vmName)
```

app/javascript/react/screens/App/Plan/PlanActions.js
```diff
@@ -206,3 +207,36 @@ export const downloadLogAction = task => dispatch => {
...
+          const cancelSuccessMsgWithVMNames = `${cancelSuccessMsg} ${vmNames}`;
```
michaelkro: Minor nit, but if we pass an array (with multiple entries) into the template literal, there will not be spaces after each comma

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
```diff
@@ -270,6 +277,70 @@ class PlanRequestDetailList extends React.Component {
...
+  onCancelMigrationsClick = () => {
+    this.setState({ showConfirmCancel: true });
+  };
+
+  onCancelMigrationsConfirm = () => {
+    // todo: call this from the Confirmation Modal instead
+    // gather the selected tasks in state and feed them through
+    const { selectedCancelTasks } = this.state;
+    const { cancelPlanRequestTasksAction, cancelPlanRequestTasksUrl } = this.props;
+    cancelPlanRequestTasksAction(cancelPlanRequestTasksUrl, selectedCancelTasks);
```
michaelkro: We might want to refetch the tasks here.  Currently, after confirming the cancel, the success notification appears, but we need to wait for the next poll before the ListItem is updated

AllenBW:
Ahhhh that's the missing piece, thanks @AparnaKarve !

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
```diff
@@ -270,6 +277,70 @@ class PlanRequestDetailList extends React.Component {
...
+    this.setState({ showConfirmCancel: false });
```
AllenBW: Modal window isn't hiding after firing this 😕

AllenBW:
It might be something wacky I did... or desired behavior... but after canceling two in progress migrations... they get the plan appears to be stuck in inprogress purgatory...  meaning... it just hangs out inprogress tab... with no action aside from download logs available

AparnaKarve:
I have noticed some strange behavior with Cancel locally as well... hence I think the real valid test of canceling a migration should be done on an appliance - it seemed to work a lot better on an appliance yesterday.

So based on the testing that can be done locally, I will try and fix some of the UI/UX issues that @michaelkro noticed above, but the bulk of the testing needs to be done on an appliance.

I will try and setup an appliance sometime today for testing this.

vconzola:
@michaelkro I concur with @AparnaKarve expected behavior comment above.

app/javascript/react/screens/App/Plan/PlanActions.js
mturley: Agreed, it might be worth doing something like this:

```
const vmNamesStr = vmNames.join(', ');
const cancelSuccessMsgWithVMNames = `${cancelSuccessMsg} ${vmNamesStr}`;
```

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
```diff
@@ -520,6 +657,23 @@ class PlanRequestDetailList extends React.Component {
...
+            <React.Fragment>
+              <p>
+                {__('If you cancel, these VMs will not be migrated and will be fully restored in the source provider.')}
+              </p>
+              <ul>{selectedCancelTasks.map(task => <li key={task.id}>{task.vmName}</li>)}</ul>
+              <p>{__('Do you want to cancel?')}</p>
+            </React.Fragment>
+          }
+          cancelButtonLabel={__('No')}
+          confirmButtonLabel={__('Cancel Migrations')}
```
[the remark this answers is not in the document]
vconzola: @AllenBW This is a classic UX conundrum - how to confirm an action when that action is cancelling something. The normal button labels (OK, Cancel) lead to confusion - Does Cancel cancel or cancel the Cancel. So I ran this by Stacey (our info architect) and this is what we came up with. Maybe not consistent, but in this case necessary.

mturley:
Working on some of the quirky behavior around the checkboxes, since it's related to stuff I did. I'll have a commit for @AparnaKarve to cherry-pick shortly

mturley:
Posting this here for posterity, it might conflict with @AparnaKarve's redux stuff, but it should fix some behavior around the bulk selector: https://github.com/mturley/miq_v2v_ui_plugin/commit/a65868d17984a92e710ca5965d9fb5867d905eab

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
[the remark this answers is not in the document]
mturley: @AllenBW @vconzola Just my $0.02, but I feel like the single-word restriction is specifically confusing in this "cancelling the cancel" case. I've been happy with products that I see use the opposite tactic when one word isn't enough-- very explicit and verbose buttons.

"Yes, cancel these migrations"   /   "No, do not cancel anything"   or something to that effect.

I think using one word where it's not confusing, and then just going straight for a sentence otherwise instead of trying to find the two word button that works, is good.  End of non-designer opinion.

AparnaKarve:
After implementing the redux approach, came across a few use cases where the user can sneak in a Cancel on an already canceled task (in the brief window until the next polling happens)

Trying to address some of these cases...

app/javascript/react/screens/App/Plan/PlanActions.js
AparnaKarve: Thanks for the tip.
Addressed this in the last commit

app/javascript/react/screens/App/Plan/PlanConstants.js
```diff
@@ -29,7 +30,8 @@ export const STATUS_MESSAGE_KEYS = {
   VALIDATING: 'Validating',
   VM_MIGRATIONS_COMPLETED: 'VM Transformations completed',
   VM_MIGRATED: 'Virtual machine migrated',
-  VM_MIGRATIONS_FAILED: 'VM Transformations failed'
+  VM_MIGRATIONS_FAILED: 'VM Transformations failed',
+  CANCELLED: 'VM cancelled'
```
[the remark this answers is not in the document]
AparnaKarve: I have removed this constant for now, since we don't know yet what string the state machine actually uses for a cancelled VM.

This will become more clear after some appliance testing and once Fabien's cancellation implementation is complete

app/javascript/react/screens/App/Plan/PlanReducer.js
```diff
@@ -79,6 +81,10 @@ const processVMTasks = vmTasks => {
       taskDetails.options.playbooks = task.options.playbooks;
     }
 
+    if (task.options.cancel_requested) {
```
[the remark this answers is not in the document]
AparnaKarve: done in the last commit

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
AparnaKarve: Can you check again? The modal seems to be hiding OK for me.

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
AparnaKarve: did not do a refetch here.

The listview reflects the cancelled state immediately upon a successful return from the API POST `cancel` action.
I have used `markedForCancellation` redux state to mark the cancelled items in the listview before the task is officially marked with `cancel_requested=true` by the API.

AparnaKarve:
@michaelkro @AllenBW @mturley Please test/review when you get a chance.

> the plan appears to be stuck in inprogress purgatory... meaning... it just hangs out inprogress tab

@AllenBW We usually don't see this issue on appliances... although I have tried to address this in c3fe254, 
86acf0d and 
407f8b9

I will try and setup an appliance to test this, but note that, currently migrations are failing on the appliances in the "Pre-Migration" step - we need an automate PR to fix this

mturley:
Thank you so much for your hard work on this @AparnaKarve.

app/javascript/react/screens/App/Plan/PlanActions.js
```diff
@@ -206,3 +211,81 @@ export const downloadLogAction = task => dispatch => {
...
   });
 };
+
+// *****************************************************************************
+// * CANCEL MIGRATION TASKS
+// *****************************************************************************
+export const cancelPlanRequestTasksAction = (url, tasks) => dispatch => {
+  const completedTasks = [];
+  const incompleteTasks = [];
+  tasks.map(t => (t.completed ? completedTasks.push(t) : incompleteTasks.push(t)));
```
michaelkro: While this ultimately has the desired effect, I think we may want to use `forEach` here.  In my experience, we use `map` when we want to do something with the returned array.  In this case, it seems that we are iterating through `tasks` (rather than mapping it to a new array)

mturley: Since you don't use `map`'s return value here, it could be a `forEach` instead.. but it really doesn't matter, and if this is the only thing I can find to comment on, I think your code looks great!

app/javascript/react/screens/App/Plan/PlanReducer.js
```diff
@@ -50,21 +56,32 @@ const excludeDownloadDoneTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
 const includeDownloadInProgressTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
   allDownloadLogInProgressTaskIds ? allDownloadLogInProgressTaskIds.concat(taskId) : [taskId];
 
+const removeCancelledTasksFromSelection = (allSelectedTasksForCancel, alreadyCancelledTasks) => {
```
mturley: This is fine, but a little odd looking, and will run in n^2 time.  Here's my approach on this one, taking the map out and only using one `filter`, and then using `every` which will short circuit if it sees a non-matching element.

Not tested, but:
```
const removeCancelledTasksFromSelection = (allSelectedTasksForCancel, alreadyCancelledTasks) => (
  allSelectedTasksForCancel.filter(selectedTask =>
    alreadyCancelledTasks.every(cancelledTask => selectedTask.id !== cancelledTask.id)
  )
);
```

Not that we need to optimize for runtime here, but I think that's more readable too.

app/javascript/react/screens/App/Plan/PlanActions.js
mturley: Beat me to it!

mturley:
@michaelkro I think that's the bug I thought we were seeing at first yesterday, which should be fixed by my commit here in the old implementation: https://github.com/mturley/miq_v2v_ui_plugin/commit/a65868d17984a92e710ca5965d9fb5867d905eab.  @AparnaKarve I agree that it's not a blocker, but you could port this line from my commit into your logic to fix this.

```
const allSelected = !noneSelected && selectedCancelTasks.length === incompleteTasks.length;
```

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
```diff
@@ -270,6 +276,74 @@ class PlanRequestDetailList extends React.Component {
...
+  getCancelSelectionState = () => {
+    const { selectedTasksForCancel, markedForCancellation } = this.props;
+    const filteredTasks = this.filterPlanRequestTasks();
+    const incompleteTasks = filteredTasks.filter(
+      task => !task.completed && !task.cancel_requested && !markedForCancellation.find(t => t.id === task.id)
+    );
+    return {
+      filteredTasks,
+      incompleteTasks,
+      allSelected: selectedTasksForCancel.length === incompleteTasks.length,
```
mturley: Ah, @AparnaKarve I didn't notice you based your work on a slightly older version of `getCancelSelectionState`.  I pushed a newer one on Wednesday that fixed the bug @michaelkro was just reporting (select all checkbox is checked when there are no selectable tasks).

We just need to pull `noneSelected` up into a variable, and then do `allSelected: !noneSelected && ...`

mturley: Ah, @AparnaKarve I didn't notice you based your work on a slightly older version of getCancelSelectionState. I pushed a newer one on Wednesday that fixed the bug @michaelkro was just reporting (select all checkbox is checked when there are no selectable tasks).

We just need to pull noneSelected up into a variable, and then do allSelected: !noneSelected && ...

michaelkro:
I've been trying (unsuccessfully) to find what is making this happen, but it looks like the listview resorts after a cancellation request (after the next poll)

resort-gif

AparnaKarve:
Ok, so just a reminder that we need the API PR merged in first before this.

While I have some time until the API PR is merged, let me check what more I can address from the above comments...
(plus squashing some commits is also still TBD)

@michaelkro The resorting thing is independent of this PR (KK also mentioned to me about it once) - so yes, it could be handled as a separate issue.

mturley:
Code all looks good to me, I haven't tested your new commits locally yet since yesterday morning though. While you're wrapping up I'll see if I can find any more issues.

app/javascript/react/screens/App/Plan/PlanActions.js
AparnaKarve: Addressed in the last commit

app/javascript/react/screens/App/Plan/PlanReducer.js
AparnaKarve: Good suggestion.

I have applied this change in the last commit.

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
AparnaKarve: Fixed the bulk selector issues in the last commit

AparnaKarve:
Thanks everyone for the reviews and testing!

I have done yet another round of addressing feedback - 5f9c89b
... but I am gonna stop now.

Currently keeping an eye on the API PR... will let you all know when this can be merged.

app/javascript/react/screens/App/Plan/PlanReducer.js
```diff
@@ -56,13 +56,10 @@ const excludeDownloadDoneTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
 const includeDownloadInProgressTaskId = (allDownloadLogInProgressTaskIds, taskId) =>
   allDownloadLogInProgressTaskIds ? allDownloadLogInProgressTaskIds.concat(taskId) : [taskId];
 
-const removeCancelledTasksFromSelection = (allSelectedTasksForCancel, alreadyCancelledTasks) => {
-  alreadyCancelledTasks.map(
-    cancelledTask =>
-      (allSelectedTasksForCancel = allSelectedTasksForCancel.filter(element => element.id !== cancelledTask.id))
+const removeCancelledTasksFromSelection = (allSelectedTasksForCancel, alreadyCancelledTasks) =>
+  allSelectedTasksForCancel.filter(selectedTask =>
+    alreadyCancelledTasks.every(cancelledTask => selectedTask.id !== cancelledTask.id)
```
mturley: 🎉  I love the lesser-known `Array` methods :)

app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
```diff
@@ -285,7 +285,7 @@ class PlanRequestDetailList extends React.Component {
     return {
       filteredTasks,
       incompleteTasks,
-      allSelected: selectedTasksForCancel.length === incompleteTasks.length,
+      allSelected: selectedTasksForCancel.length === incompleteTasks.length && selectedTasksForCancel.length > 0,
```
mturley: 👍

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/525

PR #522, opened 2018-07-30
Adds ids to progress cards and labels for automation purposes

### LINKED ISSUE #415: [QEDevCollab] Need ids for in-progress cards
Hello,

Need two ids at following locations
screenshot from 2018-06-15 19-38-17

1. `<div id="progress-bar-items" class="col-xs-12">` to select div items

2. For fedora2 card we should have```<div id="progress-bar-item" class="card-pf card-pf-match-height in-progress" style="position: relative; height: 261px;">``` 
For ubuntu1 card we should have also have same ```<div id="progress-bar-item" class="card-pf card-pf-match-height in-progress" style="position: relative; height: 261px;">```

Please note here in second case, each item of that div items (first case) should have same ` id="progress-bar-item" ` .

I required it to select div of items and single-single item in it via automation

AllenBW:
fixes #415 

screen shot 2018-07-30 at 4 44 23 pm

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/InProgressCard.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/InProgressCard.js
@@ -4,7 +4,12 @@ import { Card, Grid, noop } from 'patternfly-react';
 
 const InProgressCard = ({ title, children, onClick, ...props }) => (
   <Grid.Col sm={12} md={6} lg={4}>
-    <Card matchHeight onClick={onClick} {...props}>
+    <Card
+      id={`${Array.isArray(title.props.children) ? title.props.children[1] : title.props.children}-progress-card`}
+      matchHeight
+      onClick={onClick}
+      {...props}
+    >
       <Card.Heading>{title}</Card.Heading>
       <Card.Body>{children}</Card.Body>
     </Card>
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
@@ -170,13 +170,18 @@ const MigrationsInProgressCard = ({
   // Tooltips
   const vmBarLabel = (
     <span>
-      <strong className="label-strong">{sprintf(__('%s of %s VMs'), completedVMs, totalVMs)}</strong> {__('migrated')}
+      <strong id="vms-migrated" className="label-strong">
+        {sprintf(__('%s of %s VMs'), completedVMs, totalVMs)}
+      </strong>{' '}
+      {__('migrated')}
     </span>
   );
 
   const diskSpaceBarLabel = (
     <span>
-      <strong className="label-strong">{sprintf(__('%s of %s'), totalMigratedDiskSpaceGb, totalDiskSpaceGb)}</strong>{' '}
+      <strong id="size-migrated" className="label-strong">
+        {sprintf(__('%s of %s'), totalMigratedDiskSpaceGb, totalDiskSpaceGb)}
+      </strong>{' '}
       {__('migrated')}
     </span>
   );


AllenBW:
cc @Yadnyawalkya 

card ids follow the pattern `migrationName-progress-card` and the progress labels are as requested, `size-migrated` and `vms-migrated`

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
```diff
@@ -170,13 +170,18 @@ const MigrationsInProgressCard = ({
   // Tooltips
   const vmBarLabel = (
     <span>
-      <strong className="label-strong">{sprintf(__('%s of %s VMs'), completedVMs, totalVMs)}</strong> {__('migrated')}
+      <strong id="vms-migrated" className="label-strong">
```
michaelkro: I think an `id` has to be unique in the page

kedark3: Negative. current one looks good. "vms-migrated" is good enough. no need to make it unique.

AllenBW: 🤔 oh hmmm yeah... also no spaces https://www.w3.org/TR/html5/dom.html#global-attributes  updating...

AllenBW: so yeah, these id are unique to their tree, but unsure about the id with spaces... it doesn't seem to cause much trouble 🔨

AllenBW: @kedark3 does an id with spaces cause you any grief? is this something we ought to change?

app/javascript/react/screens/App/Overview/components/Migrations/InProgressCard.js
```diff
@@ -4,7 +4,7 @@ import { Card, Grid, noop } from 'patternfly-react';
 
 const InProgressCard = ({ title, children, onClick, ...props }) => (
   <Grid.Col sm={12} md={6} lg={4}>
-    <Card matchHeight onClick={onClick} {...props}>
+    <Card id={`${title}-progress-card`} matchHeight onClick={onClick} {...props}>
```
michaelkro: I see your screenshot, but for some reason, my browser does this

devtools_-_localhost_8080_migration

AllenBW: I suppose it would help I pushed my work 😆 This will fix yah

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
kedark3: @AllenBW Have not come across this issue before. So only time will tell. But I hope it won't cause any issues to have spaces.

michaelkro: One last thought, and then I'll leave it alone

https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById

> The id global attribute defines a unique identifier (ID) which must be unique in the whole document.

Also, if I try `document.getElementById('vms-migrated')`, it only ever returns the first one

app/javascript/react/screens/App/Overview/components/Migrations/InProgressCard.js
```diff
@@ -4,7 +4,7 @@ import { Card, Grid, noop } from 'patternfly-react';
 
 const InProgressCard = ({ title, children, onClick, ...props }) => (
   <Grid.Col sm={12} md={6} lg={4}>
-    <Card matchHeight onClick={onClick} {...props}>
+    <Card id={`${title.props.children}-progress-card`} matchHeight onClick={onClick} {...props}>
```
michaelkro: It seems that `title.props.children` is an array

devtools_-_localhost_8080_migration

AllenBW: how'd yah manage that?

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsInProgressCard.js
AllenBW: Yeah this makes sense, not unexpected, I believe when qe tests though they traverse down the tree, one reason why we need two ids here (I'll eat 🍝 if wrong @kedark3 😬 )

app/javascript/react/screens/App/Overview/components/Migrations/InProgressCard.js
michaelkro: ah, maybe in yours, the first array element is null, so you don't see what I'm seeing.  Check out the third console log

devtools_-_localhost_8080_migration

michaelkro: Oh wait, but there is still a comma

devtools_-_localhost_8080_migration

michaelkro: I see what the difference is now.  In your screenshot, your migrations are still in the pending state.  The ones on my screen are in the in progress state

manageiq__migration

AllenBW: @michaelkro thanks for the db, able to seeee what you seee, n FIXED IT <3 🙇‍♀️

AparnaKarve:
Per https://github.com/ManageIQ/manageiq-v2v/issues/415#issuecomment-419390513, we need this for 5.9 hence marking it g/yes

PR #533, opened 2018-08-03
Use ui classic redux and router

Hyperkid123:
In order to fix https://github.com/ManageIQ/manageiq-ui-classic/pull/4361#issuecomment-409544861. It is required to merge v2v router and redux with ui-classic.

Because ui-classic cannot use BrowserRouter (mainly because it would take an insane amount of time), we were forced to use hashRouter. Additionally,  most of the screens, are controlled via toolbars, trees and other controls, that are not part of most of the components it was also required to use redux to handle some routing. This creates conflict between ui-classic routing and v2v routing.

Fortunately it can be fixed fairly easily.

### PR dependencies in ui-classic
- [x]  https://github.com/ManageIQ/manageiq-ui-classic/pull/4407
- [x] https://github.com/ManageIQ/manageiq-ui-classic/pull/4406

### Necessary changes
- [x] use one redux store for react
- [x] use same routing as in ui-classic
- this will not break current v2v routing. Only change is in the URL. It will start use hash routing, but the react-router-dom Links will be unaffected. Only the route must be adapted by removing the `/migration` prefix from them.
- [x] Enable manageIQ as a global variable
- [x] Use ui-classic mounting method
- this is conflicting with the mounter from v2v. I have also found out, that the mounter is specifying some default prop values. I personally think this is not a good decision and these values should be actually default props. Or if this was meant to be configurable, just move them to redux store, or wrap them into object and pass them as a input prop to root component and distribute them from there. If there is some other reason for this to exist, please let me know a will try to find some solution, that will keep this unchanged. 

~~remove the `app/javascript/components/index.js` for now i have moved these default props to corresponding components (mainly the `index.html.hml`, but it is not pretty and we should try to find some other place to put them)~~

Test are green and from what i can tell, it works and fixes the error from the comment above. But this is the first time i see the v2v component, so somebody elso should go through it and try to find out if i forget something.


Hyperkid123:
cc @michaelkro @priley86

Hyperkid123:
cc @skateman @himdel

app/assets/stylesheets/manageiq-v2v/base.scss
```diff
@@ -1 +1,5 @@
 @import '../../../javascript/react/app.scss';
+
+.react-wrapper {
+  height: 100%;
+}
```
skateman: Missing newline ;)

Hyperkid123: :fearful:

michaelkro:
It looks like it was always the intention to use `ManageIQ.redux.addReducer` and `ManageIQ.redux.store`

https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/17

AllenBW:
[Because its relevant, discussion about this also happened here.](https://github.com/ManageIQ/manageiq-ui-classic/pull/4398)

priley86:
hi @Hyperkid123 - apologies for late response (I am returning from extended PTO 😄 ). 

Thanks for revisiting this one (and thanks to @michaelkro and @AllenBW for keeping this moving!). I have no issues at all with introducing [connected-react-router](https://github.com/supasate/connected-react-router/) if that is consistent w/ redux routing in MIQ. It seems like you may be missing some of the benefits of that library though after taking a quick glance. I will post a couple of comments inline and see what you think...this looks like a good direction though!

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -189,7 +189,7 @@ class MigrationsCompletedList extends React.Component {
                         onClick={e => {
                           e.stopPropagation();
 
-                          redirectTo(`/migration/plan/${plan.id}`);
+                          redirectTo(`/plan/${plan.id}`);
```
priley86: do you not want to use the connected react router's [push method](https://github.com/supasate/connected-react-router/blob/master/FAQ.md#with-storedispatch) now? This should enable the history to be tracked in redux state.

Hyperkid123: Yep i know about this. My mindset about tthis is PR, is to make as little changes as possible (we use the push, goBack etc. methods in ui-classic) to minimize change of breaking something.

But if you will require this changes i will happily do it

app/javascript/react/index.js
```diff
@@ -1,19 +1,43 @@
...
-  <Provider store={store}>
-    <Router>
-      <React.Fragment>
-        <NotificationList />
-        <Routes store={store} />
-      </React.Fragment>
-    </Router>
-  </Provider>
-);
+class OwerView extends React.Component {
```
priley86: did you mean to call this "Overview" or something else? I guess since this is the root of the application (and not technically the Overview), maybe we can just call this class "Index" ? just a consideration...

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
priley86: @Hyperkid123 i think it would be a nice to test this feature if it's not too much additional (otherwise, yes, fine w/ waiting). 

I will try to pull this branch down and do some more testing on my side soon...

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
```diff
@@ -151,7 +151,11 @@ MappingWizardDatastoresStep.propTypes = {
 };
 MappingWizardDatastoresStep.defaultProps = {
   clusterMappings: [],
-  fetchStoragesUrls: {},
+  fetchStoragesUrls: {
```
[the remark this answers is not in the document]
Hyperkid123: I figured it was used for configuration of the plugin. So if there is scenario, when it might be required to change those values (like using it in different product) i would maybe recommend to move them to redux and have some "read only" reducer for configuration.  I think all components that are using this are connected anyway, so basically only the origin of the data would change. We would not have move the data to `defaultProps` and we could eventually delete the whole `coreComponents` file.

Only other thing that is depending on this are the routes i think, but it should not be hard to replace.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
Hyperkid123: OK i will do it, its very simple change anyway. The worst thing will be finding all the places where it should be used 😄 .

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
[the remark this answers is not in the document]
Hyperkid123: @priley86 ok i will move it to default props for now. You guys can decide which path you want to take later. Also ,removing the original component registry should not be part of this PR. That is maybe something for future refactoring.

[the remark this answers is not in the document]
priley86: Works!

Hyperkid123:
@priley86 i have encountered an error with the dropdown component. It turns out that after the update to webpack4 the react-bootstrap dropdown (which is used as a base for select if patternfly react), i got error similar to [this issue](https://github.com/JedWatson/react-select/issues/2025). This is cause by the react mounter in ui-classic.

I will use the original until we found where is the problem.

priley86:
this is looking much better! are you ready for another review @Hyperkid123 ?

Hyperkid123:
@priley86 yes

AllenBW:
@Hyperkid123 can yah fix the travis prettier failures? or if its a bother I can jump in and push a commit 🤔

Hyperkid123:
@AllenBW i can do it tomorrow

AllenBW:
Ok gonna do a proper review with morning eyes, but jumped in on this earlier, saw no errors **and** everything worked as expected 😍 😋

Hyperkid123:
@AllenBW OK i had to do some minor adjustments.

### eslint jsx-a11y/label-has-for
this one is actually deprecated and should not be used anymore (it is not very good if the target is in variable...). I have added weaker rule for this

### jest: testURL
This is a bit weird, sometimes it shows sometimes it doesn't, we had a similar issue here https://github.com/ManageIQ/react-ui-components/pull/67

Hyperkid123:
@AllenBW @priley86 i've got another question. I've noticed that you are using `IntlProvider`, but you are also using getText from miq? Is it intentional or can it be removed?

priley86:
@Hyperkid123 some background on react-intl can be found [here](https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/19#issuecomment-363836395)... 

Initially in the project we were not sure how to best add lanaguage/date/time internationalization support in React and decided to follow this Foreman [PR](https://github.com/theforeman/foreman/pull/5184). However, it looks like Foreman did not move forward with this. 

Since we are using gettext and happy w/ that for messages, and now using Moment JS for dates/time, it seems like we can safely remove react-intl.

@AllenBW @AparnaKarve - any thoughts on this?

AllenBW:
@priley86 I agree. 😏 🍶 😋 👍

AparnaKarve:
@priley86 Yes, OK to remove react-intl

priley86:
> @priley86 I agree. 😏 🍶 😋 👍

ok - opened #559 for this @AllenBW . 🌮 💃

Hyperkid123:
@priley86 ok, i was just asking because i did not add the intl context to the component.

Hyperkid123:
@priley86 @AllenBW is there something else that i should do?
btw travis is failing with node version 6
```
error typescript-eslint-parser@16.0.1: The engine "node" is incompatible with this module. Expected version ">=6.14.0".
error Found incompatible module
```
Did you add something recently?

AllenBW:
Nah nothing of that nature was recently added @Hyperkid123, looking into it... maybe we'll just 🗑 6.10 tests 🤔

himdel:
Oh yeah, sorry about that, looks like we don't support node 6 anymore... (Right now, this is just a `...` in our webpack config, so we can undo that if there's a problem .. but the build is using node 8 since last week.)

himdel:
can can :dancer: :D But yeah, can all the tests :)

Hyperkid123:
Np, happy to help

AparnaKarve:
Tested this and seems to work well.

It looks like with this change, we won't see the redux logging in the console anymore.

We do have the Redux Dev tools Chrome extension to monitor the redux activity, but I do like the console logging better. Can we enable that if possible?

Other than the above observation, everything else looks great!
Thanks @Hyperkid123 !

AllenBW:
Gonna merge this before some really 🤕 merge conflicts start popping up.. @AparnaKarve totally not bailing on the need to console log redux activity!  Can be done in another pr 🤗

martinpovolny:
Happy to see this landed! :fireworks: Thanks to all for the effort :beers: 

May I suggest the V2V team to ping @himdel, @skateman, @Hyperkid123 and @martinpovolny in case you are about to introduce some major concept in V2V. I mean something like routing, redux store, notifications or i18n. Something that has a bit potential of clashing with the ui-classic. Please?

It's absolutely not that I'd want to slow you down in your great job. I'd like the cooperation to be as smooth as possible.

michaelkro:
Yes, we definitely will @martinpovolny! 🍻

Issue #546, opened 2018-08-06
OSP Support: Add a new step to the Plans wizard when migrating to OpenStack


vconzola:
If the infrastructure mapping selected in step 1 of the Plans wizard has OpenStack as a target, then after VM selection add a step to the wizard for instance properties.
![9cccc3-createplan-vms-selection osp option4](https://user-images.githubusercontent.com/22823035/43740094-45434aa2-9998-11e8-9d80-43da64beafa8.png)

![screenshot 2018-08-14 11 17 13](https://user-images.githubusercontent.com/811963/44100713-a3d820e4-9fb3-11e8-9edd-2a0553486baf.png)

mturley:
This issue tracks the entire Instance Properties step, which has been worked on in the following PRs so far:

https://github.com/ManageIQ/manageiq-v2v/pull/547
https://github.com/ManageIQ/manageiq-v2v/pull/562
https://github.com/ManageIQ/manageiq-v2v/pull/627

we still need to alter the POST body when submitting the new plan to the API to include these selections.

michaelkro:
We can close this issue right?

mturley:
Yeah, thanks for following up. This one's been done for a while.

PR #547, opened 2018-08-06
[#546][1 of 2] OSP Support: Add (disabled) Instance Properties step in Plan Wizard

mturley:
This is the first part of the work for #546, but that issue will be closed by the next PR.

- [x] Add step to Plan Wizard conditionally on the special mapping name `fake-osp-mapping`.
- [x] Refactor Plan Wizard to use string identifiers instead of step indexes in logic
- [x] Replace `fake-osp-mapping` condition with detection of an actual Openstack target

Remaining tasks are being moved to a new PR.


mturley:
So far, this one just gets the selected mapping object into a variable in PlanWizardBody so it can be used to conditonally render steps.

mturley:
We have a conditionally rendering step now, but I need to convert all our logic based on step index to use unique step ids in order to fix all the off-by-one errors this causes.

vsbfuoxsiw

mturley:
Ok, @priley86 @michaelkro @AllenBW , this is now in a stable place which just conditionally adds an empty step to the wizard. I would love some help stress-testing it to make sure I didn't break anything existing in the Plan Wizard.

For now, you can trigger the new Instance Properties step to appear by creating and selecting a mapping with the exact name `fake-osp-mapping`.

AllenBW:
@mturley so is this ready for review? orrrrrrrrrrrrr 🤔 ❓

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -5,25 +5,88 @@ import { createMigrationPlans } from './helpers';
...
     const { activeStepIndex } = this.state;
 
-    if (activeStepIndex === 1) {
+    const activeStep = this.getActiveWizardStep();
+
+    if (activeStep.id === 'planWizardVMStep') {
       // reset all vm step values if going back from that step
       resetVmStepAction();
-    } else if (activeStepIndex === 2) {
+    } else if (activeStep.id === 'planWizardAdvancedOptionsStep') {
```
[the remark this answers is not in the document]
mturley: That's true, I could use shared constants. I'll do that, thanks @priley86

mturley:
Sorry for confusion @AllenBW , this is not ready for a final review, but the rest of the work will mostly be inside the new `PlanWizardInstancePropertiesStep`, so the plan wizard refactor parts are ready for review. Feel free to wait until I'm all done.

Unless you guys think I should do the rest in a separate PR?

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -5,25 +5,88 @@ import { createMigrationPlans } from './helpers';
...
+        disableGoto: !this.props.planWizardGeneralStep.values
+      },
+      {
+        id: 'planWizardVMStep',
+        title: __('VMs'),
+        render: () => this.planWizardVMStepContainer,
+        disableGoto: !this.props.planWizardVMStep.values
+      },
+      {
+        openstackSpecific: true,
```
michaelkro: I'm concerned about the scalability of this approach.  While there are no definite plans to add other target providers, I think it would be wise to build in some flexibility to do so.

Perhaps it would be better to insert a step, [instead of filtering](https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/547/files#diff-a7507d1fbaba12698eb13826effa926bR72)

mturley: I was originally thinking we could just iterate on that filtering behavior, but I think you're right that it'll get confusing with more target providers. I'll rework this to be a little more flexible.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -86,11 +172,11 @@ class PlanWizard extends React.Component {
         setMigrationsFilterAction(MIGRATIONS_FILTERS.notStarted);
       }
       this.setState({
-        activeStepIndex: Math.min(activeStepIndex + 1, planWizardSteps.length - 1)
+        activeStepIndex: Math.min(activeStepIndex + 1, wizardSteps.length - 1)
       });
     } else {
       this.setState({
-        activeStepIndex: Math.min(activeStepIndex + 1, planWizardSteps.length - 1)
+        activeStepIndex: Math.min(activeStepIndex + 1, wizardSteps.length - 1)
```
michaelkro: I'm sure I'm missing something, but I'm not sure I follow why we're still using the array index to track the active step if we're introducing `id`s for each step

mturley: That's only still at the end of the logic here in the `else` case, to proceed to the next step whatever it may be, if no specific step navigation was called out. It's just to avoid "and also if you're on A go to B and if you're on B go to C" in case we hit that case.  I missed that it was in the above case though.

With this in place, I didn't have to add any extra logic here when I added the empty new step, it'll still proceed through it.

mturley:
I pushed my unfinished work on the new Instance Properties table, it needs another day of work or so, this reactabular / pf-react table stuff is new to me and requires a staggering amount of boilerplate.  I'm going to turn this PR into just the new empty step but disabled, so we can avoid this being one large PR.

mturley:
@michaelkro @AllenBW @priley86 I decided to try and get this merged with the new step disabled, because it involves some refactoring and I didn't want to hold up a tricky rebase while I finish the table work. This one is now ready to review and merge, if you guys are ok with that.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardBody.js
```diff
@@ -69,10 +35,12 @@ PlanWizardBody.propTypes = {
   plansBody: PropTypes.object,
   planWizardGeneralStep: PropTypes.object,
   planWizardVMStep: PropTypes.object,
+  transformationMappings: PropTypes.array,
   hideAlertAction: PropTypes.func
 };
 
 PlanWizardBody.defaultProps = {
+  wizardSteps: [],
```
AllenBW: If on 29 yah say this is array of objects and default value as an empty array (with no objects) 🤔

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -5,25 +5,109 @@ import { createMigrationPlans } from './helpers';
...
+      planWizardGeneralStep && planWizardGeneralStep.values && planWizardGeneralStep.values.infrastructure_mapping;
+    const selectedMapping = transformationMappings.find(mapping => mapping.id === selectedMappingId);
+
+    // eslint-disable-next-line no-unused-vars
+    const openstackTargetSelected =
+      selectedMapping &&
+      selectedMapping.transformation_mapping_items &&
+      selectedMapping.transformation_mapping_items.some(
+        item =>
+          item.destination_type === 'CloudTenant' ||
```
AllenBW: Any chance these can be abstracted into the constants file?

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardBody.js
mturley: An array with zero objects is still an array of objects :) I think maybe this one just shouldn't have a default prop though... it would never be useful.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
mturley: Good point. I'm not sure this is the final logic for this, though. I'll move them up anyway, can always change them.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardBody.js
```diff
@@ -81,6 +48,7 @@ PlanWizardBody.defaultProps = {
   plansBody: {},
   planWizardGeneralStep: {},
   planWizardVMStep: {},
+  transformationMappings: null,
```
michaelkro: We should set this to an empty array to avoid `transformationMappings.<method> is not a function` type errors

mturley: Good catch, thanks.

mturley:
All feedback so far is addressed and squashed.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardBody.js
```diff
@@ -69,6 +35,7 @@ PlanWizardBody.propTypes = {
   plansBody: PropTypes.object,
   planWizardGeneralStep: PropTypes.object,
   planWizardVMStep: PropTypes.object,
```
michaelkro: The redux-form related props were lifted to `PlanWizard`.  Should we remove the prop types and default props?

mturley: Ah, true! Good call.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -133,17 +227,16 @@ class PlanWizard extends React.Component {
...
             activeStep={activeStep}
             goToStep={this.goToStep}
             disableNextStep={disableNextStep}
             plansBody={plansBody}
-            planWizardGeneralStep={planWizardGeneralStep}
-            planWizardVMStep={planWizardVMStep}
-            planWizardScheduleStep={planWizardScheduleStep}
             alertText={alertText}
             alertType={alertType}
+            transformationMappings={transformationMappings}
```
michaelkro: One last thing to remove

vconzola:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1632355

PR #530, opened 2018-08-01
[#508] OSP - Infra Mapping Wizard Step 4

AllenBW:
fixes #508 

#531 ought to be merged before this, so you can seeeeeeeee da AWESOME 🌮 💃 (but that aside, this is g2g)

### Rhevm
screen shot 2018-08-02 at 10 40 30 am

### Openstack
screen shot 2018-08-07 at 9 36 00 am

Issue #508, opened 2018-07-19
OSP Support: Changes to Infra Mapping Wizard - Step 4


vconzola:
Clearer instructional text

![6a-createmapping-networks_2](https://user-images.githubusercontent.com/22823035/42962312-508c0314-8b5f-11e8-8f56-06a030f006b8.png)

Network mapping for OSP, no real changes from RHV 
![6b-createmapping-networks_2](https://user-images.githubusercontent.com/22823035/42962318-557b9f56-8b5f-11e8-8125-95b04bba7a52.png)

AllenBW:
@vconzola So are we hiding the dual pane mapper until source cluster network is selected from the dropdown? Additionally,  target cluster, the label, is that dynamic (same thing as from step 2? )

vconzola:
@AllenBW 1. "are we hiding the dual pane mapper until source cluster network is selected" - I believe that's the way it works today. That's the design intent, but i can't verify it was implemented that way because the only environment I can access has only one source cluster so its networks (and datastores) are selected by default on page load. 2. Yes, dynamic label, just like step 2.

AllenBW:
@vconzola looking for confirmation that rhevm target label should remain `Target Networks`, yah say 'no real changes', but every other osp issue has changed the rhevm label in some way 🤔

vconzola:
@AllenBW Target Networks is what I was told when I met with Marek and Brett to discuss mapping details and what I put in the mockups I showed during the Sprint 10 demo. But that doesn't mean it worn't change once folks start playing with OpenStack.  :-)

Issue #507, opened 2018-07-19
OSP Support: Changes to Infra Mapping Wizard - Step 3


vconzola:
Clearer instructional text

![5a-createmapping-datastores_2](https://user-images.githubusercontent.com/22823035/42962269-3434de02-8b5f-11e8-9109-6a56850f3a11.png)

For storage mapping for OSP, map VMWare datastores to OSP volume types

![5b-createmapping-datastores_2](https://user-images.githubusercontent.com/22823035/42962281-3cd7b232-8b5f-11e8-83a3-87b1e8defc2f.png)

michaelkro:
@vconzola I have a question about the label for the Target Storages panel.  The mockup seems to suggest that for VMW -> OSP migrations the list items should be of the form, 

```
target_project.name \ target_project.volume_type (btw, I'm not sure if this attribute exists)
```

an example could be `ocp-registry \ <SOME_SORT_OF_VOLUME_TYPE>`

On the other hand, for VMW -> RHV migrations, the list items are

```
target_provider.name \ datastore.name
```

an example could be `RHV \ Export`

Is this discrepancy intentional?

vconzola:
When Brett and I met with Marek from the OpenStack team he said that VMW datastores should map to OSP volume types. didn't make sense to me either. I think we need to ask the CF team for a sample API that shows what data we'll be receiving when OSP is selected as the target. We can either wait until next week or if you don't mind firing off a note to Adam and Greg B. they can probably respond sooner. I could send it but probably wouldn't use the right terms and just confuse everything.

michaelkro:
@vconzola I'm drafting an email to Marek, Adam, and Greg  (I've cc'ed you)

michaelkro:
Closed by #531

Issue #506, opened 2018-07-19
OSP Support: Changes to Infra Mapping Wizard - Step 2


vconzola:
![4aa-createmapping-clusters_2](https://user-images.githubusercontent.com/22823035/42962246-24f7ddc2-8b5f-11e8-8645-8d59bc091da6.png)

vconzola:
@AllenBW by second part do you mean, "list boxes labels will change dynamically based on selected target provider"?  If so, then what I meant is that when target provider is RHV the label for the list box should be "Target Provider/Datacenter/Cluster", but if the target is OSP the label should be "Target Provider/Project". Should this be a separate issue since it involves making label text dynamic and isn't related to OSP support itself?

AllenBW:
AH! Understood, nah that fits in this scope of work, though one lil question.. mockup has using forward slashes, but current state is backslashes, is this part of the requested work, the switching of slash directions?

vconzola:
I always get those dang slashes wrong. I think back in the days of DOS they looked like this \ when designating a path, but on the web they use /. So I guess we should use /. Do you agree?

AllenBW:
no preference either way, less changes, already in place is / so if i was to force a preference.. thats where it would be 😏

vconzola:
Then let's keep it that way.

PR #553, opened 2018-08-08
Update createTransformationMappings helper

michaelkro:
Update regexes to account for OSP mapping types.

# Notes
1.  Once #530 is merged, this is needed in order to create OSP specific transformation mappings
2.  The `Infrastructure Mappings` section in Overview will need to be updated for OSP mappings.

Addendum to issues #505, #506, #507, #508 

# OSP Transformation Mapping
fullscreen_8_8_18__12_53_pm

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/MappingWizard/helpers.js
+++ b/app/javascript/react/screens/App/Overview/screens/MappingWizard/helpers.js
@@ -4,14 +4,15 @@ export const createTransformationMappings = (
   mappingWizardDatastoresStep,
   mappingWizardNetworksStep
 ) => {
-  const clustersUrlRegEx = /\/api\/clusters\/\d{1,}/;
+  const sourceComputeUrlRegEx = /\/api\/clusters\/\d{1,}/;
+  const targetComputeUrlRegEx = /\/api\/clusters|cloud_tenants\/\d{1,}/;
   const clusterTransformationMappings = mappingWizardClustersStep.values.clusterMappings.reduce(
     (clusterTransformationsArray, targetClusterWithSourceClusters) => {
-      const destination = targetClusterWithSourceClusters.href.match(clustersUrlRegEx)[0];
+      const destination = targetClusterWithSourceClusters.href.match(targetComputeUrlRegEx)[0];
       const transformations = targetClusterWithSourceClusters.nodes.reduce(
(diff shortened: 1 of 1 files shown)


app/javascript/react/screens/App/Overview/screens/MappingWizard/helpers.js
```diff
@@ -21,16 +22,17 @@ export const createTransformationMappings = (
     []
   );
 
-  const datastoresUrlRegEx = /\/api\/data_stores\/\d{1,}/;
+  const sourceStoragesUrlRegEx = /\/api\/data_stores\/\d{1,}/;
+  const targetStoragesUrlRegEx = /\/api\/data_stores|cloud_volumes\/\d{1,}/;
```
AllenBW: Ok real quick, this could be crazy, but what about using the constants from  `MappingWizardConstants.js` here?  (only for the `cloud_volumes`/`cloud_networks') part

michaelkro: Yes, that is a great suggestion.  It seems likely that a least the OSP storages urls/things will be changing, and the whole point of those constants is to not have to chase down every `cloud_volumes`

Good call

michaelkro:
Ok @AllenBW, sadly, the endpoint names and the attribute types didn't quite line up (e.g. api/data_stores vs. <CLUSTER>.storages), but I think using a generic regex that matches on everything following the protocol://domain should work

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardGeneralStep/MappingWizardGeneralStep.js
```diff
@@ -57,7 +57,7 @@ const MappingWizardGeneralStep = props => {
         controlWidth={9}
         inline_label
         onSelect={onSelect}
-        style={{ visibility: 'hidden' }}
+        // style={{ visibility: 'hidden' }}
```
AllenBW: NAO... 🤔 well not yet?

michaelkro: ah oops :P

michaelkro:
@AllenBW updated :)

PR #569, opened 2018-08-13
Update gettext catalogs for master

mzazrivec:
* deleted old catalogs (to reflect the project name change)
* added `zanata.xml` for master
* added new catalog & translations for master


locale/zanata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<config xmlns="http://zanata.org/namespace/config/">
+  <url>https://translate.zanata.org/</url>
+  <project>manageiq-v2v-ui-plugin</project>
```
AllenBW: Unsure if this was auto generated or not, but the project name should reflect the recent repo name change, managiq-v2v (if auto generated, maybe we need to rename more tools 🤔)

mzazrivec: Nah, this is just the project identifier as it is set in Zanata. I just updated zanata's settings and updated this PR to reflect what you're suggesting above.

PR #580, opened 2018-08-17
[3579] Long string overflow resolved by removing popover max-width property

### LINKED ISSUE #579: UI: Text should be wrapped in popover properly
**Description**
This is UI issue. Popover should adjust its width if text length is greater than its size.
Text should be wrapped in popover properly. 

In above image while displaying hostname, `.com` is displaying outside of popover so it should be wrapper inside.

**Version**
5.9.4.4.20180816162527_c00eb23 

**Impact**
None

screenshot from 2018-08-17 15-03-29

AllenBW:
fixes #579

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.scss
+++ b/app/javascript/react/screens/App/Plan/Plan.scss
@@ -30,3 +30,7 @@
 .task-info-popover .popover-title {
   text-transform: capitalize;
 }
+
+.no-max-width {
+  max-width: none;
+}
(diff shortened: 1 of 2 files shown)


AparnaKarve:
Fix looks good to me.

Is this issue more pronounced in Gaprindashvili?

Also, should this be `gaprindashvili/yes`?

AllenBW:
yeah this can be... @Yadnyawalkya can we get a bz for the issue?

Yadnyawalkya:
@AllenBW  BZ for you -> https://bugzilla.redhat.com/show_bug.cgi?id=1618813

PR #593, opened 2018-08-20
[#592] Change text to `Start Time` in the popover

### LINKED ISSUE #592: Elapsed time field in details tooltip shows migration start time instead
When the tooltip for an in progress or completed migration is displayed by selecting the info icon from the plan details listview, the first line of the tooltip is labeled "Elapsed Time:", however what is shown is the start time of the migration. The label should be changed to "Start Time:".
screen shot 2018-08-20 at 1 52 45 pm

AparnaKarve:
Fixes #592

screen shot 2018-08-20 at 2 21 26 pm


AparnaKarve:
Setting it to `gaprindashvili/no` since it's not a blocker.
We can change it to `gaprindashvili/yes` if there is an explicit request for this fix.

PR #596, opened 2018-08-21
Updating rails -v it to match cui, keep v2v safe (from Hakiri criticism)

AllenBW:


manageiq-v2v.gemspec
```diff
@@ -9,13 +9,13 @@ Gem::Specification.new do |s|
...
-  s.summary     = 'ManageIQ plugin for v2v trasformations'
-  s.description = 'ManageIQ plugin for v2v trasformations'
+  s.summary     = 'ManageIQ plugin for v2v transformations'
+  s.description = 'ManageIQ plugin for v2v transformations'
   s.license     = 'MIT'
 
   s.files = Dir['{app,config,db,lib}/**/*', 'Rakefile', 'README.md']
 
-  s.add_dependency 'rails', '~> 5.0.0', '>= 5.0.0.1'
+  s.add_dependency 'rails', '>= 5.0.0.1', '< 5.2'
```
Fryguy: Why do you need this at all?  You don't need the dependency, because this plugin will run as part of ManageIQ which already provides this dep.

manageiq-v2v.gemspec
```diff
@@ -9,13 +9,13 @@ Gem::Specification.new do |s|
...
 
   s.add_development_dependency 'sqlite3'
```
Fryguy: Presumably you also don't need this, so you should probably remove it.

Fryguy:
Aside, I'm not sure why this doesn't work, when it clearly works in ui-classic.  Try reversing it?

`s.add_dependency 'rails', '< 5.2', '>= 5.0.0.1'`

PR #599, opened 2018-08-21
[#577] Removes dead retry button from failed mapping creation

### LINKED ISSUE #577: Non operational "Retry" button in Mapping Wizard
# Steps to Reproduce
- Currently, there is a bug that allows the user to progress past the first step of the Mapping Wizard after entering a duplicate name (please see issue #576) 
- At the end of the wizard, attempting to create a mapping with a duplicate name results with the error state shown in the screenshot below
- There is no action bound to the `Retry` button

# Screens
fullscreen_8_15_18__8_58_am

AllenBW:
closes #577

### We did what?!
screen shot 2018-08-21 at 2 44 12 pm


AparnaKarve:
@AllenBW Per the discussion here https://github.com/ManageIQ/manageiq-v2v/issues/577#issuecomment-414755187, should we send the user back to Step1?
or at least enable the Back button?

AllenBW:
@AparnaKarve sorry, shoulda updated, talked with @vconzola offgithub and he said removing it would be sufficient...

PR #603, opened 2018-08-23
Add plugin display name

mzazrivec:
See https://github.com/ManageIQ/manageiq-ui-classic/pull/4496 for primary motivation.


AllenBW:
Gotta echo my sentiment [here](https://github.com/ManageIQ/manageiq-ui-classic/pull/4496#issuecomment-415372420) unsure how useful v2v as a name is (vs using repo name)

PR #605, opened 2018-08-24
Load mapping info for a plan only when available

AparnaKarve:
Infrastructure mapping is displayed as unavailable for a plan although it *is* available.

The reason can be attributed to the dependency of one redux state `transformationPlans` with the other `transformationMappings`

Introduced a new state `plansMutatedWithMappingInfo` which tracks if the plans were mutated with the transformation mapping info based on a simple indicator - `state.transformationMappings.length > 0`

While we are waiting for the Mapping info to be available, we can display the `Loading...` message as shown below -
screen shot 2018-08-24 at 12 38 49 pm

Only after the Mapping info is available, we display it -
screen shot 2018-08-24 at 12 46 52 pm

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -251,7 +251,8 @@ class Overview extends React.Component {
       toggleScheduleMigrationModal,
       scheduleMigrationModal,
       scheduleMigrationPlan,
-      scheduleMigration
+      scheduleMigration,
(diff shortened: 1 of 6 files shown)


AllenBW:
@AparnaKarve can yah update the `OverviewReducer.test.js`? Travis is not a fan 🤔

AparnaKarve:
@AllenBW For some reason, today I was fortunate enough to see this issue every time during a reload, hence the PR :)

Travis is now resolved.

PR #614, opened 2018-08-30
[#611] When no clusters are discovered (and loading is done) display message

### LINKED ISSUE #611: Show "No clusters found" in the infra mapping wizard if clusters have not been configured
In the source and target cluster list boxes in step 2 of the infra mapping wizard, alert the user if either the source or target provider is present but no clusters are found. Currently, we just show an empty list box which leads to user confusion.

AllenBW:
fixes #611

### here's the message we show (with this pr)
screen shot 2018-08-30 at 7 09 45 am


AllenBW:
Maybe, but the issue was very specific source clusters only, will defer to @vconzola for expanding the scope of this work 🤔

vconzola:
@AllenBW @michaelkro While Miguel's issue was related to source clusters not being found I tried to write the problem description so that any fix would apply if either source or target clusters are not found. If this PR only addresses source, should we create another one to address not found target clusters?

AllenBW:
fine @codeclimate, call me out on being lazy, i'll fix it and still love you

PR #616, opened 2018-08-30
[#615] Do not render Migrations if there are no plans

### LINKED ISSUE #615: Migrations section rendering when there are no migrations
Regression introduced by #597 

Currently, the Migrations section (in the empty state) is being rendered when there are no infrastructure mappings and no migration plans.  The expected behavior is that only the Infrastructure Mappings section is shown (in the empty state)

# Screens
## Current
manageiq__migration

## Expected
manageiq__migration

michaelkro:
Fixes #615 

Current check is not sufficient b/c empty JS arrays are truthy

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -311,7 +311,7 @@ class Overview extends React.Component {
           }
           style={{ marginTop: 200 }}
         >
-          {(transformationMappings.length > 0 || transformationPlans.length > 0 || archivedTransformationPlans) && (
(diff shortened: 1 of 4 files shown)


app/javascript/react/screens/App/Overview/Overview.js
```diff
@@ -311,7 +311,9 @@ class Overview extends React.Component {
           }
           style={{ marginTop: 200 }}
         >
-          {(transformationMappings.length > 0 || transformationPlans.length > 0 || archivedTransformationPlans) && (
+          {(transformationMappings.length > 0 ||
```
AllenBW: Weeeeeeeeell technically so a length of 0 will be false and anything greater than 0 will be true, the` > 0` feels like overkill 🤔

PR #618, opened 2018-08-30
Adjust UI based on the backend refactored logic for VM discovery/CSV process

AparnaKarve:
Few adjustments related to discovery of VMs and CSV process that are necessary after https://github.com/ManageIQ/manageiq/pull/17364 is merged

To be merged after https://github.com/ManageIQ/manageiq/pull/17364 and https://github.com/ManageIQ/manageiq-api/pull/463 are merged.

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
@@ -124,9 +124,13 @@ class PlanWizardVMStep extends React.Component {
     } else if (!isValidatingVms && validationServiceCalled) {
       // set make rows editable so they can be selected
       const validVms = Immutable.asMutable(valid_vms, { deep: true });
-      const inValidsVms = Immutable.asMutable(invalid_vms, { deep: true });
+      const inValidsVms = Immutable.asMutable(invalid_vms, { deep: true }).concat(
+        validVms.filter(vm => vm.valid === false)
+      );
       const conflictVms = Immutable.asMutable(conflict_vms, { deep: true });
-      const validVmsWithSelections = discoveryMode ? validVms : validVms.map(vm => ({ ...vm, selected: true }));
+      const validVmsWithSelections = discoveryMode
+        ? validVms
+        : validVms.filter(vm => vm.valid === true).map(vm => ({ ...vm, selected: true }));
       const combined = [...inValidsVms, ...conflictVms, ...validVmsWithSelections];
 
       if (combined.length) {
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/helpers.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/helpers.js
@@ -20,7 +20,14 @@ const manageBlankReason = vm => {
   }
 };
 
+const manageNoIdVMRows = (vm, vmIndex) => {
+  if (!vm.id) {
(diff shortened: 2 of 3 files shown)


gmcculloug:
@AparnaKarve Who can review this PR so it is ready to merge?

AparnaKarve:
@michaelkro @mturley Would you be able to test this alongside ManageIQ/manageiq#17364 and ManageIQ/manageiq-api#463?
Thanks.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/helpers.js
```diff
@@ -21,7 +21,6 @@ const manageBlankReason = vm => {
 };
 
 const manageOddCSVImportErrors = (vm, vmIndex, uniqueIds) => {
-  manageDuplicateVMRows(vm, vmIndex, uniqueIds);
```
michaelkro: If we remove this, duplicate invalid rows will show up in the table, is this desired?

manageiq__migration

AparnaKarve: The duplicate rows in general are never eliminated from the table.
`manageDuplicateVMRows` just adds the 'duplicate' status to the VM

In case of invalid rows, it really does not matter if we mark duplicate rows as duplicates or not, because they are invalid and unselectable anyway.

michaelkro: Whoops, I take back my 👍 😉 

Looks like if a VM is in another plan, and I import a CSV with duplicate rows of that VM, I get the following

fullscreen_9_13_18__2_04_pm

AparnaKarve: Ok, good catch :)

We will need `manageDuplicateVMRows` with a few adjustments

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
```diff
@@ -124,9 +124,13 @@ class PlanWizardVMStep extends React.Component {
     } else if (!isValidatingVms && validationServiceCalled) {
       // set make rows editable so they can be selected
       const validVms = Immutable.asMutable(valid_vms, { deep: true });
-      const inValidsVms = Immutable.asMutable(invalid_vms, { deep: true });
+      const inValidsVms = Immutable.asMutable(invalid_vms, { deep: true }).concat(
+        validVms.filter(vm => vm.valid === false)
```
mturley: If there are VMs returned in `valid_vms` with `vm.valid === false`, won't this cause them to appear as duplicates in `combined` below?

AparnaKarve: `valid` VMs and `invalid` VMs are two distinct sets that we get from the backend.

The UI vets the `valid` VMs for duplicates and marks the duplicate valid rows as invalid.
So in the above code I'm just ensuring that we append those duplicate valid rows to the invalid array, so that these VMs would be selectable.

So to answer you question, by doing the above, we won't see duplicates :)

mturley: Makes sense! Thanks.

gmcculloug:
@mturley @michaelkro Dependent PRs have been merged.

PR #619, opened 2018-08-30
[#610] Migration tasks in progress: Move description/playbook out of tooltip

### LINKED ISSUE #610: Move migration descriptions from automate to the details page without clicking the tooltip
This will require removing the description field from the tooltip and putting the text from that field in the status field on the details page.

screen shot 2018-08-29 at 9 54 03 am

Is there any implication for when playbooks are running? Is some logic required to decide what to display?

mturley:
After:
screenshot 2018-08-30 15 16 53

(moved the After image to the top here so it appears in the project board)

Before:
screenshot 2018-08-30 15 14 34

Closes #610

### DIFF
--- a/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
+++ b/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
@@ -553,7 +553,7 @@ class PlanRequestDetailList extends React.Component {
                 </Popover>
               );
 
-              const descriptionMessage =
+              const statusMessage =
                 task.options.prePlaybookRunning || task.options.postPlaybookRunning ? (
                   <div>
                     <b>{__('Running playbook service')}: </b>
@@ -561,7 +561,7 @@ class PlanRequestDetailList extends React.Component {
                   </div>
                 ) : (
                   <div>
-                    <b>{__('Description')}: </b>
+                    <b>{__('Status')}: </b>
                     {task.options.progress && V2V_MIGRATION_STATUS_MESSAGES[task.options.progress.current_description]}
                   </div>
                 );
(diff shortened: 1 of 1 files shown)


AparnaKarve:
New layout - LGTM!

Could we replace the word `Description` with something else?

mturley:
@vconzola do you have any input on this? I think maybe we could use my new layout here but maybe replace the word "Description" with "Status"?

All the strings that could appear in this field are listed here: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/PlanConstants.js#L18

vconzola:
@mturley I'm sorry. I've owed you a response on this for a week. I promise to look at it today and get back to you.

mturley:
No worries @vconzola. Just looking for something to do :)

vconzola:
@mturley Finally got a chance to look at this. I agree we should change "Description" to "Status". Also, I'm not sure that the total solution is as simple as I thought - just moving the Status field text in the tooltip to the plan details listviews. In looking at the list of [strings](https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/PlanConstants.js) I notice that the messages displayed when playbooks are running aren't included. I think @michaelkro added those, but I'm not sure where he's getting them. So unless you've already considered the playbooks case you'll need to add some logic to decide where to get the message text depending on whether a playbook is running.

mturley:
@vconzola I forgot to mention that I already considered that case; I just moved the same logic that was already in the tooltip down into the new location, playbooks included. When a playbook is running, it'll display "Running playbook service:" and the playbook name, instead of "Status: ..."

vconzola:
https://bugzilla.redhat.com/show_bug.cgi?id=1638527

Issue #607, opened 2018-08-27
Edit an existing Migration Plan


vconzola:
![screen shot 2018-08-29 at 3 39 24 pm](https://user-images.githubusercontent.com/22823035/44811118-c6e85780-aba1-11e8-80d9-b355a98ca3ab.png)
Notes:
- Only plans in Not Started state can be edited. (What about Complete - Failed state?)
- Edit action is located in a kebab menu.
- Selecting Edit opens the Migration Plan Wizard with saved values pre-filled for all fields on all steps.

mturley:
I'm starting to look at this next.

vconzola:
![](https://github.trello.services/images/mini-trello-icon.png) [Support for editing infrastructure mappings and migration plans](https://trello.com/c/xGHQDIny/198-support-for-editing-infrastructure-mappings-and-migration-plans)

PR #620, opened 2018-08-30
[WIP][#607] Edit an existing migration plan

mturley:
WIP PR for starting on #607 

- [x] Add kebab menu with edit option to plans not started list
- [ ] Figure out how to open the plan wizard in edit mode for a specific plan
- [ ] Open the plan wizard in edit mode
- [ ] When the plan wizard is in edit mode, pre-fill selections
- [ ] ???
- [ ] When the plan wizard is submitted in edit mode, use edit API endpoint instead of creating a new plan


mturley:
As soon as @AparnaKarve's PR https://github.com/ManageIQ/manageiq/pull/17976 is merged, I'll be able to fix the path column in the VM step, and then this PR will be ready for review.

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js
```diff
@@ -125,6 +126,18 @@ class PlanWizard extends React.Component {
         return;
       }
       hideAlertAction();
+
+      if (editingPlan && planWizardGeneralStep.values.vm_choice_radio === 'vms_via_csv') {
+        // If the user selected CSV and they are editing a plan, warn before proceeding.
```
michaelkro: Please remove comment

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardAdvancedOptionsStep/PlanWizardAdvancedOptionsStep.js
```diff
@@ -99,13 +105,6 @@ PlanWizardAdvancedOptionsStep.defaultProps = {
...
-  initialValues: {
-    playbookVms: {
-      preMigration: [],
-      postMigration: []
-    },
-    preMigrationPlaybook: '',
-    postMigrationPlaybook: ''
-  },
   destroyOnUnmount: false
+  // initialValues prop is passed in via ./index
```
michaelkro: I think this guy can be removed too

michaelkro:
Sort of an weird interaction between CSV import auto selecting all valid VMs and edit pre-fill.  Not sure if this is a non-issue though @vconzola?

# Steps to Reproduce
1.  Create a plan using VM discovery, select a few VMs to migrate
2.  Edit the plan, but this time use CSV import, making sure that at least one of the previously selected VMs is in your CSV
3.  On the VMs step, the previously selected VM, i.e., the VM that should be pre-selected for edit will not be selected.  This is b/c it is considered to already be part of a plan and is in the "warn but don't block state"

vconzola:
@michaelkro Not sure what you mean by in the "warn but don't block state" for VMs. Maybe you can show me? But this has me re-thinking about how we don't allow VMs to be in more than one plan. Is this really what we want? Also, what happens if we delete a plan before it's run - do the unmigrated VMs get released back into the wild?

michaelkro:
@vconzola here is a screen

fullscreen_9_13_18__9_08_am

I'm still kind of confused as to which VMs can and cannot be discovered, to be honest.  The results seem to vary for me depending on the status of the plan in which the VM is included, @AparnaKarve can you bring us into the light with regard to this?

I do believe that if a plan is deleted before running, its VMs become re-discoverable though

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardGeneralStep/PlanWizardGeneralStep.js
```diff
@@ -8,7 +8,8 @@ import { BootstrapSelect } from '../../../../../common/forms/BootstrapSelect';
 import { validation } from '../../../../../../../../common/constants';
 import { asyncValidate, onChange } from './helpers';
 
-const PlanWizardGeneralStep = ({ transformationMappings }) => (
+// TODO disable the mapping selection when editing
```
michaelkro: I think this `TODO` is `TODONE`

mturley: Oops! Thanks.

AparnaKarve:
> The results seem to vary for me depending on the status of the plan in which the VM is included, @AparnaKarve can you bring us into the light with regard to this?

@michaelkro Sure. I can get more info on this based on how the backend code is written and my own observations.

AparnaKarve:
@mturley Maybe we should not allow editing a plan that does not have an associated infrastructure mapping - because we get this error during the VM discovery step.

screen shot 2018-09-13 at 8 24 03 am

AparnaKarve:
The pre-selected VM in the VM Step shows up twice in Advanced Options

screen shot 2018-09-13 at 8 37 12 am


screen shot 2018-09-13 at 8 36 56 am

mturley:
@AparnaKarve I'm confused, how did you get a plan that has no associated mapping? I thought that was required when creating a plan?

mturley:
Thanks for reporting the other problem; it looks like this only happens when editing a second time, and only for the VMs that were added when editing the first time. I wasn't able to reproduce it with the first edit. I'll look into it

AparnaKarve:
>  I'm confused, how did you get a plan that has no associated mapping? I thought that was required when creating a plan?

@mturley I picked one of the aug21 plan from this list -

screen shot 2018-09-13 at 8 44 28 am

michaelkro:
Advanced Options Step Table selected count can get confused during edit

# Screens
fullscreen_9_13_18__11_42_am

# Steps to Reproduce
- Create a plan configured to run a pre or post playbook on some VMs
- Edit the plan, and on the VMs step, remove all the pre-selected VMs and choose some different ones
- Notice that the count at the bottom of the select column seems to be reflecting pre-loaded values from the currently being edited plan

AparnaKarve:
> Maybe we should not allow editing a plan that does not have an associated infrastructure mapping - because we get this error during the VM discovery step.

Or the other option for this case would be to enable the Infra Mapping dropdown, and allow the user to select one.

michaelkro:
VMs reverting to preselected values when moving backwards from Advanced Options Step

# Steps to Reproduce
- Open an existing plan in edit mode
- On the VMs step, edit the selected VMs and navigate forward to the Advanced Options Step
- Navigate back to the VMs step and notice that the table has reverted to the pre-selected values

AparnaKarve:
> The results seem to vary for me depending on the status of the plan in which the VM is included, @AparnaKarve can you bring us into the light with regard to this?

| Use case  | VM visible in discovery mode| VM visible in CSV import mode |
| ------------- | ------------- | ------------- |
| VM not included in any plan  | Yes | Yes |
| VM included in some plan  | No  | Yes, with a warning that VM exists in a different plan |
| Plan with VM is deleted  | Yes, in subsequent VM discoveries after plan deletion  | Yes  |
| Plan with VM fails migration  | Yes, in subsequent VM discoveries after the plan fails  | Yes  |
| Plan with VM completes migration successfully | No  | Yes, with a warning that VM already migrated  |

My test results do match the above table

mturley:
@michaelkro, I opened https://github.com/ManageIQ/manageiq-v2v/issues/639 to address https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-421058720, and I addressed https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-421055126 in https://github.com/ManageIQ/manageiq-v2v/pull/620/commits/f0f10bc102d7b4ee0cef52bfb909681c77a4808f

@AparnaKarve, I addressed https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-421052404 in https://github.com/ManageIQ/manageiq-v2v/pull/620/commits/81cb59a01b4b6a1397fd6d0fe8ec2b4d53462913, and https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-421050631 in https://github.com/ManageIQ/manageiq-v2v/pull/620/commits/616747466e3a390b66921a2717a32af53528f8f4.

I think the only remaining feedback I have not addressed here is https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-420993932, and I'm not sure whether we need to do anything about that here, @vconzola ?

mturley:
@michaelkro, I opened https://github.com/ManageIQ/manageiq-v2v/issues/640 to track the remaining issue so we can merge this this sprint.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardResultsStep/PlanWizardResultsStepActions.js
```diff
@@ -35,3 +39,21 @@ const _postMigrationPlansActionCreator = (url, migrationPlans, planSchedule) =>
...
+          resolve(response);
+          if (planSchedule === 'migration_plan_now') {
+            postMigrationRequestsAction(response, dispatch);
+          }
+        })
+        .catch(e => reject(e));
+    })
+  });
+
+export const putMigrationPlansAction = (url, migrationPlans, planSchedule) =>
```
michaelkro: There is something going on with this function signature.  Should there be four arguments with the second being `id`?

AparnaKarve:
Noticed that the wizard opens up when I click on the disabled Edit menu item.

AparnaKarve:
There seems to be a discrepancy in the VM discovery process with Create and Edit (and this is probably a backend issue)

The table above https://github.com/ManageIQ/manageiq-v2v/pull/620#issuecomment-421084070 shows the behavior during a Create Plan operation.
I would have expected the same behavior for Edit, but that does not seem to be the case.

Steps to reproduce
1. Create a plan with VM `A`
2. Edit the plan and add VM `B` and save the plan.  This plan now has VMs `A` and `B`
3. Create another plan. The discovery process in this Create operaion seems to be discovering `B`

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardAdvancedOptionsStep/index.js
```diff
@@ -2,15 +2,17 @@ import { connect } from 'react-redux';
 
 import PlanWizardAdvancedOptionsStep from './PlanWizardAdvancedOptionsStep';
 import * as PlanWizardAdvancedOptionsStepActions from './PlanWizardAdvancedOptionsStepActions';
-import { getVMStepSelectedVms } from './PlanWizardAdvancedOptionsStepSelectors';
+import { getVMStepSelectedVms, getVmIdsWithProperty } from './helpers';
 import reducer from './PlanWizardAdvancedOptionsStepReducer';
+import { findEditingPlan } from '../../PlanWizardSelectors';
```
michaelkro: Since we use this helper to add `editingPlan` to `mapStateToProps` in multiple containers, might it make sense to do this only once?

Maybe in `PlanWizard/index.js` or `Overview/index.js`.  Then we could just read the `editingPlan` from the redux store in the steps that need it

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardResultsStep/PlanWizardResultsStepActions.js
mturley: Wow, how did that even work... I guess it was dropping the actual `planSchedule` argument, thanks for that catch!

mturley:
> Create a plan with VM A
> Edit the plan and add VM B and save the plan. This plan now has VMs A and B
> Create another plan. The discovery process in this Create operaion seems to be discovering B

@AparnaKarve I noticed that too, and in fact it was discovering B again even when editing the plan, which is why I filtered out duplicates here: https://github.com/ManageIQ/manageiq-v2v/pull/620/files#diff-5c1096a11ccff6f726753cbd48bc02f8R130

I'm not sure what's going on there, I also suspect it's a backend issue.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardAdvancedOptionsStep/index.js
mturley: That's a good point. I'll do that.

mturley: (Note: after discussion with @michaelkro we decided to leave this as-is)

michaelkro:
Seeing a possible error attempting to actually run an edited plan

fullscreen_9_14_18__10_39_am

@mturley @AparnaKarve can you confirm whether or not you can migrate?

AparnaKarve:
Interesting test @michaelkro !

I think the Edit `service_template` API needs to be looked at...
It could very well be responsible for all the VM discovery issues we are seeing and now even the actual plan execution.

mturley:
@AparnaKarve @michaelkro I can confirm these issues with running an edited plan. I tried converting the PUT to a POST with action: edit, and I'm still getting the same issues. I'll push that commit anyway, since that seems like the pattern for editing things in this API.

AparnaKarve:
So this is indeed a backend bug.

When you save a Plan with Create or Edit it is supposed to create a `ServiceResource` record for the VM.
Currently, that is happening only for Create, but not for Edit.

AparnaKarve:
Backend PR to fix the VM discovery issues - https://github.com/ManageIQ/manageiq/pull/17989

PR #626, opened 2018-09-04
[#298] Delete a migration plan

### LINKED ISSUE #298: Users need to be able to delete migration plans
Users need to be able to delete migration plans

BEFORE STARTING WORK - need to find out the following:
- what restrictions are there around deleting a migration plan?  FYI - what states are delete available for?  Just Completed & Not Started ?
- need a design for this from @vconzola showing 1) where will delete action be available 
2) text for delete confirmation
3) what if you are trying to delete something that cannot be deleted, how do we information the user? 
4) what if the deletion is unsuccessful, how do we deal with that ?

API support already exists.  Ping Bill Wei with any questions.

mturley:
Closes #298.

- [x] Create kebab menu with delete option in Plans Not Started
- [x] Create component for delete menu item that opens a confirm dialog
- [x] Create an action for deleting the plan
- [x] Call action from the confirm dialog
- [x] Use delete menu item in Plans Completed - Success
- [x] Use delete menu item in Plans Completed - Failed
- [x] Use delete menu item in Plans Archived

Deleting a plan from Not Started state:

irehsra7jh

Deleting a plan from Completed state:

tu2usgsmku

Deleting a plan from Archived state:

31fzf6kgbn


mturley:
Fixed an issue with deleting from archived state, and added screen recordings to the PR description.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -297,9 +302,22 @@ class MigrationsCompletedList extends React.Component {
                                 >
                                   {__('Archive')}
                                 </MenuItem>
-                              </DropdownKebab>
-                            </div>
-                          )
+                              )}
+                              <DeleteMigrationMenuItem
```
michaelkro: Can we use the new `StopPropagationOnClick` component here as well?

mturley: Ah, good call.

michaelkro: Was there a particular reason you implemented `<StopPropagationOnClick>` differently here than [here](https://github.com/ManageIQ/manageiq-v2v/pull/626/files#diff-1475e5bbe3c2cf1ac448c09f7ecf928bR125)?

It might be nice to not have to introduce [this conditional](https://github.com/ManageIQ/manageiq-v2v/pull/626/files#diff-4802a1e9ccfaa210190b7623d319b9c5R14)

mturley: I only did that to not disrupt the existing structure too much. We could make them consistent, but i figured it might be useful to support both div and span there.

michaelkro: I think it makes sense to refactor for consistency's sake.  Unless it breaks something, lol

mturley: Sounds good. Refactored and pushed.

vconzola:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1638508

PR #627, opened 2018-09-05
Integrate best fit flavor API with OSP Instance Properties

AparnaKarve:
Implemented a non-redux form approach here since there the OSP Properties are pre-filled and there is no validation as such.

screen shot 2018-09-04 at 7 15 07 pm


michaelkro:
We need to disable the `Next` button in the wizard footer when a row is being edited

disable

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardActions.js
```diff
@@ -32,6 +32,20 @@ export const planWizardExitedAction = () => dispatch => {
...
+  });
+};
+
+export const planWizardBackAction = body => dispatch => {
+  dispatch({
+    type: PLAN_WIZARD_BACK,
+    payload: body
+  });
+};
+
```
michaelkro: I don't see reducer cases for these actions.  Perhaps these were missed during cleanup?

mturley: +1. What are we doing with these actions? Is the activeStepIndex from PlanWizard's react state being lifted up into Redux?

AparnaKarve: I need a way for the Instance Properties step to tell us if we would be navigating `Back` or `Next`.

- Navigating `Back` -- possibly means an update in the VM list.
If the VM list is updated, we need to calculate the best fit flavors again in the Instance Properties step

- Navigating `Next` -- This takes us to the Advanced Options Step. 
If we navigate back to the Instance Properties step, we should not be calculating the best fit flavors again because we might be overriding the flavor/security group values that the user may have edited previously.

So that's really the premise for these new actions in PlanWizardActions

I don't need the Reducer in PlanWizard, since I don't need to set any states there.
Instead I need to handle the  `PLAN_WIZARD_NEXT` case in `PlanWizardInstancePropertiesStepReducer`, where I am setting a flag that tells me if we need to recalculate the best fit flavors depending on where we are in the Wizard.
(recalculate only if in VM Step)

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
```diff
@@ -23,6 +43,54 @@ export default (state = initialState, action) => {
...
+        .set('instancePropertiesRows', instancePropertiesRowsUpdatedWithBestFlavor);
+    }
+    case SET_V2V_INSTANCE_PROPERTIES_ROWS:
+      return state.set('instancePropertiesRows', action.payload);
+    case SET_UPDATED_INSTANCE_PROPERTIES_ROW_ON_STANDBY:
+      return state.set('updatedInstancePropertiesRowOnStandby', action.payload);
+    case PLAN_WIZARD_EXITED:
+      return initialState;
+    case PLAN_WIZARD_NEXT:
+      return state.set('recalculateBestFitFlavorAndSecurityGroup', action.payload === stepIDs.vmStep);
```
mturley: This will fire when leaving the VM step even in the non-OSP case. That's probably fine, just thinking it through..

app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizardActions.js
michaelkro: ahh, ok, I'm with ya now.

Maybe, can we rename these actions though?  Currently, they read as if they should control whether or not the Wizard advances forward or backward.

Maybe something like `setPlanWizardForwardMetadata` (ugh) or `setPlanWizardBackAdditionalState` (uggghhh) or something

Naming is not fun

AparnaKarve: Sounds good... will figure something out.

Heads-up - I may actually end up using either the ugh or the uggghhh names... :)  -- we will see.

AparnaKarve: I went with `setMetadataWithNextButtonClickedAction` (and `setMetadataWithBackButtonClickedAction`) -- it's a mouthful, but hopefully there are no ambiguities now.

AparnaKarve:
@michaelkro For the Next button disabling, the other steps in the wizard rely on redux-form validation or check on specific redux-form values -- the VM Step, for example, checks for -
`selectedVms.length === 0`

In order to leverage this existing Next button disabling logic, I brought redux-form back in the picture.

ed22b9b contains the implementation related to `row in edit state` = `disabled Next button`.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
```diff
@@ -1,17 +1,37 @@
...
   isRejectedTenantsWithAttributes: false,
-  errorTenantsWithAttributes: null
+  errorTenantsWithAttributes: null,
+  bestFitFlavors: [],
+  isFetchingBestFitFlavor: false,
+  isRejectedBestFitFlavor: false,
+  errorBestFitFlavor: null,
+  instancePropertiesRows: [],
+  isSettingSecurityGroupsAndBestFitFlavors: true,
+  recalculateBestFitFlavorAndSecurityGroup: true
```
michaelkro: Please add default value for `updatedInstancePropertiesRowOnStandby`

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
```diff
@@ -36,23 +36,23 @@ class PlanWizardInstancePropertiesStepTable extends React.Component {
...
+    const { rows, updatedInstancePropertiesRowOnStandby, instancePropertiesRowsAction } = this.props;
     return {
       isEditing: ({ rowData }) => rowData.id === this.state.backup.id,
       onActivate: ({ rowData }) => {
         const index = findIndex(rows, { id: rowData.id });
         const backup = rows[index];
 
         this.setState({ backup, editing: true });
+        const { setUpdatedRowOnStandbyAction } = this.props;
+        setUpdatedRowOnStandbyAction(rowData);
```
michaelkro: I like that we're storing the backup row in redux here, b/c, that'll allow us to remove our dependency on `findIndex` :grin:

Can we please yank that from `package.json`, remove all instances of `findIndex()` from `inlineEditController`, remove `backup` from component state, and add `setUpdatedRowOnStandbyAction({})` to `onConfirm`, and `onCancel`

:tada:

mturley:
@AparnaKarve, I found a case we need to handle which will cause the "Loading OpenStack Properties" state to spin indefinitely.

If you try selecting a VM that has no best fit flavor (in my case I was selecting `CF45_Azure1`) and then clicking Next, since `all_fit` is an empty array and `best_fit` is null, `isSettingSecurityGroupsAndBestFitFlavors` remains true which keeps us in the loading state. We need to handle the case where there is "no fit".

The available VMs depend on the mapping, to create the mapping that exposed these VMs I just added all available datastores and LANs. a few other of the `CF45_` VMs have the same problem.

We should populate the dropdown with all available flavors, regardless of whether they fit.  But @vconzola, what should we select by default if no flavors fit the VM?

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/TableInlineEditRow/TableConfirmButtonsRow.js
```diff
@@ -73,6 +75,7 @@ class TableConfirmButtonsRow extends React.Component {
     const buttonsClass = `inline-edit-buttons ${this.props.buttonsClassName}`;
     return (
       <div style={divStyle} className={buttonsClass} key="confirmButtons">
+        <Field name="isRowEditable" component={CheckRowEditable} />
```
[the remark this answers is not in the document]
AparnaKarve: Sure, agree! Now that we know that we *do* need some kind of validation in the Instance Properties step, it makes sense to use `redux-form` for not only the validation part but also for our primary data storage.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
```diff
@@ -79,11 +104,16 @@ class PlanWizardInstancePropertiesStepTable extends React.Component {
...
           <FormControl
             componentClass="select"
-            defaultValue={value.name}
+            defaultValue={
+              additionalData.property === 'osp_security_group'
+                ? updatedInstancePropertiesRowOnStandby.osp_security_group.id
+                : updatedInstancePropertiesRowOnStandby.osp_flavor.id
+            }
+            onChange={e => this.handleSelectChange(e, additionalData)}
             onBlur={e => this.inlineEditController().onChange(e.target.value, additionalData)}
```
michaelkro: I think this can either be removed, or bound to the `onCancel()` action

AparnaKarve:
@michaelkro I have addressed all feedback in the last commit -- hopefully it's all readable.

AparnaKarve:
@mturley Thanks. 

Let's wait for @vconzola to weigh in here and we can decide on the next steps.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
```diff
@@ -11,6 +10,22 @@ import { paginate, Grid, PaginationRow, Table, PAGINATION_VIEW, Icon, Button, Fo
...
+      minimalInstancePropertiesRows.push({
+        vm_id: id,
+        osp_security_group_id: osp_security_group.id,
+        osp_flavor_id: osp_flavor.id
+      });
+    });
+
+    input.onChange({ rows: minimalInstancePropertiesRows, updatedInstancePropertiesRowOnStandby: {} });
+  }
+
```
michaelkro: Minor nit, but can we please move this method down past the `state` declaration?  Typically component lifecycle methods appear "in order", and while `state` is not exactly a lifecycle method, it used to be declared in `constructor`

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/index.js
```diff
@@ -41,8 +41,8 @@ const mapStateToProps = (
     //        integrating with back end
```
michaelkro: Can we please remove my `TODO` comment here :)

vconzola:
@mturley @AparnaKarve @michaelkro As I understand we should always receive a "best fit flavor" for every VM being migrated. if that's not the case we need to discuss with Greg B. UI should not have to provide any logic for this. We just supposed to provide the best fit and the option to override it. can you see why some of the VMs don't have a best fit?

michaelkro:
@vconzola when we request the "best fit flavor", the back end takes all the available flavors and compares their specifications to that of the VM to be migrated.

The flavors are not dynamically generated and must be configured by the user, so this issue arrises if all the available flavors are underprovisioned compared to the VM

vconzola:
@michaelkro they should still return a best fit, even if it's underprovisioned. do they tell us all flavors are undersprovisioned, or does the ui need to imply that by the fact that there's no best fit? the way this should work, imho, is that backend should provide a best fit (in this case it would be the largest flavor) and tell us "hey, here's the best we got, but just so you know, it's underprovisioned". then, following the marco manifesto we would flag it and tell the user it's underprovisioned but not stop them from migrating.

michaelkro:
@vconzola with the current implementation, if none of the available flavors can appropriately accommodate the VM, the response returns empty best fit and all fit attributes.  The front end could then use this info to infer that all flavors associated with the OSP project (aka cloud tenant) are underprovisioned.

I think it would be reasonable for us to ask for the best fit response to return some sort of default value in the case where none of the flavors fit.

@AparnaKarve, is this something that you could possibly create a PR for?

vconzola:
@michaelkro rather than a default, the backend should always return a best fit. that's why it's "best fit" and not "perfect fit". it should also always return the all fit list. we shouldn't have to complicate the UI for the case that all flavors are underprovisioned. backend logic/API should address that.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
```diff
@@ -23,6 +42,52 @@ export default (state = initialState, action) => {
...
+        const tenant = state.tenantsWithAttributes.find(
+          tenantsWithAttribute => tenantsWithAttribute.id === vmFlavor.tenant_id
+        );
+        const tenantFlavors = tenant.flavors;
+        const tenantSecurityGroups = tenant.security_groups;
+        const bestFitFlavorName = tenantFlavors.find(flavor => flavor.id === vmFlavor.flavor_id).name;
+        const defaultSecurityGroupName = 'default';
+        const defaultSecurityGroupId = tenantSecurityGroups.find(
+          securityGroup => securityGroup.name === defaultSecurityGroupName
+        ).id;
```
michaelkro: We should guard against a `TypeError` here.  `[].find()` returns `undefined` if there is no matching array element

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
```diff
@@ -36,40 +50,62 @@ class PlanWizardInstancePropertiesStepTable extends React.Component {
...
+        };
+
+        if (property === 'osp_security_group') {
+          updatedRowdata = {
+            ...input.value.updatedInstancePropertiesRowOnStandby,
+            osp_security_group: updatedInstanceProp
+          };
+        } else if (property === 'osp_flavor') {
+          updatedRowdata = { ...input.value.updatedInstancePropertiesRowOnStandby, osp_flavor: updatedInstanceProp };
+        }
```
michaelkro: Oops, sorry, I should have caught this when I made [comment](https://github.com/ManageIQ/manageiq-v2v/pull/627#discussion_r215941040)

This conditional block can be simplified to

```js
updatedRowdata = {
  ...input.value.updatedInstancePropertiesRowOnStandby,
  [property]: updatedInstanceProp
};
```

Also, up in line 75 can we please initialize `updatedRowdata` with an empty object :)

michaelkro:
@vconzola sorry, another case of poor word choice on my part.  I was agreeing with you that the back end should default to picking the "largest" flavor in the case when no flavors fit

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
AparnaKarve: Nice!
I had actually intended to make it generic, so thanks for the above suggestion!!

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/PlanWizardInstancePropertiesStepTable.js
michaelkro: Just wanted to say that I really like this approach, of waiting to set the `redux-form` data until the user is ready to leave the step.  It feels much cleaner and more robust than updating the store per user change/interaction.

We may want to revisit this in the future and see if it makes sense to refactor the other wizard steps to follow the same pattern

AparnaKarve:
@michaelkro Thank you for the invaluable feedback every step of the way!!

vconzola:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1632355

PR #631, opened 2018-09-06
[Fix #623] Add warning before overwriting CSV in plan wizard VM step

### LINKED ISSUE #623: Plan Wizard VM Step: There is no warning when overriding the uploaded CSV with another file
After uploading a CSV, if the user clicks this Import CSV button, there should be a warning to the effect of "Uploading a new CSV will overwrite any selections you have made here".

import-csv-screenshot

Here is the warning as originally mocked up:

screenshot 2018-08-31 13 17 26

mturley:
Fixes #623 

sjcqvmjnwt

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
@@ -3,11 +3,15 @@ import PropTypes from 'prop-types';
 import Immutable from 'seamless-immutable';
 import { Field, reduxForm } from 'redux-form';
 import { length } from 'redux-form-validators';
-import { Button } from 'patternfly-react';
+import { Button, Icon } from 'patternfly-react';
 import PlanWizardVMStepTable from './components/PlanWizardVMStepTable';
 import CSVDropzoneField from './components/CSVDropzoneField';
+import ConfirmModal from '../../../../../common/ConfirmModal';
 
 class PlanWizardVMStep extends React.Component {
+  state = {
+    overwriteCsvModalVisible: false
+  };
   componentDidMount() {
     const { vm_choice_radio } = this.props;
     if (vm_choice_radio === 'vms_via_discovery') {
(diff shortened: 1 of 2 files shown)


app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
```diff
@@ -49,6 +59,26 @@ class PlanWizardVMStep extends React.Component {
     } = this.props;
     const discoveryMode = vm_choice_radio === 'vms_via_discovery';
 
+    const overwriteCsvConfirmModal = (
```
michaelkro: Can we please move this out of the `render` method and make it an instance method?  This variable is being set every time this component re-renders

You can see this by placing a break point on line 63 and clicking on the select all checkbox

michaelkro: Actually, nvm.  It actually gets called more often as an instance method.  My bad.  As you were...

mturley:
Fixed the backdrop per @michaelkro's feedback, and updated the screen recording in the description.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
```diff
@@ -49,6 +59,28 @@ class PlanWizardVMStep extends React.Component {
...
+      <ConfirmModal
```
michaelkro: There is already an instance of `<ConfirmModal />` and associated actions to use on `Overview` screens.  For an example please see

- [actions import](https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Overview/screens/PlanWizard/index.js)

- [confirmation dialog invocation](https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Overview/screens/PlanWizard/PlanWizard.js#L137-L150)

It would be nice to reuse this modal to maintain consistency and reduce duplicate code

PR #660, opened 2018-09-24
[#606] Edit Mapping

### LINKED ISSUE #606: Edit an existing Infrastructure Mapping
Can't have associated migration plans

### LINKED ISSUE #647: Change the title of the infrastructure mapping wizard
The title of the infrastructure mapping wizard should match the action being performed in the wizard, so it should either be "Create Infrastructure Mapping" or "Edit Infrastructure Mapping".

### LINKED ISSUE #659: [MAPPING WIZARD] Associated mapped storages and networks do not clear when removing a cluster mapping
# Steps to Reproduce
1.  Start mapping creation, add a cluster mapping and a storages mapping
2.  Navigate back to the clusters step and remove the existing mapping
3.  Create a new (and different) cluster mapping and navigate forward to the storages step
4.  Notice that the previous (and now invalid) storages mapping still exists

michaelkro:
Fixes #606 
Fixes #659 
Fixes #647

Adds the ability to edit Infrastructure Mappings

# Edit Infrastructure Mappings
1. Only mappings that do not have associated plans (in any state) can be edited.  We will refresh mappings after plan creation/deletion so that the UI properly disables/enables mapping edit buttons
2. Wizard steps are pre-filled
    - Because we do not fetch source datastores/networks up front, it is necessary to fire off extra API calls for those steps
    - When ask the API for a target cluster with its storages/networks, we get something that looks like
      ```
      |- target cluster
      |--- [...storages/networks]
      ```
      So in order to reconstruct the saved storages/networks mappings, we select the mapped target storages/networks and mapped source storages/networks, and create objects that look like the following
        ```
        |- target cluster
        |--- target storage/network
        |----- source storage/network #1
        |----- source storage/network #2
        |----- ...
        ```
    - ~Pre-filling is based on the returned `transformation_mapping`, that is, the data used to set up the individual steps is static.  While it is ok if a user appends source clusters to the pre-filled mappings, the pre-filled state of the following steps cannot dynamically adjust to removal of cluster mappings.  So, if a cluster mapping is removed, exit edit mode~

      ***EDIT:*** This has been resolved by adding functionality that auto updates storages/networks mappings when cluster mappings are removed (described in the next section below)
    - `initialized` is a prop supplied by `redux-form` that indicates whether a form has been initialized with data
    - `initialize` is an action supplied by `redux-form` that will initialize a form field with the supplied data
3. ~If the target provider is changed, the wizard exits edit mode, and the remaining steps will not be pre-filled~
    - ~In edit mode, the first time the user clicks on the target provider dropdown on the Mapping Wizard General Step, a dialog will appear to warn the user~
    - ***EDIT:*** This has been removed in favor of disabling the dropdown in edit mode.  If the target provider is changed, we are essentially building an entirely new mapping, so it would make more sense just to ask the user to do that
4. ~Can test alongside https://github.com/ManageIQ/manageiq-api/pull/467~

    ***EDIT:*** This has been merged

## Auto update storages/networks when cluster mappings are removed
- Please see #659 for issue details
- Implemented for both editing mode and normal mapping creation using [this helper method](https://github.com/ManageIQ/manageiq-v2v/pull/660/commits/891717bd3f716aaaafc280cbec32bd120be8e70a#diff-0ff112c3341ea43b6c0965216c506131R52)
- The clusters step mappings and storages/networks step mappings roughly have the following structure
    ```
    // cluster step mapping
    |- target cluster/tenant
    |--- source cluster/tenant #1
    |--- source cluster/tenant #2
    |--- ...

    // storage/network step mapping
    |- target cluster/tenant
    |--- target storage/network
    |----- source storage/network #1 (from source cluster/tenant #1)
    |----- source storage/network #2 (from source cluster/tenant #2)
    |----- ...
    ```
- When we enter the storages/networks step of the wizard, we compare the mappings with the current cluster step mappings and remove any storage/network mappings for which there is no associated clusters step mapping
- ***NOTE:*** This has been implemented to handle mappings with multiple source clusters, but due to our test dbs, we can only verify for the single source cluster case


AparnaKarve:
@michaelkro Do you think we should prohibit editing the Target Provider for the mapping?

michaelkro:
> Do you think we should prohibit editing the Target Provider for the mapping?

@vconzola The current implementation allows changing the target provider by "exiting edit mode" if it is changed.  The first time the user clicks on the target provider dropdown, a confirm dialog pops open to warn them that this will cause all compute, storages, and networks mappings to be removed, and does not show on subsequent click to actually allow the change (only in edit mode)

fullscreen_9_25_18__8_41_am

I can remove all this and disable the dropdown in edit mode, if that is preferred.

vconzola:
@michaelkro To me, this sounds analogous to changing the Mapping when editing a plan - if a user does this it changes everything that comes after it. And I think in the Edit Plan Wizard we disabled the Mapping field. Is that correct @mturley? If so, then we should also disable the target provider in the Edit Mapping Wizard to be consistent.

mturley:
That's correct @vconzola, in the Edit Plan Wizard we disabled the Mapping field.

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStepReducer.js
```diff
@@ -72,6 +73,8 @@ export default (state = initialState, action) => {
         .set('errorHostsQuery', action.payload)
         .set('isFetchingHostsQuery', false)
         .set('isRejectedHostsQuery', true);
+    case MAPPING_WIZARD_EXITED:
+      return Immutable.merge(state, initialState);
```
mturley: We should probably just `return initialState;` here.

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStepReducer.js
```diff
@@ -40,6 +41,10 @@ export default (state = initialState, action) => {
         .set('errorTargetDatastores', action.payload)
         .set('isRejectedTargetDatastores', true)
         .set('isFetchingTargetDatastores', false);
+
+    case MAPPING_WIZARD_EXITED:
+      return Immutable.merge(state, initialState);
```
mturley: Same here.

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardGeneralStep/MappingWizardGeneralStepReducer.js
```diff
@@ -13,7 +13,7 @@ export default (state = initialState, action) => {
       return state.set('editingMapping', action.payload);
 
     case MAPPING_WIZARD_EXITED:
-      return state.set('editingMapping', null);
+      return Immutable.merge(state, initialState);
```
mturley: `return initialState`

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardNetworksStep/MappingWizardNetworksStepReducer.js
```diff
@@ -40,6 +41,9 @@ export default (state = initialState, action) => {
         .set('errorTargetNetworks', action.payload)
         .set('isRejectedTargetNetworks', true)
         .set('isFetchingTargetNetworks', false);
+    case MAPPING_WIZARD_EXITED:
+      return Immutable.merge(state, initialState);
```
mturley: `return initialState`

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
```diff
@@ -190,7 +191,13 @@ class MappingWizard extends React.Component {
             onClick={onFinalStep ? () => hideMappingWizardAction(onFinalStep) : this.nextStep}
             disabled={disableNextStep}
           >
-            {onFinalStep ? __('Close') : activeStepIndex === 3 ? __('Create') : __('Next')}
+            {onFinalStep
+              ? __('Close')
+              : activeStepIndex === 3
+                ? editingMapping
+                  ? __('Save')
```
AparnaKarve: Not seeing the `Save` text on the button while editing the mapping -- still seeing `Create`

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/MappingWizardResultsStep.js
```diff
@@ -4,42 +4,64 @@ import { noop, Button, Spinner } from 'patternfly-react';
...
+      postMappingsUrl,
+      postTransformMappingsAction,
+      transformationsBody,
+      editingMapping,
+      updateTransformationMappingAction
+    } = this.props;
 
-    postTransformMappingsAction(postMappingsUrl, transformationsBody);
+    if (editingMapping) {
+      updateTransformationMappingAction(postMappingsUrl, editingMapping.id, transformationsBody);
```
AparnaKarve: If this `transformationsBody` is exactly identical to it's initial state, maybe we should not make a backend edit call. What do you think?
Sometimes users may use the wizard to just browse through things and not necessarily make a change.

@mturley This is also applicable to your Edit Plan PR.

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardActions.js
```diff
@@ -22,10 +22,14 @@ export const mappingWizardExitedAction = () => dispatch => {
   });
 
   // Dispatch reset for all the wizard step forms here
-  dispatch(reset('mappingWizardGeneralStep'));
-  dispatch(reset('mappingWizardClustersStep'));
-  dispatch(reset('mappingWizardDatastoresStep'));
-  dispatch(reset('mappingWizardNetworksStep'));
+  dispatch(
+    destroy(
```
AparnaKarve: is `destroy` better than `reset`?

If so, should we also make this change in the Plan Wizard?

michaelkro:
https://github.com/ManageIQ/manageiq-v2v/pull/660#issuecomment-424143195 handled by 2df6d76

***NOTE:***  I removed the old commit titled `Exit editing mode if target provider is changed` using `git rebase --onto`.  The old SHA was 743c158eccca98190038ae1fdadb40fe5ea78380

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardActions.js
michaelkro: I changed to `destroy` because `reset` reverts the form state to its `initialValues`.  Now that we're using the initial values to pre-set the forms when editing, `reset` doesn't behave as expected in edit mode (perhaps we are abusing redux-form's initial value functionality here)

`destroy` completely wipes out the form's state, so the downside is that it adds one more step to the process.  Redux-form needs to reinitialize the forms when re-entering the steps

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizard.js
michaelkro: This is fixed by 2ee24e4

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/MappingWizardResultsStep.js
michaelkro: It seems the `transformation_mapping_items` arrays are quite different between `transformationBody` and `editingMapping`

Example of transformation item from `editingMapping`
```
{
  created_at: "2018-09-26T00:44:22Z"
  destination_id: "1"
  destination_type: "EmsCluster"
  id: "6"
  options: {merge: ƒ, replace: ƒ, without: ƒ, asMutable: ƒ, set: ƒ, …}
  source_id: "3"
  source_type: "EmsCluster"
  transformation_mapping_id: "1"
  updated_at: "2018-09-26T00:44:22Z"
}
```

Example of transformation item from `transformationBody`
```
{
  destination: "/api/clusters/1"
  source: "/api/clusters/3"
}
```

It will require a little more code, but I can make this happen.  What would you guys like for me to do?  @AparnaKarve @mturley

AparnaKarve: I would say let's not make the API edit call, if we don't have to.

This is in fact a use case that is consistently handled in ui-classic in other screens -- you will notice that the 'Save' button does not light up unless the form is truly dirty (i.e. different before and after values)

app/javascript/react/screens/App/Overview/screens/MappingWizard/MappingWizardActions.js
AparnaKarve: I see, Ok.

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/MappingWizardResultsStep.js
michaelkro: Resolved by b39135e4c033c92148298ee11b71b3df855bf0eb

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/helpers.js
```diff
@@ -0,0 +1,20 @@
+import { SOURCE_HREF_SLUGS, DESTINATION_HREF_SLUGS } from './MappingWizardResultsStepConstants';
+
+const normalizeTransformationItem = (transformationItem, targetProvider) => ({
+  source: SOURCE_HREF_SLUGS[transformationItem.source_type] + transformationItem.source_id,
+  destination:
+    DESTINATION_HREF_SLUGS[targetProvider][transformationItem.destination_type] + transformationItem.destination_id
```
AparnaKarve: Can we use interpolation here like we normally do?

```js
`${DESTINATION_HREF_SLUGS[targetProvider][transformationItem.destination_type]}${transformationItem.destination_id}`
```

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+    transformation.transformation_mapping_items.map(item => normalizeTransformationItem(item, targetProvider))
+  );
+  const currentTransformationItems = JSON.stringify(postBody.transformation_mapping_items);
+
+  return (
+    transformation.name !== postBody.name ||
+    transformation.description !== postBody.description ||
+    originalTransformationItems !== currentTransformationItems
+  );
+};
```
AparnaKarve: Thanks @michaelkro -- I really like this implementation!!

app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardResultsStep/helpers.js
michaelkro: Hahaha, yes :) I did it that way first, but it felt funny b/c I wasn't adding any strings between or around the variables.  I'll update

AparnaKarve: For some reason, everyone seems to prefer interpolation over string math :)

I don't have a preference as such, but would just like to see things uniform.

vconzola:
https://bugzilla.redhat.com/show_bug.cgi?id=1638853

PR #661, opened 2018-09-24
Request completion notifications would now be backend-initiated hence remove the frontend request completion notifications

### LINKED ISSUE #265: Migration completed notifications should also appear on detail screen
* to fix this, we can bubble up the `_getTransformationPlansActionCreator` redux call a level in the hierarchy (so that polling occurs for that request on both Overview and Overview Detail).

I'm fine w/ deferring this for now...

### LINKED ISSUE #578: Notifications created by v2v aren't displayed on pages other than v2v
I noticed on the [last sprint review](https://www.youtube.com/watch?time_continue=445&v=dQueeyktr6U) that V2V has notifications implemented. As I implemented the notifications in the core ManageIQ, I was curious how it is done here and found two major issues.

1. The notifications are displayed only when the user is on the V2V page
I assume that most of the tasks are running longer than the time spent on the page by the user, i.e. the user might visit other pages or simply just close the browser and come back later. In both cases the user will not receive any information about the migrations. Also when on the V2V page as the information displayed in the toast notifications is also reflected on the page, this makes the notifications less valuable and more redundant.

2. The notifications are missing after doing a full page refresh (F5)
This is closely related to the first point, the notification data is not backed by the server and so there is no history of older notifications that happened before visiting/refreshing the page.

We already have a solution for these problems in the core/classic UI using websockets and the notifications endpoint provided by the REST API.

ping @AparnaKarve @hyperkid123 @martinpovolny

AparnaKarve:
Removing frontend Request completion notifications since they would now be emitted by the backend.

The advantage that we now get is that the Migration Plan completion notifications would be visible from any part of the UI.

To be merged after - https://github.com/ManageIQ/manageiq/pull/18012 has been merged.

Fixes #265 
Fixes #578 

screen shot 2018-09-21 at 12 53 09 pm

https://bugzilla.redhat.com/show_bug.cgi?id=1642573

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -7,7 +7,6 @@ import * as AggregateCards from './components/AggregateCards';
 import Migrations from './components/Migrations/Migrations';
 import ShowWizardEmptyState from '../common/ShowWizardEmptyState/ShowWizardEmptyState';
 import componentRegistry from '../../../../components/componentRegistry';
-import getMostRecentRequest from '../common/getMostRecentRequest';
 import ConfirmModal from '../common/ConfirmModal';
 import EditPlanNameModal from './components/EditPlanNameModal';
 import {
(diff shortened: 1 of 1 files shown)


AparnaKarve:
Thanks @michaelkro - That was a good test!

AparnaKarve:
Hey @michaelkro - the backend PR was merged today...

Rebased my branch and it looks like we don't need `componentDidUpdate(prevProps)`

michaelkro:
@AparnaKarve that's awesome!  Especially b/c the `componentWillReceiveProps` lifecycle method will be deprecated in the future.  This will make migrating that code much easier.  🎊

mturley:
I'm seeing the same thing as @michaelkro. The "Request ____ has been approved" notification does appear on an unrelated plan's page, but the failed notification does not.

AparnaKarve:
Very strange.

I checked again, and sure enough could see the Plan completion notification - not sure why our local setups are so different.

I will try and verify this on an appliance.

simaishi:
Hammer backport details:

```
$ git log -1
commit ea1fc6c9df72f8e75ca4ea6eae5bd9dd3d4bd73e
Date:   Thu Oct 25 16:50:30 2018 -0400

    Merge pull request #661 from AparnaKarve/remove_request_completion_notification
    
    Request completion notifications would now be backend-initiated hence remove the frontend request completion notifications
    
    (cherry picked from commit b3581756330f11840704652b4a0d67a707ac8e4d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1642573
```

PR #664, opened 2018-09-24
[#639] Fix VM step reverting to pre-selected values on back button

### LINKED ISSUE #639: Plan Wizard: When moving back to VM step, table reverts to pre-selected values
When the VM step mounts, it triggers a query for discovery of new VMs, and if editing a plan, a query for the details of those VMs being pre-filled. This causes the table to reset to its initial state whenever the user enters the VM step.  Moving backwards from the Advanced Options step should instead retain whatever VMs the user had previously selected (so these queries probably need to happen only when the Next button is being pressed on the prior step).

mturley:
Fixes #639.

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
@@ -11,13 +11,15 @@ import { overwriteCsvConfirmModalProps } from '../../PlanWizardConstants';
 
 class PlanWizardVMStep extends React.Component {
   componentDidMount() {
-    const { vm_choice_radio, editingPlan, queryPreselectedVmsAction } = this.props;
-    if (vm_choice_radio === 'vms_via_discovery') {
-      this.validateVms();
-    }
-    if (editingPlan) {
-      const vmIds = getVmIds(editingPlan);
-      queryPreselectedVmsAction(vmIds);
+    const { vm_choice_radio, editingPlan, queryPreselectedVmsAction, pristine } = this.props;
+    if (pristine) {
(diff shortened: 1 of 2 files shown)


michaelkro:
Not related to the changes introduced here, but related to the issue at hand

It seems that if the user navigates all the way back to the General step, moving forward to the VMs step resets the VMs table to initial values.  This could be particularly frustrating in the OSP case, if say

1.  The user creates/edits a plan and selects/adds many many VMs
2.  The user moves forward to the Instance Properties step and customizes the security group/flavor selection for several rows
3.  The user wants to change the plan name, so navigates back to the general step
4.  The user moves forward to the VMs step only to see that all of his selections have been reset, and also in the instance properties step.

So, my question, is this something that can be handled as a separate issue, or must both the issue at hand and this new observation be handled together?

michaelkro:
@AparnaKarve my main concern is that we rely on detecting whether the wizard is moving forward to achieve the desired behavior with these changes, and we'd basically have to undo those changes when trying to handle the aforementioned issue

mturley:
I wonder if this could instead be fixed by checking if the redux-form values are dirty instead of whether the wizard is moving forward. I'll look into that. But I agree @michaelkro that we would probably have to undo these changes to fix the issue you pointed out. I think it's worth investigating a fix for both.

simaishi:
Hammer backport details:

```
$ git log -1
commit 8e727c73a5be87cc8444edb206a9a5631bb4ed1f
Author: Aparna Karve <aparna.karve@gmail.com>
Date:   Fri Oct 5 14:11:54 2018 -0700

    Merge pull request #664 from mturley/639-vm-step-reverting
    
    [#639] Fix VM step reverting to pre-selected values on back button
    
    (cherry picked from commit 2162f1a83937f761f2abf7a21e64ef3ecbe3845f)
```

PR #668, opened 2018-09-25
Make menu item keys more unique

### LINKED ISSUE #667: [MAPPING WIZARD] Keys for bootstrap-select menu items are not unique enough
We recently acquired a new test db with two source clusters, both named `Cluster`.  When both of these are mapped, the UI crashes on the storages step due to duplicate keys

michaelkro:
Fixes #667

# Steps to Reproduce
1.  Ping me for a new test db that has two source clusters
2.  Create a new infra mapping where you map both source clusters to the available target cluster
3.  UI will crash when entering the storages step


simaishi:
Removing `hammer/yes` as this was merged before branching

PR #686, opened 2018-10-04
Add new Migration status string to dictionary 

AparnaKarve:
Added `'Pre Transform VM'` to dictionary

[Reference](https://github.com/ManageIQ/manageiq-content/blame/0e388f9b20c6aa5c6951e7ce4df64316b1457d24/content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/transformation.yaml#L35)


app/javascript/react/screens/App/Plan/PlanConstants.js
```diff
@@ -29,6 +29,7 @@ export const STATUS_MESSAGE_KEYS = {
   POWER_OFF: 'Power off',
   POWER_ON: 'Power-on VM',
   PRE_MIGRATION: 'Pre-migration',
+  PRE_MIGRATE_VM: 'Pre Transform VM',
```
michaelkro: Do we want the value to be `Pre Migrate VM`, or do we want our constants to be `PRE_TRANSFORM_VM`?

AparnaKarve: down below we have a `MIGRATE_VM: 'Transform VM',` -- so just wanted to be consistent with that.
But I'd be OK to change both these entries to `TRANSFORM`
LMK if I should.

michaelkro: Ah yes, I see that now.  Let's change to `TRANSFORM`, if it's not too much trouble (not sure how many places we'd need to change that).  It's confusing to me because I can't figure out which part of the dictionary entry is "correct".  Plus it's consistent with the rest of the key/values :)

AparnaKarve: Ok, not a problem.

I have changed all MIGRATE/MIGRATION entries to TRANSFORM/TRANSFORMATION in `STATUS_MESSAGE_KEYS`

simaishi:
Hammer backport details:

```
$ git log -1
commit 9d7eb19915a71d028c2f09d0cb79c2b7fe6c3fb0
Date:   Fri Oct 5 10:25:46 2018 -0400

    Merge pull request #686 from AparnaKarve/add_new_statusmsg_to_dictionary
    
    Add new Migration status string to dictionary
    
    (cherry picked from commit 5cafdd3f60028c7f9ad04f71f4b73610d0c5a451)
```

PR #698, opened 2018-10-09
Use new ManageIQ.redux.addReducer.

Hyperkid123:
Update addReducer call to match the new interface. Details are described here https://github.com/ManageIQ/manageiq-ui-classic/pull/4744

This should be merged as soon as possible, after the https://github.com/ManageIQ/manageiq-ui-classic/pull/4744 gets merged, otherwise adding reducers will fail.

### DIFF
--- a/app/javascript/react/index.js
+++ b/app/javascript/react/index.js
@@ -7,13 +7,10 @@ import Routes from './config/Routes';
 import NotificationList from './screens/App/common/NotificationList';
 import createReducers from '../redux/reducers';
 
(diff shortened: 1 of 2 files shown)


Hyperkid123:
cc @AparnaKarve @AllenBW

Hyperkid123:
@miq-bot assign @AllenBW

himdel:
Added `hammer/yes` to match https://github.com/ManageIQ/manageiq-ui-classic/pull/4744

AparnaKarve:
@michaelkro @mturley Let us all validate this fix since it is making some foundational changes related to redux. Plus, it will be included in hammer.
Thanks!

PR #702, opened 2018-10-09
[#681] Add icons for mapping type to the infra mapping list view

### LINKED ISSUE #681: Show the Mapping type in the Infrastructure Mappings list view
We should show the mapping type in the list view now that we have multiple types (RHV and OpenStack). Question is where to put it?

mturley:
Closes #681.

This needs specific icons for OpenStack and RHV from @vconzola , currently the PR just uses pficon-cloud-tenant and pficon-virtual-machine.

screenshot 2018-10-09 19 02 55


mturley:
Vince says we'll keep this one as-is, for now. One possible solution might be to add a `<ListView.InfoItem>` at the front of the `additionalInfo` prop, with a static width and some text, but this clutters the row quite a bit. We might also wait until the brand team has icons for these products.

Issue #563, opened 2018-08-09
Explore using migrations plan status cards as selection controls for the plan list/card view below


vconzola:

michaelkro:
https://bugzilla.redhat.com/show_bug.cgi?id=1640594

PR #709, opened 2018-10-15
[#609] Infra Mappings Page

### LINKED ISSUE #609: Create a separate Infrastructure Mappings page available from Migration in the nav
infrastructure mappings page
Notes:

- Infrastructure Mappings will be available as an option under Migration in the CF navigation.
- Page should have breadcrumbs allowing navigation back to Migrate - Overview.
- List view of mappings should be filterable on Name and Description fields and sortable by Name and number of Associated Plans.
- All other behavior is the same as when infra mappings were on the Overview page.

michaelkro:
Fixes #609

# Notes
1. `<ListViewToolbar />` 
    - is an abstraction for `ListView` filtering, sorting, and pagination using the `renderProps` pattern [(React Docs)](https://reactjs.org/docs/render-props.html).
    - It contains and controls all state related to the aforementioned operations, and passes control to the composed component by passing methods via the renderProp function's argument list.
    - It is meant to be used at least also with `PlanRequestDetailList`, but I think it will also be possible to use with the migration plan lists on Overview.  Was thinking about doing this as a followup as this PR is already pretty massive.  But can do it all in one go, will let reviewers make the call
2. We need to poll `planRequest`s on the Infra Mappings page because the delete button for each mapping depends on whether or not it has any plans in progress
3. Transitions to Mappings page and automatically opens Mapping Wizard if user clicks on "Create Infrastructure Mapping" button in Overview empty state (no plans, no mappings)
4. Clicking on "Continue to plan" after mapping creation transitions user to Overview and automatically opens Plan Wizard
5. Did not get to refactoring `InfrastructureMappingList/helpers.js`.  Will let reviewers decide whether this should be done in this PR or as a followup
6. ~Did not move `MappingWizard` from `Overview/screens`.  Should live under `Mappings/screens`.  Will let reviewers decide whether this should be done in this PR or as a followup~

    ***EDIT:*** Done in 822490b
7.  We may want to add a link to the Infra Mappings Page on Overview.  I find it slow and clunky to have to use the vert nav every time I want to see my mappings
# Screens (Infra Mappings Page)
## ListView with sorting, filtering, and pagination
fullscreen_10_15_18__9_43_am

## Empty State
fullscreen_10_15_18__5_11_pm

## Error Loading Mappings
fullscreen_10_15_18__5_10_pm

## Insufficient Providers
fullscreen_10_15_18__5_09_pm

# Screens (Overview Page)
## No Mappings nor Plans
fullscreen_10_15_18__1_53_pm

## No Plans but Mappings Exist
fullscreen_10_15_18__1_55_pm


mturley:
An incredible amount of effort went into this! I particularly like the `ListViewToolbar` implementation. I've tested all the use cases you listed and found no issues! Still combing through the code.

> We may want to add a link to the Infra Mappings Page on Overview. I find it slow and clunky to have to use the vert nav every time I want to see my mappings

I agree, and I wonder if we should do this by making the Infrastructure Mappings card clickable as part of #563.

mturley:
One minor thing I noticed that I'm not sure how we would fix:

1. Navigate to the Infrastructure Mappings page using the main nav
2. Click on `Migration` in the breadcrumb bar to return to the Overview
3. Open the main nav again, and notice that the Infrastructure Mappings item is still styled as active even though we're on the Overview page:

screenshot 2018-10-16 15 16 25

I'm thinking this shouldn't block this PR, though, unless there is a simple fix. I suspect there is no easy way to update that rails-driven nav when our react-router updates.

AparnaKarve:
> Open the main nav again, and notice that the Infrastructure Mappings item is still styled as active even though we're on the Overview page:

@michaelkro This should fix the styling for the active menu item - can you give it a try?

```diff
diff --git a/app/controllers/migration_controller.rb b/app/controllers/migration_controller.rb
index 9383b20d..2215aaa1 100644
--- a/app/controllers/migration_controller.rb
+++ b/app/controllers/migration_controller.rb
@@ -6,7 +6,12 @@ class MigrationController < ApplicationController
 
   def index
     # this sets the active menu item, must match the item name in lib/manageiq-v2v/engine.rb
-    @layout = 'migration'
+    @layout = case params[:page]
+              when 'mappings'
+                'mappings'
+              else
+                'overview'
+              end
     @page_title = _('Migration')
   end
```

app/javascript/react/screens/App/Mappings/Mappings.js
```diff
@@ -166,9 +165,7 @@ class Mappings extends Component {
       <React.Fragment>
         <Toolbar>
           <Breadcrumb.Item href="/dashboard/maintab?tab=compute">{__('Compute')}</Breadcrumb.Item>
-          <li>
-            <Link to="/">{__('Migration')}</Link>
-          </li>
+          <Breadcrumb.Item href="/migration">{__('Migration')}</Breadcrumb.Item>
```
AparnaKarve: Pity, we can't take advantage of the more efficient `react-router-dom`

I'm facing a similar issue in https://github.com/ManageIQ/manageiq-v2v/pull/661 ...
When in the Migration screen, the `View Details` link does a full page load when transitioning to the Plan Detail page

michaelkro:
aww, I tried adding the layout code in [comment](https://github.com/ManageIQ/manageiq-v2v/pull/709#issuecomment-430386172) and no luck for me

Is it highlighting the correct menu item for you @AparnaKarve?

michaelkro:
ah, I restarted my servers and now it's working @AparnaKarve!

michaelkro:
Ok, I've added the [suggested layout code](https://github.com/ManageIQ/manageiq-v2v/pull/709#issuecomment-430386172) and reverted 1ea2e2a

Back to a nice snappy transition!  🎊

michaelkro:
Uh oh wait...  maybe it's not working lol

AparnaKarve:
Let me verify as well

AparnaKarve:
@michaelkro We might have to bring back 1ea2e2a98daccd7712b6f4d1ca28dfba4e18be39 unfortunately.

One other thing that needs to work is, when we go to `Infrastructure Mappings` page, then navigate elsewhere and then come back to Compute->Migration, it should land in the `Infrastructure Mappings` page and not the `Overview` page.

michaelkro:
> One other thing that needs to work is, when we go to Infrastructure Mappings page, then navigate elsewhere and then come back to Compute->Migration, it should land in the Infrastructure Mappings page and not the Overview page.

Def gonna need some help with this one.  When we navigate away, the V2V UI is unmounted and we lose all state.  Not sure how to track the last active page in this case :/

mturley:
Maybe we should finish reviewing and merging this PR and treat that as a separate issue, so we don't hold things up and create more rebase conflicts here?

michaelkro:
> Maybe we should finish reviewing and merging this PR and treat that as a separate issue, so we don't hold things up and create more rebase conflicts here?

I'm ok with this as this particular issue doesn't break any functionality

To this end, I've reverted the addition to `migrations_controller.rb` and brought back the commit that brings back the normal `Breadcrumb` component

vconzola:
@michaelkro I'm not sure I understand the scenario you're describing, but I do know that in CF when there are submenus the higher level menu items aren't active. So in our case Migration will never be active - only the submenu items Overview and Infrastructure Mappings. Does this help?

vconzola:
Whoops...take that back. I thought that's the way CF used to work (that's why Migration was never enabled as a breadcrumb link), but doesn't seem to be the case anymore. Maybe I'm confused. Regardless, it doesn't look like other parts of CF navigation go back to a previous state either.

michaelkro:
Thanks @vconzola :)

@AparnaKarve was referring to the actual behavior of clicking on the parent nav items.  For example, if you navigate to

```
Storage -> Block Storage -> Volume Backups
```

navigate away, and then click on

```
Storage -> Block Storage
```

It takes you back to `Volume Backups` automatically.  Currently it does not work that way with the V2V UI.

I still agree with @mturley that this particular issue can be handled as a followup PR.  Especially since, as of now, I have no idea how to make this happen for our stuff

AparnaKarve:
I agree that we could fix the above in a follow-up - it would be a change from the rails controller, I think.

So are we GTG here?
This will be a big change, especially since we are used to the old UI :)

michaelkro:
I feel good about it :)

As long the rest of us feel the same, I think we're good to go

@vconzola, @mturley, anything before we make the plunge?

simaishi:
Hammer backport details:

```
$ git log -1
commit e7ef845be02057b25a4bf5dbdfc0073e4f8e1a1c
Author: Aparna Karve <aparna.karve@gmail.com>
Date:   Wed Oct 17 10:31:27 2018 -0700

    Merge pull request #709 from michaelkro/609-mappings-page
    
    [#609] Infra Mappings Page
    
    (cherry picked from commit 13f5af1c5efe6b620cdf0e75a68693464472e69e)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1640275
```

PR #711, opened 2018-10-15
[#710] Don't show Schedule button when scheduled migration is about to start

### LINKED ISSUE #710: Unschedule button changes back to Schedule just before plan runs
https://bugzilla.redhat.com/show_bug.cgi?id=1618804#c11

From the BZ referenced above

> On latest 5.10.0.14 build,  Just before it starts running at scheduled time, "Unschedule" button changes to "Schedule" and then it starts running. So there is a brief window where :
> 1) User may think it is not scheduled any more.
> 2) User may try to Schedule it again.

mturley:
Fixes #710.

When a scheduled migration's time comes to run, there are 5-10 seconds of lag before the API recognizes that the migration has started. When we poll during this period, we were displaying the Schedule button again.

With this change, during that lag period, the scheduled time will be replaced by "Migration in progress" and the Unschedule and Migrate/Retry buttons will be disabled.

Not Started plan before the scheduled time:
screenshot 2018-10-15 16 30 01

Not Started plan during the lag period:
screenshot 2018-10-15 16 31 49

Failed plan before the scheduled time:
screenshot 2018-10-15 16 32 53

Failed plan during the lag period:
screenshot 2018-10-15 16 34 34

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
@@ -12,6 +12,7 @@ import ScheduleMigrationModal from '../ScheduleMigrationModal/ScheduleMigrationM
 import { formatDateTime } from '../../../../../../components/dates/MomentDate';
 import DeleteMigrationMenuItem from './DeleteMigrationMenuItem';
 import StopPropagationOnClick from '../../../common/StopPropagationOnClick';
+import getPlanScheduleInfo from './helpers/getPlanScheduleInfo';
 
 class MigrationsCompletedList extends React.Component {
   state = {
@@ -102,8 +103,7 @@ class MigrationsCompletedList extends React.Component {
                 )}
                 <ListView className="plans-complete-list" style={{ marginTop: 0 }}>
                   {sortedMigrations.map(plan => {
-                    const migrationScheduled = plan.schedules && plan.schedules[0].run_at.start_time;
-                    const staleMigrationSchedule = (new Date(migrationScheduled).getTime() || 0) < Date.now();
+                    const { migrationScheduled, staleMigrationSchedule, migrationStarting } = getPlanScheduleInfo(plan);
 
                     const requestsOfAssociatedPlan = allRequestsWithTasks.filter(
                       request => request.source_id === plan.id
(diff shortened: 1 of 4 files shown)


michaelkro:
I think this is the BZ for this issue https://bugzilla.redhat.com/show_bug.cgi?id=1618804#c11

app/javascript/react/screens/App/Overview/components/Migrations/ScheduleMigrationButton.js
```diff
@@ -14,13 +15,14 @@ const ScheduleMigrationButton = ({
   plan,
   isMissingMapping
 }) => {
-  const migrationScheduled = (plan.schedules && plan.schedules[0].run_at.start_time) || 0;
-  const staleMigrationSchedule = new Date(migrationScheduled).getTime() < Date.now();
+  const { migrationScheduled, staleMigrationSchedule, migrationStarting } = getPlanScheduleInfo(plan);
+  const showScheduleButton = staleMigrationSchedule && !migrationStarting;
```
michaelkro: This would allow a user to edit a schedule that was never run, correct?

Is this scenario actually possible @AparnaKarve?  If so, are there any possible negative side effects of editing a stale schedule to run in the future?

AparnaKarve: We do allow editing a schedule that was never run, in other areas of the UI (in Configuration -> Schedules)

Can't think of any possible negative effects of editing a stale schedule to run in the future.
I think this is good!

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -254,14 +254,20 @@ class MigrationsCompletedList extends React.Component {
...
                                 <Icon type="fa" name="clock-o" />
                                 {__(`Migration scheduled`)}
                                 <br />
                                 {formatDateTime(migrationScheduled)}
                               </ListView.InfoItem>
-                            )
+                            ),
+                          migrationStarting && (
+                            <ListView.InfoItem key={`${plan.id}-starting`} style={{ textAlign: 'left' }}>
+                              {__(`Migration in progress`)}
```
AparnaKarve: gettext, if I am not mistaken, supports only single quotes or double quotes.

backticks are not supported.

@mturley Can you change all backticks here and below to single quotes?

/cc @mzazrivec

mturley: Whoops! Looks like this came from copy-pasting @AllenBW's use of backticks for the Migration Scheduled getText call. I'll fix them. Good catch!

app/javascript/react/screens/App/Overview/components/Migrations/ScheduleMigrationButton.js
mturley: The existing code already allowed editing a schedule that was never run, I believe.

simaishi:
Hammer backport details:

```
$ git log -1
commit 43f99c2968faf92fbd0ee3afde8c50b6a72694e9
Date:   Tue Oct 16 14:24:57 2018 -0400

    Merge pull request #711 from mturley/710-no-confusing-schedule-button
    
    [#710] Don't show Schedule button when scheduled migration is about to start
    
    (cherry picked from commit a561e5e20170af108c62e709e916d8917bd43d73)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1618804
```

PR #721, opened 2018-10-18
convert `scheduleTime` to a Date object before passing to backend

AparnaKarve:
Converts `scheduleTime` to a Date object before passing to backend.

This seems to be kicking off scheduled migrations without any other timezone related settings applied to the appliance.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1630446

@mturley Can you confirm if this indeed fixes the issue?
Thanks.

### DIFF
--- a/app/javascript/react/screens/App/Overview/OverviewActions.js
+++ b/app/javascript/react/screens/App/Overview/OverviewActions.js
@@ -203,7 +203,7 @@ export const scheduleMigration = payload => dispatch =>
       let body = {
         action: 'order',
         resource: {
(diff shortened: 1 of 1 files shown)


AparnaKarve:
I did some testing on an appliance which did not have this change, and it appears that migrations are getting kicked off just fine.

So closing this PR for 2 reasons -
1. This seems to be a non-issue on an appliance, at least with the current master branch
2. As @mturley said,  changing it to a Date object does not really make a difference

It could be quite possible that this is a Gaprindashvili bug, and we would just have to confirm that.

PR #742, opened 2018-10-26
[#715] Link directly from the Overview page to the Infrastructure Mappings page with filter

### LINKED ISSUE #715: Link directly from the Overview page to the Infrastructure Mappings page
User should have a way to link directly from the Overview page to the InfraStructure Mappings page without having to use the left navigation. i think we have a couple of options. (1) find a place to put a big ol' ugly link on the page - probably in the listview header, below(?) the Create Migration plan link, or (2) make the name of the associated mapping for each plan a link that would take the user to the infra mappings page with the list scrolled or filtered to the correct mapping. Leaning toward option (2).

mturley:
Closes #715.
https://bugzilla.redhat.com/show_bug.cgi?id=1643610

On the Overview list views, the name of the associated infra mapping is now a link:

screenshot 2018-10-26 13 22 45

When this new link is clicked, the user is redirected to the mappings page and an initial filter is applied with that mapping's name:

screenshot 2018-10-26 13 22 59

When the `Mappings` component unmounts (the user is leaving the Infrastructure Mappings page without a full page refresh, such as by completing the mappings wizard and continuing to the plan wizard), this initial filter is cleared, just in case.


app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -235,7 +236,19 @@ const MigrationsCompletedList = ({
...
-                              {plan.infraMappingName}
+                              <a
+                                href="#"
+                                onClick={e => {
+                                  e.preventDefault();
+                                  filterMappingsListOnTransitionAction({
+                                    filterTypeId: 'name',
+                                    filterValue: plan.infraMappingName
+                                  });
+                                  redirectTo('/mappings');
```
michaelkro: It looks like using the `redirectTo` helper is messing up the vert nav (i.e., the highlighted nav item is not correct after transition)

app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
```diff
@@ -235,7 +236,19 @@ const MigrationsCompletedList = ({
...
                           !isMissingMapping && (
                             <ListView.InfoItem key={`${plan.id}-infraMappingName`}>
-                              {plan.infraMappingName}
+                              <a
+                                href="#"
+                                onClick={e => {
+                                  e.preventDefault();
+                                  filterMappingsListOnTransitionAction({
+                                    filterTypeId: 'name',
+                                    filterValue: plan.infraMappingName
```
michaelkro: There may not be a way to solve this, but because our filters match on partial strings, `infraMappingName` is not always going to be unique enough to obtain the exact mapping we want.

For example, you could have a plan created with mapping `test`, but also have a mapping named `test #2`

mturley:
After discussion with @vconzola, we decided to do away with the filtering by name. Closing this in favor of the simpler #745

PR #745, opened 2018-10-29
[#715] Link directly from the Overview page to the Infra Mappings page without filter

### LINKED ISSUE #715: Link directly from the Overview page to the Infrastructure Mappings page

mturley:
Closes #715.
https://bugzilla.redhat.com/show_bug.cgi?id=1643610

This is an alternative to #742 which does not filter the mappings list upon arrival.

On the Overview list views, the name of the associated infra mapping is now a link:

screenshot 2018-10-26 13 22 45

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsCompletedList.js
@@ -235,7 +235,7 @@ const MigrationsCompletedList = ({
                           ),
                           !isMissingMapping && (
                             <ListView.InfoItem key={`${plan.id}-infraMappingName`}>
-                              {plan.infraMappingName}
+                              <a href="/migration/mappings#">{plan.infraMappingName}</a>
                             </ListView.InfoItem>
                           ),
                           <ListView.InfoItem key={`${plan.id}-elapsed`}>
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsNotStartedList.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationsNotStartedList.js
@@ -152,7 +152,7 @@ const MigrationsNotStartedList = ({
                           ),
                           !isMissingMapping && (
                             <ListView.InfoItem key={`${plan.id}-infraMappingName`}>
-                              {plan.infraMappingName}
+                              <a href="/migration/mappings#">{plan.infraMappingName}</a>
                             </ListView.InfoItem>
                           ),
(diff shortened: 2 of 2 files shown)


mturley:
This will be incompatible with https://github.com/ManageIQ/manageiq-v2v/pull/769, so we should wait for that to merge and update this one to use `href="/migration/mappings"`

AparnaKarve:
I do like the idea of navigating to the Mappings page (without a full page reload of the Mappings page) from the Overview screen - but when you click on a specific Mapping, I guess somewhere in my mind I had the expectation that, that particular Mapping item would be selected in the list.

Would it be possible to select that specific item from the list?
(Note: This is just a question -- just wondering if it is something that is doable)

----------

> This will be incompatible with #769

Yes, let's wait on that PR.
Sorry that you might have to compromise with a full page reload on the Mappings page :(

mturley:
> when you click on a specific Mapping, I guess somewhere in my mind I had the expectation that, that particular Mapping item would be selected in the list.

I had initially implemented something close to this, by filtering on the mapping name after the page transition in #742. @vconzola and @michaelkro didn't like that the results included any mappings that had the selected mapping's name as a substring of their name, so we decided it was less confusing not to filter at all.

I would be open to revisiting that decision (maybe introduce some way to do an exact match filter there), but our hands might be tied by the full page reload anyway.

vconzola:
@mturley My thinking on this is that IRL the number of infrastructure mappings will be very small and so that they will all likely be visible on the first page without filtering. Not ideal, but I think @michaelkro said that filtering will be a good deal of work. We can certainly revisit if we get feedback that tells us what we have isn't sufficient.

mturley:
That makes sense.

simaishi:
Hammer backport details:

```
$ git log -1
commit 99fbcd2337c78e506a0bc6fc2f9c0c29c3e17ee2
Date:   Mon Nov 19 17:35:02 2018 -0500

    Merge pull request #745 from mturley/715-mapping-link-no-filter
    
    [#715] Link directly from the Overview page to the Infra Mappings page without filter
    
    (cherry picked from commit 691f0a74b63603c97dad7b4211ab6a6c35d06d1c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1643610
```

PR #763, opened 2018-11-01
[#761] [Code Cleanup] Incorporate <ListViewToolbar /> with PlanRequestDetailList

### LINKED ISSUE #761: [Code Cleanup] Incorporate <ListViewToolbar /> with PlanRequestDetailList
This will be a refactor with no new functionality. To test it, we should verify that the filtering, sorting and pagination still works correctly in the list view displayed on the detail page for a migration plan that has been started or completed.

This is a sub-issue of #716.

https://bugzilla.redhat.com/show_bug.cgi?id=1645714

### LINKED ISSUE #767: PlanRequestDetailList: When a plan finishes, available filters change but current filter selection does not update
Discovered as part of the work for https://github.com/ManageIQ/manageiq-v2v/pull/763.

When a plan finishes while the user is on the `PlanRequestDetailList`, the `filterTypes` and `sortFields` are updated to different available options. However, the `currentFilterType` and `currentSortType` do not update after this change, so the user could be left in a scenario where they can select and apply irrelevant filters.

The code in `getDerivedStateFromProps` at the top of `PlanRequestDetailList` should also reset the relevant state to the defaults from the new sort and filter options when they change.

mturley:
Closes #761.
Closes #767.

https://bugzilla.redhat.com/show_bug.cgi?id=1645714

# Notes

* The `PlanRequestDetailList` has an interesting property that required a change to `ListViewToolbar`: when a plan finishes, its available filter types and sort fields change. There is actually a bug currently in `PlanRequestDetailList` where the currently selected filter will not properly update after this change, so I introduced the `componentDidUpdate` lifecycle method in `ListViewToolbar` to reset the current filter and sort selections to their defaults when this change occurs (I wrote this up in #767, but fixed it here as part of the abstraction). We were also unnecessarily mirroring `filterTypes` and `sortFields` in state, when they fit better as props.

* Because `PlanRequestDetailList` requires access to the filtered items in the scope of its class methods for the purposes of "select all" behavior, I was unable to use `ListViewToolbar` as a child in the render method. (I thought this made the abstraction incompatible, which is why this was closed). But instead I wrapped the whole `PlanRequestDetailList` in a `ListViewToolbar` and passed the filtered items down as a prop, making it a perfect fit (which is why this was reopened!)

### DIFF
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -182,6 +182,7 @@ class Plan extends React.Component {
                 filterTypes={planFinished ? FINISHED_PLAN_FILTER_TYPES : ACTIVE_PLAN_FILTER_TYPES}
                 sortFields={planFinished ? FINISHED_PLAN_SORT_FIELDS : ACTIVE_PLAN_SORT_FIELDS}
                 listItems={planRequestTasksMutable}
+                defaultSortTypeIndex={1}
               >
                 {({
                   filteredListItems,
--- a/app/javascript/react/screens/App/Plan/Plan.js
+++ b/app/javascript/react/screens/App/Plan/Plan.js
@@ -8,6 +8,13 @@ import PlanRequestDetailList from './components/PlanRequestDetailList/PlanReques
 import PlanVmsList from './components/PlanVmsList';
 import PlanEmptyState from './components/PlanEmptyState';
 import getMostRecentRequest from '../common/getMostRecentRequest';
+import ListViewToolbar from '../common/ListViewToolbar/ListViewToolbar';
+import {
+  ACTIVE_PLAN_FILTER_TYPES,
+  FINISHED_PLAN_FILTER_TYPES,
+  ACTIVE_PLAN_SORT_FIELDS,
+  FINISHED_PLAN_SORT_FIELDS
+} from './components/PlanRequestDetailList/PlanRequestDetailListConstants';
 
 class Plan extends React.Component {
   static getDerivedStateFromProps(nextProps, prevState) {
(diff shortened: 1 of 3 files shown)


mturley:
Since the "Select tasks for cancel" behavior in `PlanRequestDetailList` requires knowledge of the active filters (so that Select All can select only tasks that match the filter), this refactor would require abstraction of that selection behavior into `ListViewToolbar`. After discussion with @michaelkro, we've decided our abstraction doesn’t work for this particular ListView.  If and when we need to provide select/checkbox behavior on list view rows in another view, we can revisit adding that behavior to `ListViewToolbar` and revisiting this.

mturley:
I'm reopening this for now because I figured out a potential way to work around the problem outlined above, by making the whole PlanRequestDetailList a child of a ListViewToolbar.

mturley:
@michaelkro I decided some friday night refactoring was in order, and this ended up working out great. 🍺

app/javascript/react/screens/App/common/ListViewToolbar/ListViewToolbar.js
```diff
@@ -8,11 +8,9 @@ import paginate from './paginate';
 
 class ListViewToolbar extends Component {
   state = {
-    filterTypes: this.props.filterTypes,
     currentFilterType: this.props.filterTypes[this.props.defaultFilterTypeIndex],
     currentValue: '',
     activeFilters: [],
-    sortFields: this.props.sortFields,
```
michaelkro: Yes, thank you for taking these out of the state, makes much more sense!

app/javascript/react/screens/App/Plan/Plan.js
```diff
@@ -171,29 +178,50 @@ class Plan extends React.Component {
...
-                markedForCancellation={markedForCancellation}
-                failedMigrations={failedMigrations}
-                successfulMigrations={successfulMigrations}
-                notificationsSentList={notificationsSentList}
-                dispatchVMTasksCompletionNotificationAction={dispatchVMTasksCompletionNotificationAction}
-              />
+              <ListViewToolbar
+                filterTypes={planFinished ? FINISHED_PLAN_FILTER_TYPES : ACTIVE_PLAN_FILTER_TYPES}
+                sortFields={planFinished ? FINISHED_PLAN_SORT_FIELDS : ACTIVE_PLAN_SORT_FIELDS}
+                listItems={planRequestTasksMutable}
```
michaelkro: The default sort type for `PlanRequestDetailList` should be `name`

app/javascript/react/screens/App/common/ListViewToolbar/ListViewToolbar.js
```diff
@@ -288,6 +300,7 @@ class ListViewToolbar extends Component {
   };
 
   render() {
+    const filteredListItems = this.filterListItems();
```
michaelkro: Can we move the return value of this function call to the argument object below :)

app/javascript/react/screens/App/Plan/Plan.js
mturley: Ah, good catch

app/javascript/react/screens/App/common/ListViewToolbar/ListViewToolbar.js
mturley: The reason I didn't move this to the argument object is because I wanted to both pass it directly into the object, and also pass it into `this.filterSortPaginateListItems(filteredListItems)`, so we don't have to run the filtering twice per render. I think instead I should actually remove the one we use as a default argument for `filterSortPaginateListItems`, as that is no longer exposed.

michaelkro: Ah, oops, sorry I totally missed that :P

simaishi:
Hammer backport details:

```
$ git log -1
commit 6fb60819014cd4ec1e6e92049d3ee7d15e1cbfb9
Date:   Tue Nov 6 16:04:14 2018 -0500

    Merge pull request #763 from mturley/761-filter-paginate-plan-request
    
    [#761] [Code Cleanup] Incorporate <ListViewToolbar /> with PlanRequestDetailList
    
    (cherry picked from commit 55244b764666572e92ab5cbc05eabd94219c0559)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1645714
```

PR #779, opened 2018-11-08
Remove outdated mount_react_component.

Hyperkid123:
mount_react_component was replaced in ManageIQ/manageiq-v2v#533 and is no longer used.


Hyperkid123:
@miq-bot add_label refactoring
@miq-bot assign @michaelkro

PR #782, opened 2018-11-08
[#776] Add Migration Settings page with max concurrent migrations per host setting

### LINKED ISSUE #747: v2v - Expose number of concurrent migrations per conversion host in UI
Issue:
Expose this in the Migration UI so that consumers can change how many concurrent VM migrations are allowed per conversion host

BZ:  https://bugzilla.redhat.com/show_bug.cgi?id=1613848

### LINKED ISSUE #776: Add Max Concurrent Migrations Setting to the UI
As a first step toward migration throttling we need to add a setting for maximum concurrent migrations. Mockup:
migration settings page
Notes:
Default value is 10.
Apply button is disabled until the selected value is changed. When Apply is selected disable the button until we know that the change was successful.
Will provide text for the tooltip later.

mturley:
Closes #776.
Closes #747.

screenshot 2018-11-09 09 52 45

This will conflict with #747 in `engine.rb`, we'll need to change the new route to be `/migration/settings` rather than `/migration#/settings`.

https://bugzilla.redhat.com/show_bug.cgi?id=1613848


app/javascript/react/screens/App/Settings/Settings.js
```diff
@@ -0,0 +1,105 @@
...
+              <Grid.Col componentClass={Form.ControlLabel} sm={5}>
+                {__('Maximum concurrent migrations per conversion host')}
+              </Grid.Col>
+              <Grid.Col sm={7}>
+                <div style={{ width: 100 }}>
+                  <Field
+                    id="max_concurrent_tasks_per_host"
+                    name="max_concurrent_tasks_per_host"
+                    component={NumberInput}
+                    normalize={NumberInput.normalizeStringToInt}
```
michaelkro: I'm probably just missing something, but can you clarify where this is being used?  Also, I think this prop is being passed down into the `NumberInput` component, so if `normalizeStringToInt` is an instance property of the class, do we need it here?

michaelkro: nvm, just checked.  This is a special `redux-form` thing.  Nothing to see here

mturley:
@michaelkro That form field placement is a little weird on larger screens, I'll admit. The label is just right-aligned in a column that takes up 5/12 of the screen width, it looks more natural on a smaller screen. Maybe we shouldn't bother using `Grid.Col` for those?

simaishi:
Hammer backport details:

```
$ git log -1
commit 36be3599432eefa2eea10fe64628dd7d779e376a
Date:   Fri Nov 9 10:58:15 2018 -0500

    Merge pull request #782 from mturley/776-settings-page
    
    [#776] Add Migration Settings page with max concurrent migrations per host setting
    
    (cherry picked from commit acd891245c35df9219ffd10f0de1e4ad41fdd109)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1613848
```

PR #757, opened 2018-10-31
[Edit Migration Plan] Allow changing the associated infrastructure mapping

mturley:

PR #788, opened 2018-11-13
Enable editing orphaned plans and handle undefined mapping when editing

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1599868

In https://github.com/ManageIQ/manageiq-v2v/pull/757, I made it possible to change the associated mapping for a plan, but I did not enable editing of a plan with a missing mapping, which is the original reason we wanted that functionality. This PR enables that edit button for orphaned plans, and handles some errors that were caused by proceeding through the wizard with an orphaned plan.

We also introduce an error message here when the user attempts to create a new plan, or edit an existing plan, while there are no mappings:
screenshot 2018-11-14 10 56 43

### DIFF
--- a/app/javascript/react/screens/App/Overview/Overview.js
+++ b/app/javascript/react/screens/App/Overview/Overview.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Card, Breadcrumb, CardGrid, Spinner } from 'patternfly-react';
+import { Card, Breadcrumb, CardGrid, Spinner, Icon } from 'patternfly-react';
 
 import Toolbar from '../../../config/Toolbar';
 import * as AggregateCards from './components/AggregateCards';
@@ -66,13 +66,12 @@ class Overview extends React.Component {
       isContinuingToPlan,
       fetchTransformationPlansUrl,
       fetchTransformationPlansAction,
-      planWizardId,
-      showPlanWizardAction
+      planWizardId
     } = this.props;
     const { hasMadeInitialPlansFetch } = this.state;
 
     if (isContinuingToPlan !== nextProps.isContinuingToPlan && !nextProps.isContinuingToPlan) {
-      showPlanWizardAction(planWizardId);
+      this.showPlanWizardOrError(planWizardId);
     }
 
     // kill interval if a wizard becomes visble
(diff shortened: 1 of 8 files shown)


mturley:
I'll defer to @vconzola on this, but I agree that's a weird scenario you could end up in (deleting all mappings and then editing an orphaned plan, and having no mappings to choose from). I wonder if we should prevent editing a plan if no mappings exist? or maybe as you suggested, having some warning like "No infrastructure mappings are available. Please visit the Infrastructure Mappings page to create one", with a link to that page?

vconzola:
@mturley I like option B.

mturley:
@vconzola just some exact phrasing, if you don't mind. I'm thinking I'll just use the existing warning mechanism, so it'll look just like this one:

screenshot 2018-11-02 12 50 20

If that's okay, no mockup needed.

vconzola:
@mturley  We can't use the same warning mechanism because we need to prevent the user from selecting Next and moving on if there are no infra mappings. Now that mappings are on their own page, what happens if there are no mappings and the user tries to create a plan? We need to talk through this.

mturley:
The user can't select Next and move on without selecting a mapping first, which is impossible if there are no mappings. This applies both when editing and creating a new plan. You'd see the warning and be unable to proceed.

mturley:
Notes from today's meeting: instead of using the above warning inside the wizard, we'll block the wizard from opening, and display a modal warning (or should it be a modal error @vconzola?) instead of the wizard. This will display if you try to either edit a plan or create a new plan, when there are no mappings.

vconzola:
@mturley Use the PF Message Dialog pattern (Error dialog) [https://www.patternfly.org/pattern-library/communication/message-dialog/#design] with title: "Migration plan error", primary message: "No infrastructure mapping exists" and body: "A migration plan must include an infrastructure mapping. Go to the Infrastructure Mappings page to create one."  Button: "Close". If possible, "Infrastructure Mappings page" should be a link.

mturley:
This one should now be good to go.

simaishi:
Hammer backport details:

```
$ git log -1
commit 17fc1605ed81de37de7f8a6a437b3cd8d416f0f8
Date:   Thu Nov 15 08:17:55 2018 -0500

    Merge pull request #788 from mturley/bz1599868-edit-orphaned-plan
    
    Enable editing orphaned plans and handle undefined mapping when editing
    
    (cherry picked from commit 828e1ef5a1b8a25e55a364a05d6c512f01242ab3)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1599868
```

PR #790, opened 2018-11-13
[HAMMER] Hide Migration UI by default for CF 4.7 Public Beta

AparnaKarve:
The v2v Migration menu would be hidden and inaccessible by default.

To enable it, add `:ims: true` under the `product` settings 

https://bugzilla.redhat.com/show_bug.cgi?id=1649474


dclarizio:
@simaishi can you close this PR?  We'll do this using the 5.10 MR approach you suggested. Thx, Dan

Issue #478, opened 2018-07-10
V2V should differentiate whether logs are moved or not generated


Yadnyawalkya:
**Description**
This is more of usability enhancement. V2V should differentiate whether logs are moved or not generated.
Sometime failed migration do not generate logs but while 'clicking on download logs' of details page error screen shows 'The location of transformation log was not set. Download of transformation log aborted'. This is somehow confusing since 'transformation log location is set'. We should show error with lines which differentiate whether logs are moved from transformation log directory or not generated yet.

**Version-Release number**
5.9.3.4

**Investigation**
If migration succeeds and log is created but moved, then getting following error.

`Failed to download "ytale-v2v-4-win10.log". Reason - "SCP did not finish successfully (1): scp: /var/log/vdsm/import/v2v-import-20180704T063253-126596.log: No such file or directory "`

and if migration fails and log is not get created, getting following error

```Failed to download "ytale-v2v-4-win10.log". Reason - "Conversion host was not found: ID []. Cannot queue the download of transformation log."```

Should be more specific as per description.

**Internal reference**
https://bugzilla.redhat.com/show_bug.cgi?id=1589265

vconzola:
@AparnaKarve Can you take a look at this one when you get a chance? It's on Brett's list for 1.1 release.

AparnaKarve:
@vconzola This particular error -
`Failed to download "ytale-v2v-4-win10.log". Reason - "Conversion host was not found: ID []. Cannot queue the download of transformation log."` 
occurs when we try to download the log too early in the migration process. 
So I am wondering that maybe we should keep the 'Download' button disabled until we have information about the log file location in the conversion host.

This error on the other hand -
`Failed to download "ytale-v2v-4-win10.log". Reason - "SCP did not finish successfully (1): scp: /var/log/vdsm/import/v2v-import-20180704T063253-126596.log: No such file or directory "`

is a proper download error, and the above message appears from the backend.
I think we should just have a more user friendly message here.

vconzola:
@AparnaKarve I agree - disabling the button until the file is available is a good idea. In the case of playbook logs we should just not offer the migration log in the dropdown until it's available.

vconzola:
@AparnaKarve Those messages look good to me.

bthurber:
@vconzola @AparnaKarve checking status on dev complete for this

AparnaKarve:
[...]

PR #793, opened 2018-11-15
Parse complex backend messages and display them in a simplified format

AparnaKarve:
Parse complex user-unfriendly messages from the backend and use the dictionary to interpret/translate them into simplified and more understandable messages.

Fixes https://github.com/ManageIQ/manageiq-v2v/issues/478


AparnaKarve:
@michaelkro @mturley Let me know what you think of this approach. Thanks.

app/javascript/react/screens/App/Plan/helpers.js
```diff
@@ -115,3 +115,23 @@ export const incompleteCancellationTasks = (requestWithTasks, actions, tasksForC
   const completedTasksOfPlan = tasksOfPlan.filter(task => task.state === 'finished');
   return removeCancelledTasksFromSelection(tasksForCancel, completedTasksOfPlan);
 };
+
+export const parseComplexErrorMessages = complexErrorMessage => {
+  const regex1 = RegExp('Conversion host was not found. Download of transformation log aborted.', 'g');
```
mturley: Nit-pick, take it or leave it, but maybe we should declare each of these regexes above the condition where they are used, so it's easier to correlate them if we add more?

AparnaKarve: > Nit-pick, take it or leave it

Take it! :)
I have moved those RegExp constants in the constants JS

app/javascript/react/screens/App/Plan/helpers.js
```diff
@@ -115,3 +120,19 @@ export const incompleteCancellationTasks = (requestWithTasks, actions, tasksForC
   const completedTasksOfPlan = tasksOfPlan.filter(task => task.state === 'finished');
   return removeCancelledTasksFromSelection(tasksForCancel, completedTasksOfPlan);
 };
+
+export const parseComplexErrorMessages = complexErrorMessage => {
+  if (V2V_BACKEND_ERROR_MESSAGES.V2V_DOWNLOAD_LOG_STATUS_MESSAGES.exec(complexErrorMessage) !== null) {
```
michaelkro: When switching to this branch, the error notification seems to stop working.  By any chance, should the conditional be checking

```js
V2V_BACKEND_ERROR_MESSAGES.DOWNLOAD_LOG_ERROR_NO_HOST.exec(...) !== null
```

mturley: Ah, I think Mike is right here... I admit I hadn't tested it again after the second commit. I need to break that habit.

AparnaKarve: Thanks @michaelkro @mturley -- I have made the above adjustment.

... and sorry for the hiatus on this PR (was tied up in other things)

app/javascript/react/screens/App/Plan/helpers.js
```diff
@@ -115,3 +120,16 @@ export const incompleteCancellationTasks = (requestWithTasks, actions, tasksForC
   const completedTasksOfPlan = tasksOfPlan.filter(task => task.state === 'finished');
   return removeCancelledTasksFromSelection(tasksForCancel, completedTasksOfPlan);
 };
+
+export const parseComplexErrorMessages = complexErrorMessage => {
```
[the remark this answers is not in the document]
AparnaKarve: resolved in f982c245a4f3a1c4df755951eb60dae511d9bdb4

app/javascript/react/screens/App/Plan/PlanConstants.js
```diff
@@ -73,4 +73,21 @@ STATUS_MESSAGES[STATUS_MESSAGE_KEYS.VM_MIGRATED] = __('Virtual machine migrated'
 STATUS_MESSAGES[STATUS_MESSAGE_KEYS.VM_TRANSFORMATIONS_FAILED] = __('VM migrations failed');
 STATUS_MESSAGES[STATUS_MESSAGE_KEYS.CANCELLED] = __('VM cancelled');
 
+const DOWNLOAD_LOG_STATUS_MESSAGES = {};
+DOWNLOAD_LOG_STATUS_MESSAGES.DOWNLOAD_LOG_ERROR_NO_HOST = __('Conversion host was not found');
+DOWNLOAD_LOG_STATUS_MESSAGES.DOWNLOAD_LOG_ERROR_NO_CREDS_SET = __('Credentials not set on conversion host');
+DOWNLOAD_LOG_STATUS_MESSAGES.DOWNLOAD_LOG_ERROR_SCP_ISSUE = __('Log file was not found on conversion host');
+DOWNLOAD_LOG_STATUS_MESSAGES.DOWNLOAD_LOG_ERROR_UNKNOWN_ISSUE = __('Unknown error');
```
michaelkro: Since there is no corresp property in `BACKEND_ERROR_MESSAGES`, this message will never be seen, correct?

AparnaKarve: Correct.

I intended on using it initially, but then figured it would be better to just return the original unparsed/unprocessed message instead.
Referring to https://github.com/ManageIQ/manageiq-v2v/pull/793/files#diff-0405502a3437597af99611db403b6a76R136

If it's causing confusion, we could remove it.

michaelkro: I def agree with the decision to display the unprocessed message for the unknown cases

I have to admit that it did confuse me initially, seeing `return complexErrorMessage` as well as this property.  Can we remove it for now?  We can always bring it back if it becomes necessary

AparnaKarve: Ok, no problem. Done.

PR #816, opened 2018-12-04
Replace broken axios API mock adapter, fix broken tests

mturley:
I was about to write a test for `POST /api/settings`, when @michaelkro pointed out a problem...

Ever since our move from `axios` to manageiq's global `API` in https://github.com/ManageIQ/manageiq-v2v/pull/303, the `mockRequest()` calls in our tests have been doing nothing, and some of our tests were using `.catch(() => expect(...))` on promises that would never be rejected, asserting nothing.

This PR replaces the `global.API.get = jest.fn(() => Promise.resolve());` with a simple implementation of an API mock adapter, that simply allows us to specify `method, url, status, response` when mocking and then either resolves or rejects the promise when API calls happen. Old tests are new again! And now I can write new API-mocked tests that are meaningful.

In the `distant future`, I agree with @priley86 in his comments in #303 that we could be doing something better:

> in a future architecture in a galaxy far far away... i would love to be able to document such APIs / request headers via something like an api-docgen

but for now, this gives us something a little more useful to test with.

### DIFF
--- a/.jest-setup.js
+++ b/.jest-setup.js
@@ -1,5 +1,7 @@
 import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import { APImock } from './app/javascript/common/mockRequests';
+
 configure({ adapter: new Adapter() });
 
 // Mocking translation function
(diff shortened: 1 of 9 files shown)


simaishi:
Hammer backport details:

```
$ git log -1
commit 8e04c0172af08cb22504d4d885171c7ca91da99e
Date:   Tue Dec 4 14:47:29 2018 -0500

    Merge pull request #816 from mturley/simple-api-mock
    
    Replace broken axios API mock adapter, fix broken tests
    
    (cherry picked from commit ed5249d75f12aa0b2b6b830849ef005d30a0c5f6)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1657285
```

PR #817, opened 2018-12-05
Add unit tests for Settings page components, actions and reducers

mturley:
Coverage report for new tests only: http://amusing-deer.surge.sh/


app/javascript/react/screens/App/Settings/__tests__/index.test.js
```diff
@@ -0,0 +1,57 @@
...
+import { reducer as formReducer } from 'redux-form';
+
+import Settings from '../Settings';
+import SettingsContainer, { reducers } from '../index';
+import { initialState } from '../SettingsReducer';
+import { servers, settings } from '../settings.fixures';
+
+import { mockRequest } from '../../../../../common/mockRequests';
+
+describe('Settings integration test', () => {
```
michaelkro: I'd like to discuss not continuing with these index.js integration tests. We can keep the ones that are already in place, but it's been my experience that they are brittle, and failures often are false positives that stem from test environment configuration issues, not actual issues with code/changes.

For example, in the edit mapping PR, the test for MappingWizard needed configuration updates to pass, despite there not having been any actual regressions.

While I think it's not a bad thing to have checks in place for our overall system architecture/configuration, there should probably only be one test at the highest level, not for every container component.

Thoughts?

mturley: I think this is reasonable. All we're testing in these integration tests is that the `mapStateToProps` function applies a mocked store's state to the component as specified, and in the case where we just return some keys from the store, the test is a bit of a tautology. I think they could be useful if we had complex logic in `mapStateToProps`, but we should instead favor moving that logic to selectors/helpers, and testing those helpers separately.

I'll remove this one.

app/javascript/react/screens/App/Settings/__tests__/Settings.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Spinner } from 'patternfly-react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
```
michaelkro: Let's set this up for all tests.  Can you please add

```
"snapshotSerializers": ["enzyme-to-json/serializer"]
```

to our [package.json](https://github.com/ManageIQ/manageiq-v2v/blob/master/package.json#L92)?

Then, we can just do

```js
expect(component).toMatchSnapshot();
```

mturley: Sure! Didn't know you could do that.

app/javascript/react/screens/App/Settings/__tests__/Settings.test.js
```diff
@@ -0,0 +1,36 @@
...
+import { Settings } from '../Settings';
+import { servers, settings } from '../settings.fixures';
+import { getFormValuesFromApiSettings } from '../helpers';
+
+describe('Settings component', () => {
+  const defaultFormValues = getFormValuesFromApiSettings(settings);
+  const baseProps = {
+    fetchServersAction: jest.fn(),
+    fetchSettingsAction: jest.fn(),
+    patchSettingsAction: jest.fn(),
```
michaelkro: While it doesn't affect the tests here, it's not a bad idea to make sure that each test case gets a fresh set of mocks by doing something like 

```js
let myMock;
beforeEach(() => {
  myMock = jest.fn();
});
```
Jest mock functions keep track of their number of calls and results.  See [second code snippet here](https://jestjs.io/docs/en/mock-functions).

You won't be able to store them in a `baseProps` object any more, unfortunately, and since you're not checking those values in these tests, I'll leave it up to you whether or not to leave this as is

app/javascript/react/screens/App/Settings/__tests__/Settings.test.js
```diff
@@ -0,0 +1,36 @@
...
+  it('renders the settings page with Applying spinner when saving', () => {
+    const component = shallow(<Settings {...baseProps} isSavingSettings />);
+    const applying = component.find(Spinner).findWhere(child => child.text() === ' Applying...');
+    expect(applying).toHaveLength(1);
+  });
+
+  it('properly calls patchSettingsAction on apply click', () => {
+    const patchSettingsAction = jest.fn();
+    const component = shallow(<Settings {...baseProps} patchSettingsAction={patchSettingsAction} />);
+    component.instance().onApplyClick();
```
michaelkro: We can make this test a little more "integration-y" by replacing this line with

```js
component.find('Button').simulate('click');
```

simaishi:
Hammer backport details:

```
$ git log -1
commit f3b16cd7f626fe674063e51b7134fd4989135d86
Date:   Thu Dec 6 14:00:08 2018 -0500

    Merge pull request #817 from mturley/settings-tests
    
    Add unit tests for Settings page components, actions and reducers
    
    (cherry picked from commit 430933a767febaf649edc4477c4640c05ec7477a)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1657285
```

Issue #811, opened 2018-11-27
Phase 2 of vertical menu fixes


mzazrivec:
Pull request https://github.com/ManageIQ/manageiq-v2v/pull/769 fixed several issues regarding vertical menu. With that PR in place, vertical menu highlighting works as expected. An unfortunate effect of this PR though is loss of navigation history and client side routing (I replaced `ConnectedRouter` with `BrowserRouter`).

Here's what I'm proposing to remedy the above:

1. We'd have 3 menu items for migration (and also 3 rails routes in the plugin). That way we'd have vert. menu updates whenever we click on some of the migration menu items.
2. We'd go back to using ConnectedRouter and Link, which would give us back navigation history.
3. on the backend (or better yet classic-ui) side, we'd have an action, let's call it vertical_menu_ping(), which the client (v2v react in this case) would call anytime it would navigate between overview, plain id and migration mappings screens. This backend routine would update vert. menu data in server session.

mturley:
This sounds like a great solution, and I also just want to tack on this issue I made to track things that we'll want to follow up on as part of this: https://github.com/ManageIQ/manageiq-v2v/issues/803

Issue #802, opened 2018-11-20
Migration Settings page goes blank, highlights wrong menu selection in Vertical Navigation


kedark3:
https://bugzilla.redhat.com/show_bug.cgi?id=1651403

vconzola:
@mturley @michaelkro @mzazrivec Think Milan's PR will fix at least the first part of this, right? What about the second part?

mzazrivec:
That 5.10.0.24 build doesn't include my changes from https://github.com/ManageIQ/manageiq-v2v/pull/769 Correct?

mturley:
I can confirm locally that both parts of this issue were resolved by #769. The second part (input field going blank) might have been something related to switching views with ConnectedRouter and re-rendering the settings page without a change to the stored settings in Redux. It does reliably work on a fresh page reload, though, and after #769 we do a full page reload on every nav change, so that fixes the input going blank as a result.

@mzazrivec, after we switch back to ConnectedRouter, the input going blank here might become an issue again.

kedark3:
![image](https://user-images.githubusercontent.com/8180469/48793060-cdc7c780-ecc3-11e8-8e13-50c387b02192.png)
I think we still have issue. tested in the build master.20181116223021_0f85990

vconzola:
Nav being highlighted wrong is fixed by Milan's first PR. The second part, we need to validate this is still fixed once Milan's new PR gets merged. @mturley to validate.

mturley:
This bug is not present in `hammer`, it was hidden by #769 as indicated above. Milan's new PR #822 re-exposes this bug, but we are not merging that to `hammer`. I have a new PR #839 to fix the bug in `master`.

PR #822, opened 2018-12-07
Phase 2 of vertical menu fixes

mzazrivec:
This PR:
* restores `ConnectedRouter` and hash routing: this is to have swift & fully client-side routing with browsing history
* adds code to re-style vertical menu whenever we navigate between menu items (or screens that would ordinarily fall into different menu item): this is so that correct vertical menu item is highlighted
* adds code to do `POST /migration/menu_item_url`: this is to update `session[:tab_url]` when navigating between two menu items so that clicking on `Compute ➛ Migrate` takes you to previously visited url

Fixes: https://github.com/ManageIQ/manageiq-v2v/issues/811

Depends on:
https://github.com/ManageIQ/manageiq-ui-classic/pull/5063
https://github.com/ManageIQ/manageiq-ui-classic/pull/4999

### DIFF
--- a/app/javascript/common/menu.js
+++ b/app/javascript/common/menu.js
@@ -1,0 +1,33 @@
+import http from './http';
+
+const updateUrl = '/migration/menu_section_url?section=migration&url=/migration%23';
+const activeMenuItemSelector = '[data-target="#menu-migration"] > div > ul > li[class="list-group-item active"]';
+const plansMenuItemId = 'menu_item_plans';
+const mappingsMenuItemId = 'menu_item_mappings';
+const settingsMenuItemId = 'menu_item_settings';
+const pathToMenuItem = {
+  plan: plansMenuItemId,
+  plans: plansMenuItemId,
+  mappings: mappingsMenuItemId,
+  settings: settingsMenuItemId
+};
+
+export const updateVerticalMenu = path => {
+  // POST /migration/menu_section_url to update session[:tab_url] on server
+  http.post(updateUrl + path);
+
+  // Remove 'active' class from current active menu item
+  const activeMenuItem = document.querySelector(activeMenuItemSelector);
+  if (activeMenuItem !== null) {
+    activeMenuItem.classList.remove('active');
+  }
+
+  // Based on location.pathname, set menu item id
+  const menu_item_id = pathToMenuItem[path.split('/')[1]];
+
+  // Add 'active' class to menu_item_id
+  const newActiveMenuItem = document.querySelector(`[id="${menu_item_id}"]`);
+  if (newActiveMenuItem !== null) {
+    newActiveMenuItem.classList.add('active');
+  }
+};
--- a/app/javascript/common/menu.js
+++ b/app/javascript/common/menu.js
@@ -15,12 +15,10 @@ export const updateVerticalMenu = path => {
   }
 
   // Based on location.pathname, set menu item id
-  const link_item = links.find(element => {
-    return element['path'].split('/')[0] === path.split('/')[1];
-  });
+  const link_item = links.find(element => element.path.split('/')[0] === path.split('/')[1]);
 
   // Add 'active' class to correct menu item
-  const newActiveMenuItem = document.querySelector(`[id="${link_item['menu_item_id']}"]`);
+  const newActiveMenuItem = document.querySelector(`[id="${link_item.menu_item_id}"]`);
   if (newActiveMenuItem !== null) {
     newActiveMenuItem.classList.add('active');
   }
--- a/app/javascript/common/menu.js
+++ b/app/javascript/common/menu.js
@@ -1,16 +1,8 @@
 import http from './http';
+import { links } from '../react/config/config';
 
 const updateUrl = '/migration/menu_section_url?section=migration&url=/migration%23';
 const activeMenuItemSelector = '[data-target="#menu-migration"] > div > ul > li[class="list-group-item active"]';
-const plansMenuItemId = 'menu_item_plans';
-const mappingsMenuItemId = 'menu_item_mappings';
-const settingsMenuItemId = 'menu_item_settings';
-const pathToMenuItem = {
-  plan: plansMenuItemId,
-  plans: plansMenuItemId,
-  mappings: mappingsMenuItemId,
-  settings: settingsMenuItemId
-};
 
 export const updateVerticalMenu = path => {
   // POST /migration/menu_section_url to update session[:tab_url] on server
@@ -23,10 +15,12 @@ export const updateVerticalMenu = path => {
   }
 
   // Based on location.pathname, set menu item id
-  const menu_item_id = pathToMenuItem[path.split('/')[1]];
+  const link_item = links.find(element => {
+    return element['path'].split('/')[0] === path.split('/')[1];
+  });
 
-  // Add 'active' class to menu_item_id
-  const newActiveMenuItem = document.querySelector(`[id="${menu_item_id}"]`);
+  // Add 'active' class to correct menu item
+  const newActiveMenuItem = document.querySelector(`[id="${link_item['menu_item_id']}"]`);
   if (newActiveMenuItem !== null) {
     newActiveMenuItem.classList.add('active');
   }
--- a/app/controllers/migration_controller.rb
+++ b/app/controllers/migration_controller.rb
@@ -1,18 +1,13 @@
 # Migration controller - mostly service to initialize the react top level component
 # and to handle the initial routing endpoint.
 class MigrationController < ApplicationController
+  include Mixins::MenuUpdateMixin
+
   before_action :check_privileges
   after_action :cleanup_action
 
   def index
-    @layout = case request.path
-              when '/migration/plans', /^\/migration\/plan\/.*/
-                'plans'
-              when '/migration/mappings'
-                'mappings'
-              when '/migration/settings'
-                'settings'
-              end
+    @layout = 'migration'
     @page_title = _('Migration')
   end
 
--- a/app/javascript/react/config/Routes.js
+++ b/app/javascript/react/config/Routes.js
@@ -1,12 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Route } from 'react-router-dom';
+import { Redirect, Route } from 'react-router-dom';
 import { links } from './config';
 import { componentSettings } from '../../components';
 import componentRegistry from '../../components/componentRegistry';
 
 const Routes = ({ store }) =>
-  links.map(({ path, component }) => {
+  links.map(({ path, component, redirect }) => {
+    if (typeof redirect !== 'undefined') {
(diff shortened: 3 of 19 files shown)


mzazrivec:
@michaelkro @mturley @himdel @Hyperkid123 

What I currently have in place does `ManageIQ.redux.history.listen( location =>  {})` inside `componentDidMount()` of the top-level `App` class: i.e. whenever route changes, my callback gets executed.

The problem with this implementation is that it doesn't work when you initially land on any of the migration screens: for example new login, then navigate to `Compute ➛ Migration ➛ Migration Plans`. You'll see that my callback did not run. Once you're in v2v and click on vertical menu item, it will be executed.

Any idea how to remedy this problem? Thanks.

Hyperkid123:
@mzazrivec The reason why is it now working is because the componentDidMount is called after render. The history is changed before that, so you can't catch the history change. Putting it into the constructor would work.

However using BrowserRouter is better imo. It should fix the menu on its own (because you have the default route without any fragment). There is no point in using hashRouter now. It was used because there were some issues with setting state on unmounted components. According to @himdel that should be fixed now.

Also you are setting up listener for history, but you should unsubscribe from it to avoid potential bugs:
```javascript
// somewhere in component
this.subscription = ManageIQ.redux.history.listen(...)

componentWillUnmnout() {
  // stops listening
  this.subscription()
}
```

Hyperkid123:
@mzazrivec yep the history changes is faster than the component...

himdel:
@mzazrivec The point was, the change in route never happens in those cases because the router gets loaded *after* the route changes. So anything to do with listening for changes to history will fail.

Looks like you may be able to use `ManageIQ.redux.store.getState().router.location` instead.

mzazrivec:
> ManageIQ.redux.store.getState().router.location

@himdel You're right, it works in constructor as well.

mzazrivec:
The problem with initial v2v load is resolved now.

app/javascript/common/menu.js
```diff
@@ -0,0 +1,33 @@
+import http from './http';
+
+const updateUrl = '/migration/menu_section_url?section=migration&url=/migration%23';
+const activeMenuItemSelector = '[data-target="#menu-migration"] > div > ul > li[class="list-group-item active"]';
+const plansMenuItemId = 'menu_item_plans';
+const mappingsMenuItemId = 'menu_item_mappings';
+const settingsMenuItemId = 'menu_item_settings';
+const pathToMenuItem = {
```
mturley: I wonder if we could have this path-to-menu-id mapping driven by the metadata in `app/javascript/react/config/config.js`, maybe by adding a `menuItemId` to each element in the `links` array there and doing something like `links.find(...).menuItemId` to find the matching id. That way we wouldn't have to edit code in two places if we want to change these routes.

If that's too much of a pain, I'm fine with it as-is. Just a thought.

mzazrivec: Yeah, that could be done too I guess.

mturley:
This PR re-exposes the settings form lifecycle bug reported in #802, which was hidden by the first phase of nav fixes, but never actually fixed. (the bug happens when you mount the settings page a second time with the same redux store after navigating away from it, which became impossible when we switched to BrowserRouter). I fixed the underlying bug in #839, which we should merge along with this.

mzazrivec:
@mturley This is what happens in my devel env when I click on the `Migration` parent menu item:

```
INFO -- : Started GET "/dashboard/maintab/?tab=migration" for ::1 at 2018-12-18 16:24:21 +0100
INFO -- : Processing by DashboardController#maintab as HTML
INFO -- :   Parameters: {"tab"=>"migration"}
INFO -- : Redirected to http://localhost:3000/migration#/plans

INFO -- : Completed 302 Found in 14ms (ActiveRecord: 1.9ms)

INFO -- : Started GET "/migration" for ::1 at 2018-12-18 16:24:22 +0100
INFO -- : Processing by MigrationController#index as HTML
INFO -- :   Rendering miq_v2v_ui_plugin/app/views/migration/index.html.haml within layouts/application
INFO -- :   Rendered miq_v2v_ui_plugin/app/views/migration/index.html.haml within layouts/application (1.4ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_doctype.html.haml (0.0ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/stylesheets/_template50.html.haml (0.0ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_i18n_js.html.haml (0.1ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_user_options.html.haml (2.1ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_notifications_drawer.html.haml (0.1ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_toast_list.html.haml (0.0ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_spinner.html.haml (0.0ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_lightbox_panel.html.haml (0.1ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_header.html.haml (10.1ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_vertical_navbar.html.haml (10.2ms)
INFO -- :   Rendered miq_v2v_ui_plugin/app/views/layouts/_full_center_v2v.html.haml (6.2ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_content.html.haml (18.4ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_adv_search.html.haml (0.0ms)
INFO -- :   Rendered manageiq-ui-classic/app/views/layouts/_footer.html.haml (0.8ms)
Session:         Hash of Size 4862, Elements 18
=================================
INFO -- : Completed 200 OK in 524ms (Views: 512.2ms | ActiveRecord: 1.9ms)

```

Also, from `config/routes.rb`:
```ruby
# ManageIQ::V2V::Engine.routes.draw do
Rails.application.routes.draw do
  match '/migration' => 'migration#index', :via => [:get]
  match 'migration/*page' => 'migration#index', :via => [:get]
...
```

I'm not sure why it wouldn't work on your end, but it seems to be working on my end (shrug).

mturley:
@mzazrivec, to be more specific it renders the ManageIQ nav, but the v2v React root renders nothing. I'll see if I can figure out why on my end. Thanks for looking into that.

mzazrivec:
@mturley I added (as you suggested):
* ability to configure menu item for each of our containers in `react/config/config.js`
* ability to configure redirection in `react/config/config.js` + configuration for empty path redirecting to `plans`

app/javascript/common/menu.js
```diff
@@ -0,0 +1,25 @@
...
+  const activeMenuItem = document.querySelector(activeMenuItemSelector);
+  if (activeMenuItem !== null) {
+    activeMenuItem.classList.remove('active');
+  }
+
+  // Based on location.pathname, set menu item id
+  const link_item = links.find(element => element.path.split('/')[0] === path.split('/')[1]);
+
+  // Add 'active' class to correct menu item
+  const newActiveMenuItem = document.querySelector(`[id="${link_item.menuItemId}"]`);
```
mturley: I think because you've changed `menu_item_id` to `menuItemId` here, it also needs to be changed in config.js, right?

mzazrivec: Nah, it's my mistake. Fixed.

bthurber:
https://bugzilla.redhat.com/show_bug.cgi?id=1666359

simaishi:
Hammer backport details:

```
$ git log -1
commit fff0084d4db70a595d9e3df4e1f6c1637c36b1e8
Date:   Mon Jan 7 13:40:46 2019 -0500

    Merge pull request #822 from mzazrivec/phase_2_of_vertical_menu_fixes
    
    Phase 2 of vertical menu fixes
    
    (cherry picked from commit 40a7a5f41a049b0bdef3e60a12d5f691810f193f)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1693743
```

app/javascript/common/menu.js
```diff
@@ -0,0 +1,25 @@
+import http from './http';
+import { links } from '../react/config/config';
+
+const updateUrl = '/migration/menu_section_url?section=migration&url=/migration%23';
+const activeMenuItemSelector = '[data-target="#menu-migration"] > div > ul > li[class="list-group-item active"]';
+
+export const updateVerticalMenu = path => {
+  // POST /migration/menu_section_url to update session[:tab_url] on server
+  http.post(updateUrl + path);
```
himdel: Looks like this is causing 

```
Server returned a non-200 response: undefined undefined
SyntaxError: Unexpected token h in JSON at position 1
angular.js:14961 TypeError: Cannot read property 'get' of undefined
    at ErrorModalController.$ctrl.show (error_modal_controller.self-...
```

because the `menu_section_url` method returns javascript, instead of json...

Except it shouldn't return anything, and the JS is the error handler code...

```js
$("#center_div").html("......Error text:\n\u003c/h2\u003e\n\u003cbr\u003e\n\u003ch3\u003e\nundefined method `parent_path\u0026#39; for nil:NilClass [migration/menu_section_url]...")
```

So, looks like something is not finding `parent_path` now, during `POST /migration/menu_section_url?section=migration&url=/migration%23/plans`

mzazrivec: For some reason I cannot reproduce this problem and a plain POST to that URL you showed above works correctly on my end.

himdel: I think I know .. this only happens *after* rails autoreload, when it forgets it has any plugin menu sections

mzazrivec: Right. But then the question is whether we're actually discussing the problem in the right PR (since I don't know what we can possibly do about it in a plugin PR).

himdel: Agreed, and it's not happening in production so maybe we don't have to treat it as a bug :).

(We could just make it not do anything when section is nil, but then we're just hiding errors, so idk, wdyt?)

mzazrivec: This whole thing should go away once we have a nice react-based menu. So I'd say we shouldn't do anything about it right now and just remove this logic once new menu lands.

himdel: We already have the react menu...

mzazrivec: Anyway, it seems that with `updateVerticalMenu()` in place, the migration menu items stay highlighted (i.e. even when I move between migration screens). And withouth `updateVerticalMenu()` (i.e, without the POST) the menu item don't get highlighted at all.

himdel: Ah, yes, `updateVerticalMenu` can't just set classes and expect it to work, it will have to change the react state instead of touching the html elements.

cc @Hyperkid123 , @fhlavac - do we have any JS way of telling the menu to change the active item?

Hyperkid123: No, but we can use redux for it.

PR #834, opened 2018-12-12
Add tests for OverviewReducer

mturley:


app/javascript/react/screens/App/Overview/overview.playbooks.fixtures.js
```diff
@@ -0,0 +1,20 @@
+import Immutable from 'seamless-immutable';
+
+export const playbooks = Immutable({
```
michaelkro: Since we're adding a fixtures file for playboosk, can we remove the one I added a while back?  It's not being used anywhere

https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardAdvancedOptionsStep/PlanWizardAdvancedOptionsStep.fixtures.js

mturley: Whoops! didn't notice that. Thanks.

PR #841, opened 2018-12-19
[#835] Edit Plan: fetch correct status of pre-selected VMs using VM validation service

### LINKED ISSUE #835: Editing a plan with already migrated VMs shows incorrect status in discovery mode
https://bugzilla.redhat.com/show_bug.cgi?id=1654865

> The reason this is happening is because of how we handle populating the list of VMs when editing a migration plan. The VM discovery service does not return VMs which are already part of a plan, because they must be excluded when creating new plans. As a result, when editing a plan and using VM discovery mode, in order to include the VMs that should be pre-selected in that list, we must merge together the discovery results with a list of the VMs already in the plan.

> In doing this merge, we have no status information about the VMs being merged in (because that VM data is not coming from the discovery service), so we display "available for migration", which is technically correct but just inconsistent with what you get when you import a CSV containing those VMs and you see the "already migrated" status. It's only possible to see this corner case by creating a plan which includes VMs which have already been migrated (only possible via a CSV) and then editing that plan and choosing discovery mode.

**Solution: Figure out a way to query the discovery service twice in this scenario, once to discover new VMs and once with a list of the VMs already existing in the plan (like we do for CSVs currently), and then merge those results so we have correct status information when merging.**

Not a blocker for 1.1, we'll do this for a z stream release.

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1654865
Fixes #835.

In order to get the right status info, rather than using the VM API query results to populate the VM list, I took the list of VM names from those query results and ran them through the VM validation service like we do with CSV imports, then combining those results with the initial validation (discovery) request. This includes changes to the reducer so it can update state properly when two concurrent validation requests are made.

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
@@ -139,7 +139,6 @@ class PlanWizardVMStep extends React.Component {
         validVms.filter(vm => vm.valid === false)
       );
       const conflictVms = Immutable.asMutable(conflict_vms, { deep: true });
-      // TODO handle removal of everything related to preselected_vms and isQueryingVms?
       const combined = [...inValidsVms, ...conflictVms, ...validVms];
       if (combined.length) {
         let initialSelectedRows = formSelectedVms || [];
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
@@ -12,13 +12,14 @@ import { overwriteCsvConfirmModalProps } from '../../PlanWizardConstants';
 class PlanWizardVMStep extends React.Component {
   componentDidMount() {
     const { vm_choice_radio, editingPlan, shouldPrefillForEditing, queryPreselectedVmsAction, pristine } = this.props;
-    if (pristine) {
-      if (vm_choice_radio === 'vms_via_discovery') {
-        this.validateVms();
-      }
+    if (pristine && vm_choice_radio === 'vms_via_discovery') {
+      this.validateVms([], null, { combineRequests: !!shouldPrefillForEditing });
       if (shouldPrefillForEditing) {
         const vmIds = getVmIds(editingPlan);
-        queryPreselectedVmsAction(vmIds);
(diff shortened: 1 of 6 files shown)


mturley:
Added some unit tests, this one should be ready to go now.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js
```diff
@@ -44,9 +45,9 @@ class PlanWizardVMStep extends React.Component {
     const { csvParseErrorAction } = this.props;
     csvParseErrorAction(errMsg);
   };
-  validateVms = () => {
+  validateVms = (vms = [], planId = null, meta = {}) => {
```
michaelkro: I think it doesn't matter since you've provided defaults, but since you're changing the function signature, should we also update [this line?](https://github.com/ManageIQ/manageiq-v2v/blob/efe628b8d5aea578a0cb74c7da49e422925e0edc/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardVMStep/PlanWizardVMStep.js#L29)

mturley: We can if you think it would be best, but that's why I provided defaults. No arguments = discovery of VMs with no prior information. I'm ok with either way. Would that be your preference?

michaelkro: I was just sitting here and pondering this.  It's not a big deal, just a functionality vs. convention thing lol

When it comes to assigning default values to function parameters, I think of it as a way to guard against inadvertently passing `undefined`.  On the other hand, if I want to have truly optional arguments, I like to use an object.  This allows us to avoid having to pass `null`

```js
this.validateVms([], null, { combineRequests: !!shouldPrefillForEditing })
```

Anyway, either way works, and since we've never really set a convention one way or another, I should not be so picky about it.  I think we're good to go :)

simaishi:
Hammer backport details:

```
$ git log -1
commit d21a140c254da4f9b863f865dbe300937da79db3
Date:   Tue Jan 8 08:27:00 2019 -0500

    Merge pull request #841 from mturley/835-edit-plan-vm-status
    
    [#835] Edit Plan: fetch correct status of pre-selected VMs using VM validation service
    
    (cherry picked from commit da7b93e10348b5a534fd1b4f302d50e9848b511f)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1696419
```

Issue #385, opened 2018-06-01
Be more explicit regarding the format of the file to be imported in the empty state


serenamarie125:
Re: comments in #379 
> Import a file including a list of VMs to be migrated

Can we be more explicit regarding the format of the file?  CSV, etc ?

mturley:
Yeah @serenamarie125 , I think I asked at one point when nobody was listening, and I ended up just going with that text from the mockup. I would be cool with something like:

> Import a CSV file with VMs to be migrated (i)

and then when you hover/click the (i) info icon, a little popover maybe that describes the format/columns required?

Maybe not an info icon at the end, it could instead be a hover over the words "CSV file", I see dotted underlines with hover info tips occasionally.

mturley:
or maybe

> Import a list of VMs to be migrated in CSV format (i)

serenamarie125:
I’ve actually requested our content strategist to do a text review of all our screens :)   Waiting to hear back on her availability

mturley:
Cool!

mturley:
@serenamarie125 Another thing that might be useful, when you have a CSV import option, is to have an example CSV they can download as a template. Even if it's just an empty sheet with header row only, they can drop their info into it and upload. They had that on minted.com when i did my wedding guest address CSV

vconzola:
@mturley I talked about the template idea with Brett and Marco a while ago and they liked the idea. It's just a matter of getting it prioritized, I think. Wrt having a popover describing the the format of the CSV file, I didn't think Patternfly was down with field level help, but I might be wrong. @serenamarie125 can verify I'm sure. Also, just FYI, v2v finally has a writer, so if Stacy doesn't have time to do a UI review perhaps we can run specific questions past the writer.

mturley:
@vconzola looping back on this, since it's low priority but I didn't want to lose track of it. I think the problem with field level help for the file format is, that might be a bit of a small space to effectively describe in enough detail what is needed. It would have to be something like "A CSV file with columns for x, y z", but that doesn't explain if there needs to be a header row, or whether the delimiters matter, etc.

I think a modal that pops up a simple example of the CSV you should create, and offers to let you download an empty one with header rows or something, would be perfect here.

This would be a perfect task for a writer: how can we explain what format to make the uploaded CSV file without it being confusing and convoluted?  Maybe just the example/template file should speak for itself instead?

vconzola:
[...]

PR #845, opened 2019-01-04
[#842] Add "Flavor" and "Security Group" CSV columns and associated warnings

### LINKED ISSUE #842: Plan Wizard: Add support for OSP flavor and security group columns in CSV import
When uploading a CSV file containing VMs to migrate in the Plan Wizard, users should be able to pre-select the OSP security group and flavor to be used for each VM (currently only available as dropdown menus on the Instance Properties step) by including optional CSV columns labeled "Security Group" and "Flavor".

If a group or flavor name specified in the CSV matches one of the available options, the dropdown menu should have that option selected upon entering the Instance Properties step.

If a group or flavor name specified in the CSV **does not** match one of the available options, the existing default (best fit) should be used, and the user should be warned in some way that their CSV value was not used.

mturley:
Closes #842.

# Testing

Use the `ospv2vbos` database.
To see the sunny-day case, create a new migration plan using the following CSV:
```
Name,Security Group,Flavor
Database-01,dbservers,x1.large
fdupont-test-migration,appservers,x1.xtra-large
```
Proceed to the Instance Properties step and you should see that the referenced groups and flavors have been pre-selected.

To see the warning case, create a new migration plan using the following CSV:
```
Name,Security Group,Flavor
Database-01,invalidgroup,x1.large
fdupont-test-migration,appservers,invalidflavor
```
Proceed to the Instance Properties step and you should see that the valid flavors/groups have been selected, but the invalid ones have reset to their defaults and displayed warnings (see screenshots below). If you click the Edit icon on each row and then click Confirm (whether you've made changes or not), the warnings will be dismissed (see recording below).

# Screens

screenshot 2019-01-04 16 17 29
screenshot 2019-01-04 16 17 11
screenshot 2019-01-04 16 17 17

# Recording

pz6cv2nyom

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
@@ -109,17 +109,23 @@ export default (state = initialState, action) => {
           }
         }
 
-        if (editingPlan && !preselectedFlavor && !preselectedGroup) {
+        if (editingPlan) {
           const existingVm = editingPlan.options.config_info.actions.find(
             vm => vm.vm_id === existingInstancePropertiesRow.id
           );
           if (existingVm) {
-            const existingFlavor =
-              tenantFlavors && tenantFlavors.find(flavor => flavor.id === existingVm.osp_flavor_id);
-            const existingGroup =
-              tenantSecurityGroups && tenantSecurityGroups.find(group => group.id === existingVm.osp_security_group_id);
-            preselectedFlavor = existingFlavor && { name: existingFlavor.name, id: existingFlavor.id };
-            preselectedGroup = existingGroup && { name: existingGroup.name, id: existingGroup.id };
+            const { csvFields } = existingInstancePropertiesRow;
+            if (!csvFields || !csvFields.osp_flavor) {
+              const existingFlavor =
+                tenantFlavors && tenantFlavors.find(flavor => flavor.id === existingVm.osp_flavor_id);
+              preselectedFlavor = existingFlavor && { name: existingFlavor.name, id: existingFlavor.id };
+            }
+            if (!csvFields || !csvFields.osp_security_group) {
+              const existingGroup =
+                tenantSecurityGroups &&
+                tenantSecurityGroups.find(group => group.id === existingVm.osp_security_group_id);
+              preselectedGroup = existingGroup && { name: existingGroup.name, id: existingGroup.id };
+            }
           }
         }
 
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
@@ -92,14 +92,20 @@ export default (state = initialState, action) => {
         let preselectedFlavor;
         let preselectedGroup;
 
-        /*
-        if (csv specified flavors and groups) {
-          preselectedFlavor = what's in the csv
-          preselectedGroup = what's in the csv
-        } else if (editingPlan) { ... }
-        */
+        if (existingInstancePropertiesRow.csvFields) {
+          const { osp_flavor, osp_security_group } = existingInstancePropertiesRow.csvFields;
+          if (osp_flavor) {
+            const matchingFlavor = tenantFlavors && tenantFlavors.find(flavor => flavor.name === osp_flavor);
+            preselectedFlavor = matchingFlavor && { name: matchingFlavor.name, id: matchingFlavor.id };
+          }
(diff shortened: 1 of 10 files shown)


michaelkro:
Am reviewing (so far looks good!) and it seems that the error messaging for a malformed csv file points the user to our product documentation.

I'm guessing we should alert Avital or Steve about this enhancement once it's merged @vconzola

vconzola:
@mikero Not sure what you mean by error messaging pointing to documentation? Which error message?

mturley:
@vconzola if there is an error parsing the CSV, we display "The selected file does not have the expected format. [error details]  See product documentation for the correct format of the .csv file".

I agree with Mike, since we're adding new optional columns to the CSV we should make sure that gets documented. Also, we should maybe put some reference to the documentation in the view before you upload a CSV, there's currently no indication of what that format should be. (https://github.com/ManageIQ/manageiq-v2v/issues/385)

mturley:
Example of the error messaging we have:
screenshot 2019-01-09 09 52 10

mturley:
Hmm, ok. Do we think we can reasonably expect users to know the id of a security group in CloudForms? If security group names aren't unique, do we need to also include security group ids in the existing dropdown that only shows names?

michaelkro:
I think we need to do some investigation into whether or not duplicate security group naming is a common practice for OSP, and what an ideal solution would be.  Let's split that off into a separate issue

michaelkro:
For the sake of documentation, posting here.

>  The security group names are unique at the tenant level though. So if you filter by tenant, you should be good.

We scope to the target tenant when fetching security groups, so my concern with regard to duplicate names is a non-factor in our use case.

app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/PlanWizardInstancePropertiesStepReducer.js
```diff
@@ -90,9 +90,26 @@ export default (state = initialState, action) => {
...
+          }
+          if (osp_security_group) {
+            const matchingGroup =
+              tenantSecurityGroups && tenantSecurityGroups.find(group => group.name === osp_security_group);
+            preselectedGroup = matchingGroup && { name: matchingGroup.name, id: matchingGroup.id };
+            csvInvalidGroupWarning = !matchingGroup;
+          }
+        }
+
+        if (editingPlan && !preselectedFlavor && !preselectedGroup) {
```
michaelkro: Bit of an edge case here with regard to this conditional

1. Create a plan using CSV.  Specify security groups and flavors
2. Remove either the flavor column from the CSV, and invalidate one of the security group entries (possible use case:  user decides to edit plan to use best fit flavors, but still use predefined security groups.  Accidentally messes up a security group name during edit)
3. On the Instance Properties Step, the row with the invalid security group will correctly show the new warning, but will load the previously saved security group instead of falling back to the default, contrary to the warning messaging.

vconzola:
Thanks for looking into this @michaelkro . Wrt the edge case you describe, not sure I understand. I don't think we ever consider the case where a CSV file is changed once a plan is created.  As far as I know, it shouldn't matter. Once a plan is created we never refer to the CSV again, do we?

michaelkro:
@vconzola Just trying to be thorough :) Technically, the user doesn't have to reuse a CSV to recreate the corner case I describe, in fact, it's more likely to run into the issue when editing with a new CSV (more likely to introduce typos)

In any case, this is just me poking at the code in very specific ways to see what shakes loose.  If we're not worried about this scenario, I'll let it go

vconzola:
@michaelkro So in your use case above, step 3 is editing the plan created in step 1?  Just trying to understand the use case.

mturley:
@vconzola that's correct. The edge case is concerning what to do when you're editing a plan, uploading a new CSV, and the new CSV values for flavor or group are invalid. The question is what flavor/group values to fall back on: the ones previously saved in the plan being edited, or the default/best fit (as if this new CSV was used for a fresh plan, and as indicated by the warning message).

Initially my PR behaved like the former, Mike and I talked and determined it should behave like the latter, so I made those changes.

vconzola:
@mturley  Got it. Understand now. Thanks!

simaishi:
@mturley Backport of this PR to hammer branch conflicts because https://github.com/ManageIQ/manageiq-v2v/pull/841 isn't in hammer branch. Are you able to create a separate PR for hammer without PR841, or do we need to take PR841 (in which case an approval is needed).

cc @fdupont-redhat

mturley:
@simaishi it would be better if we can backport #841, I'm not sure why it wasn't approved (https://bugzilla.redhat.com/show_bug.cgi?id=1654865). I could make a hammer PR for this but then we'll end up with a merge conflict when we backport the other PR. I'll make a comment in that BZ.

simaishi:
Hammer backport details:

```
$ git log -1
commit eae033d693d67e4b8adae898ff802e9571e215c1
Date:   Fri Jan 11 08:04:24 2019 -0500

    Merge pull request #845 from mturley/842-csv-groups-flavors
    
    [#842] Add "Flavor" and "Security Group" CSV columns and associated warnings
    
    (cherry picked from commit d00b293a46182df2c772ef34a00581826c05dc23)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693714
```

PR #847, opened 2019-01-08
[#840] Use conversion_hosts API for RHV target cluster warnings instead of host CF tags

### LINKED ISSUE #840: Mapping Wizard Clusters Step: RHV conversion hosts warning depends on CF tags, should use /api/conversion_hosts instead
According to Brett, the CFME tags for a cluster having a conversion host enabled are no longer required after v1.1. We display the "you must have conversion hosts" warning if those tags are not present, so without this change we may give a false warning to users in v1.2+ who have conversion hosts configured but do not have these tags.

For the OSP warning, we used /api/conversion_hosts, we should use that for RHV as well. We'll need to associate the objects returned from that API call with the target clusters in order to prevent the warnings in the right clusters.

mturley:
Closes #840.

Because we were previously using the `/api/hosts` query for this warning only, it's no longer in use, so I removed it and cleaned up the relevant obsolete code. 🛀


vconzola:
BZ for this change: https://bugzilla.redhat.com/show_bug.cgi?id=1667097

mturley:
@simaishi we'll probably need to backport this one before https://github.com/ManageIQ/manageiq-v2v/pull/921, because it changes the same line (this PR changes it from a named property to a computed property, and #921 changes the value of the property).

I'm actually a little bit concerned about how many things we need to backport to hammer that may depend on each other, because there was some refactoring involved in some of them. As we're preparing the 5.10.3 release we might want to make sure we leave enough time to untangle that.

simaishi:
> I'm actually a little bit concerned about how many things we need to backport to hammer that may depend on each other,

So far, every single one of the PRs I tried to backport conflict.... Some dependent PRs aren't targeted for upcoming release and I'm not backporting them, so that makes it even more difficult... Trying to figure out if there is a better way to do this...

simaishi:
Hammer backport details:

```
$ git log -1
commit 94b8c06af324d16431f5deacb29fc3901baa00a7
Date:   Fri Jan 11 08:46:55 2019 -0500

    Merge pull request #847 from mturley/840-rhv-conversion-host-warnings
    
    [#840] Use conversion_hosts API for RHV target cluster warnings instead of host CF tags
    
    (cherry picked from commit 641ae0b0fb712866f9fa581b9c8e16bd91fcd94c)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693745
```

PR #851, opened 2019-01-17
Add routed tabs to the Settings page to prepare for Conversion Hosts settings

mturley:
The old contents of `Settings.js` are in `GeneralSettings.js` now, and the new `Settings.js` is just a breadcrumb bar and tab container. The diff is a bit noisy because of this code being moved around.

You can see the tabs in action by turning `hideConversionHostSettings` to false (https://github.com/ManageIQ/manageiq-v2v/pull/851/files#diff-6c1adc83a5ff4719fa9ef50664bcf357R12), this flag is included so that we can merge this without exposing the unfinished ConversionHostsSettings component to master should we need to make a release.

u8guynzqss


app/javascript/react/screens/App/Settings/Settings.js
```diff
@@ -1,82 +1,84 @@
...
+      <Breadcrumb.Item href="/dashboard/maintab?tab=compute">{__('Compute')}</Breadcrumb.Item>
+      <Breadcrumb.Item href="#/plans">{__('Migration')}</Breadcrumb.Item>
+      <Breadcrumb.Item active>{__('Migration Settings')}</Breadcrumb.Item>
+    </Toolbar>
+  );
 
-  render() {
-    const { isFetchingServers, isFetchingSettings, isSavingSettings, savedSettings, settingsForm } = this.props;
+  const generalSettings = (
+    <GeneralSettings
```
michaelkro: Wondering if it would make sense, architecturally, to create a container component for GeneralSettings instead of passing these props down from Settings.

Also ok with putting this off until the actual need arises, but just want to avoid another <Overview /> situation, lol.

Do we anticipate any of these props to be shared across tabs?

mturley: That's a good point, I guess I wasn't sure if we would need to use the generalized /api/settings actions and state outside of that first tab. I think you're right, I'll connect each tab to redux separately.

app/javascript/react/config/config.js
```diff
@@ -14,6 +14,11 @@ export const links = [
     component: Settings,
     menu_item_id: 'menu_item_settings'
   },
+  {
+    path: 'settings/conversion_hosts',
+    component: Settings,
+    menu_item_id: 'menu_item_settings'
+  },
```
michaelkro: Just making sure...  we don't need an entry here for `settings/general` because by default, entering the `settings` route opens the general tab?

mturley: @michaelkro That's correct. There's no `/settings/general` route, `/settings` and `/settings/conversion_hosts` are the routes to each tab.

mturley:
We will need this to be backported for https://bugzilla.redhat.com/show_bug.cgi?id=1688951.

simaishi:
Hammer backport details:

```
$ git log -1
commit 75686ef4ca01d74f2c7af03c4c5bd2ef7d5d2b3b
Date:   Thu Jan 24 15:31:21 2019 -0500

    Merge pull request #851 from mturley/settings-tabs
    
    Add routed tabs to the Settings page to prepare for Conversion Hosts settings
    
    (cherry picked from commit 339f3bf847c95c693f45af5abc107a0369f912ca)
    
```

Issue #855, opened 2019-01-24
Conversion Host Enablement - Configuration Wizard


vconzola:
Create a wizard for configuring conversion hosts for RHV or OSP.

Mockups to follow.

Part of the conversion host enablement feature: https://bugzilla.redhat.com/show_bug.cgi?id=1622728

mturley:
@vconzola when you get a chance (if it's something you're ready to determine) the first thing we'll need here is simply a list of the names of each wizard step, so we have a basis for labels and filenames.

vconzola:
Mockups: 
[Conversion Host wizard.pdf](https://github.com/ManageIQ/manageiq-v2v/files/2797454/Conversion.Host.wizard.pdf)

vconzola:
@mturley I added a link to the mockups above, but don't consider those step names final.

mturley:
Closing this since the remaining parts of the wizard were moved to #898, #899, #903 and #881.

PR #871, opened 2019-01-31
Conversion Host Configuration Wizard Step #1 (Location), and introducing common reducers

mturley:
This is part of #855.

This PR adds the contents of the Location step (Step 1) of the Configure Conversion Host wizard:

9wx9i0enag

This PR also adds a check for sufficient providers before displaying the Conversion Hosts tab of the settings page, to prevent errors in this wizard step when a user has insufficient providers configured. In this case, the empty state is shown with a button for configuring providers, similar to the Plans and Mappings views:

screenshot 2019-02-04 13 00 46

Since this PR would have otherwise duplicated a bunch of code related to providers and target clusters/projects, I also introduced a new pattern of reducers coupled to API resources rather than to views. We have a good amount of this kind of code duplication elsewhere in the codebase, but that's outside the scope of this PR, so I only created the `providersReducer` and the `targetResourcesReducer` (which so far only handles target clusters). We can move things into common reducers like these incrementally, rather than duplicate code, as the need arises.


mturley:
This needs unit tests, but is otherwise ready to go.

@michaelkro, I'm adding you as a reviewer here since you were kind enough to offer, but since it's past 5pm on your last day, feel free to revoke that offer at your discretion! Once again it's been a real pleasure working with you. Cheers 🍻

app/javascript/react/screens/App/Settings/SettingsActions.js
```diff
@@ -61,6 +66,36 @@ export const fetchConversionHostsAction = url => {
...
+  });
+};
+
+const _getProvidersActionCreator = url => dispatch =>
+  dispatch({
+    type: FETCH_V2V_PROVIDERS,
+    payload: API.get(url)
+  });
+
+export const fetchProvidersAction = url => {
```
michaelkro: Food for thought...

There is already a system in place for fetching and maintaining a list of providers in the Overview screen.  Now that we need similar functionality here in Settings, we are duplicating that code.

When resources are shared across screens like this, it might make sense to have "application level" actions/reducers.  These could be made available to the screens/components that need them in the same way we're already doing it (map state to props in the container components)

A few benefits for this approach
1. Reduce code duplication
2. It would reduce the size of our reducers and make them easier to read.
3. Screen reducers would only need to concern themselves with the UI state and resource reducers would manage data

That being said, the work here is in line with the state management system that is already in place.  I just wanted to point this out

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostWizard/ConversionHostWizardLocationStep/ConversionHostWizardLocationStep.js
```diff
@@ -1,10 +1,127 @@
...
+  FETCH_TARGET_COMPUTE_URLS
+} from '../../../../../../../../../common/constants';
 
-export default ConversionHostWizardLocationStep;
+import { stepIDs, PROVIDER_TYPES } from '../ConversionHostWizardConstants';
+
+class ConversionHostWizardLocationStep extends React.Component {
+  componentDidMount() {
+    const { fetchProvidersAction, fetchProvidersUrl } = this.props;
+    fetchProvidersAction(fetchProvidersUrl);
```
michaelkro: Since we're fetching providers on mount, we could use this data to dynamically populate the provider type dropdown with only what is available.

Oh, that makes me think, what happens on the conversion hosts tab if there are no providers?  Do we show the provider empty state like in Mappings and Overview?

mturley: That's a really good point, I hadn't considered that. I think as implemented, it would just show the conversion hosts empty state and prompt the user to open the conversion host wizard, where the providers dropdown would be empty. That's a case I need to handle.

app/javascript/react/screens/App/Settings/SettingsActions.js
mturley: @michaelkro I did a major refactor to create the new `providersReducer` and `targetResourcesReducer` based on this advice. I'm curious what you think of this implementation.

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostWizard/ConversionHostWizardLocationStep/ConversionHostWizardLocationStep.js
mturley: Also handled this.

app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardClustersStep/index.js
```diff
@@ -2,6 +2,7 @@ import { connect } from 'react-redux';
 
 import MappingWizardClustersStep from './MappingWizardClustersStep';
 import * as MappingWizardClustersStepActions from './MappingWizardClustersStepActions';
+import * as TargetResourcesActions from '../../../../../../../../redux/common/targetResources/targetResourcesActions';
```
michaelkro: We probably don't want to import all the things as a general rule any more.  As we migrate more of the resource state management here, we'll only want to grab the appropriate actions.

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/ConversionHostsSettings.js
```diff
@@ -1,66 +1,85 @@
...
-                {__('Configure Conversion Host')}
-              </a>
+      <Spinner loading={isFetchingConversionHosts || isFetchingProviders} style={{ marginTop: 15 }}>
+        {!hasSufficientProviders ? (
+          <ShowWizardEmptyState
+            description={
+              __('There are no providers available from which to configure a conversion host. You must configure a target provider before configuring conversion hosts.') // prettier-ignore
+            }
+            buttonText={__('Configure Providers')}
+            buttonHref="/ems_infra/show_list"
```
michaelkro: I think this is ok for this PR, but just wanted to point out that clicking this button takes you to

compute -> infrastructure -> providers

which is where a user would need to go to configure VMW and RHV providers, but not for OSP.  Not sure what the best thing to do is, but just wanted to point it out.

mturley: Hmm, good catch. I'll open an issue for that, not sure how we should handle it.

mturley:
@michaelkro thanks so much! I really appreciate your feedback, and your help moving towards a nicer pattern for the common reducers.

Now go relax!

simaishi:
Hammer backport details:

```
$ git log -1
commit 5d8c17ae03a55c9c6316cddbd35f7cbc3cd8883c
Date:   Tue Feb 5 15:54:52 2019 -0500

    Merge pull request #871 from mturley/conv-host-wizard-step1
    
    Conversion Host Configuration Wizard Step #1 (Location), and introducing common reducers
    
    (cherry picked from commit a0ee080c47502453d76ab6d7f0629471438c184f)
    
```

simaishi:
@mturley Can you please review the bakcport of this PR and make necessary changes in a new PR? I skipped renaming of providers.fixtures file as the file doesn't exist in `hammer` branch. But it looks some other changes are needed because of that...

```
/home/travis/build/ManageIQ/manageiq-v2v/app/javascript/redux/common/providers/__tests__/providersActions.test.js
  5:27  error  Unable to resolve path to module '../providers.fixtures'  import/no-unresolved
```

The conflits backporting this PR were:

```
	both modified:   app/javascript/react/screens/App/Overview/OverviewReducer.js
	both modified:   app/javascript/react/screens/App/Overview/__test__/OverviewReducer.test.js
	deleted by us:   app/javascript/react/screens/App/Overview/__test__/__snapshots__/OverviewReducer.test.js.snap
	added by them:   app/javascript/redux/common/providers/providers.fixtures.js

```

mturley:
@simaishi sure, I'll take a look first thing in the morning.

mturley:
@simaishi here's a PR to fix that problem: https://github.com/ManageIQ/manageiq-v2v/pull/929

I reviewed the other changes in your backport commit, everything else looks good. Thanks again for dealing with this mess :)

Issue #860, opened 2019-01-28
Conversion Host Enablement - Listview #1 - Basic list view with placeholder items


mturley:
This is part 1 of 3 for #856.

This issue includes using the `ListViewToolbar` component to provide filtering/sorting/pagination controls to a basic ListView with just the conversion host name (and possibly also provider type?) in each item. We should be able to filter and sort by conversion host name (and provider type if we include that here).

Examples of `ListViewToolbar` usage can be found in most of our other list view components, like the PlanVmsList: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/components/PlanVmsList.js.  It uses [React's "render props" pattern](https://reactjs.org/docs/render-props.html) (in this case, passing a function as the children prop so that the ListViewToolbar can provide special props used by the children).

PR #874, opened 2019-02-05
Conversion Hosts: list view phase 1

mzazrivec:
This PR implements:
- basic list view for conversion hosts
- pagination
- filtering
- sorting

Currently, the list view displays only conversion host name. The issue linked below also mentions showing provider type (or perhaps provider name): for that we'd need more information to be pulled from API.

conversion-hosts-01

Fixes: https://github.com/ManageIQ/manageiq-v2v/issues/860

### DIFF
--- a/app/javascript/react/screens/App/Settings/Settings.scss
+++ b/app/javascript/react/screens/App/Settings/Settings.scss
@@ -1,3 +1,7 @@
 .migration-settings h2 {
   margin-bottom: 25px;
 }
+
+.conversion-hosts-list .list-view-pf-main-info {
+  padding: 10px 0;
+}
--- a/app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
+++ b/app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
@@ -1,17 +1,61 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Grid, ListView, Toolbar } from 'patternfly-react';
+import ListViewToolbar from '../../../../common/ListViewToolbar/ListViewToolbar';
 
 const ConversionHostsList = ({ conversionHosts }) => (
-  <React.Fragment>
-    <h2>TODO: render a list view here using the ListViewToolbar pattern we have elsewhere</h2>
-    {conversionHosts.map(conversionHost => (
-      <pre key={conversionHost.id}>{JSON.stringify(conversionHost)}</pre>
-    ))}
-  </React.Fragment>
+  <ListViewToolbar
+    filterTypes={ConversionHostsList.filterTypes}
+    sortFields={ConversionHostsList.sortFields}
+    listItems={conversionHosts}
+  >
+    {({
+      filteredSortedPaginatedListItems,
+      renderFilterControls,
+      renderSortControls,
+      renderActiveFilters,
+      renderPaginationRow
+    }) => (
+      <React.Fragment>
+        <Grid.Row>
+          <Toolbar>
+            {renderFilterControls()}
+            {renderSortControls()}
+            {renderActiveFilters(filteredSortedPaginatedListItems)}
+          </Toolbar>
+        </Grid.Row>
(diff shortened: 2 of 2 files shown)


app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
```diff
@@ -1,17 +1,61 @@
...
+    }) => (
+      <React.Fragment>
+        <Grid.Row>
+          <Toolbar>
+            {renderFilterControls()}
+            {renderSortControls()}
+            {renderActiveFilters(filteredSortedPaginatedListItems)}
+          </Toolbar>
+        </Grid.Row>
+        <div style={{ overflow: 'auto', paddingBottom: 300, height: '100%' }}>
```
mturley: This is outside the scope of this PR, because this pattern is all over the place, but I think at some point we should probably move all of these inline `style` props into CSS, it's a little messy that we use both. It's fine for now though.

app/javascript/react/screens/App/Settings/Settings.scss
```diff
@@ -1,3 +1,7 @@
 .migration-settings h2 {
   margin-bottom: 25px;
 }
+
+.conversion-hosts-list .list-view-pf-main-info {
```
mturley: This smaller padding is fine for now, but as we add more stuff to the list view items we'll probably want to remove or tweak this to make sure the layout matches the mockups and the patternfly standards. I think maybe once we have more in here, the thicker padding will make more sense.

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
```diff
@@ -1,17 +1,61 @@
...
+          <ListView className="conversion-hosts-list" id="conversion_hosts">
+            {filteredSortedPaginatedListItems.items.map((host, n) => (
+              <ListView.Item key={host.id} heading={host.name} stacked />
```
mturley: Once we add more details to the list view items, they should probably become their own `ConversionHostListItem` component that we import here.

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
```diff
@@ -1,17 +1,61 @@
...
+      </React.Fragment>
+    )}
+  </ListViewToolbar>
 );
 
 ConversionHostsList.propTypes = {
   conversionHosts: PropTypes.arrayOf(PropTypes.object)
 };
 
+ConversionHostsList.sortFields = [
```
mturley: I was about to comment that we should move these `sortFields` and `filterTypes` into a constants file, but I actually like them here as a property on the component. Unless we end up needing to reuse them somewhere else, this is a fine place for them.

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
mzazrivec: Yep, I definitelly agree with moving this into CSS (we use it in 3 other places as well).

app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostsList.js
mzazrivec: Yes, we actually use both approaches in v2v code, I'm fine with going either way with this, just perhaps use same pattern everywhere. What do you think?

simaishi:
Hammer backport details:

```
$ git log -1
commit 96dcc9ebe2c93272cd82d7bc831dfab17c11beb4
Date:   Tue Feb 5 18:33:44 2019 -0500

    Merge pull request #874 from mzazrivec/conversion_hosts_list_view_phase_1
    
    Conversion Hosts: list view phase 1
    
    (cherry picked from commit e379015fc4764ce0ab70e323313115e30443df67)
    
```

PR #876, opened 2019-02-07
Conversion Host Configuration Wizard - Step #2 - Host(s)

mturley:
This is part of #855.

For now, to get the TypeAheadSelect working, this PR introduces CSS duplicated from patternfly-react because we can't import that CSS directly due to https://github.com/patternfly/patternfly-react/issues/673.

This PR removes the `getDerivedProps` pattern I had introduced in the Location step, because it makes more sense to compute those props in the `mapStateToProps` function so the component doesn't have to care about how they are derived. The new Host(s) step follows the same pattern.

This PR adds the `vms` attribute to the `FETCH_TARGET_COMPUTE_URLS[OPENSTACK]` constant so that we can get the list of VMs for this step, which has the side effect of unnecessarily loading those VMs when the same constant is used in the mapping wizard. However I think the performance hit there is negligible, and coming up with a way to share these URLs with different sets of attributes seemed outside the scope of this PR. I think the ease of reuse here outweighs the drawbacks.

This PR also introduces the `TypeAheadSelectField` component, which is a thin wrapper around the `TypeAheadSelect` from patternfly-react, which makes it compatible to be used as the `component` prop of a redux-form `Field`. We can reuse this new component if we need typeahead-select anywhere else in the UI.

Another important detail to note is that the selected hosts will remain in place if you step back and forward through the wizard, unless the selected Project or Cluster is changed on the Location step. This change resets the Hosts step, so that we don't allow the user to select hosts from multiple projects/clusters. You can see this reset happening towards the end of the screen recording here:

54hzkvrnav

### DIFF
--- a/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
+++ b/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
@@ -10,6 +10,10 @@
   padding: 2px 6px;
 }
 
+.form-control.rbt-input-multi {
+  height: auto;
+}
+
 .rbt-menu.dropdown-menu {
   &.show {
     margin: 0;
--- a/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
+++ b/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
@@ -1,0 +1,61 @@
+/*
+  Until we can properly import patternfly-react CSS extensions
+  (which will probably require changes to manageiq-ui-classic's webpack config)
+  we need to include the styles for react-bootstrap-typeahead manually.
+
+  This can probably be removed after https://github.com/patternfly/patternfly-react/issues/673 is fixed.
+  This is a copy of https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/sass/patternfly-react/_type-ahead-select.scss
+*/
+
+@import 'react-bootstrap-typeahead/css/Typeahead';
+
+.rbt-input-main {
+  padding: 2px 6px;
+}
+
+.rbt-input-multi {
+  height: fit-content;
+}
+
+.rbt-menu.dropdown-menu {
+  &.show {
+    margin: 0;
+  }
+
+  .dropdown-item.disabled {
+    color: $color-pf-black-600;
+  }
+}
+
+.rbt-token {
+  margin-right: 6px;
+  border-radius: 5px;
+
+  &.rbt-token-removeable {
+    background-color: $color-pf-black-200;
+    height: 20px;
+    font-size: 12px;
+    color: #363636;
+  }
+
+  .rbt-token-remove-button {
+    color: #363636;
+    font-size: 10px;
+    padding: 0 6px;
+    top: 0;
+  }
+
+  .rbt-input-multi {
+    &.rbt-input-wrapper {
+      margin-bottom: -3px;
+    }
+
+    &.rbt-input {
+      height: auto;
+    }
+  }
+}
+
+.rbt-aux .rbt-close {
+  margin-top: 0;
+}
--- a/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
+++ b/app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
@@ -1,9 +1,6 @@
 /*
-  Until we can properly import patternfly-react CSS extensions
-  (which will probably require changes to manageiq-ui-classic's webpack config)
-  we need to include the styles for react-bootstrap-typeahead manually.
-
-  This can probably be removed after https://github.com/patternfly/patternfly-react/issues/673 is fixed.
+  Until we can properly import patternfly-react CSS extensions, we need to include the styles for react-bootstrap-typeahead manually.
+  This can be replaced with a proper import after https://github.com/patternfly/patternfly-react/issues/673 is fixed.
   This is a copy of https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/sass/patternfly-react/_type-ahead-select.scss
 */
 
@@ -11,10 +8,6 @@
 
 .rbt-input-main {
   padding: 2px 6px;
-}
-
-.rbt-input-multi {
-  height: fit-content;
 }
 
 .rbt-menu.dropdown-menu {
--- a/app/assets/stylesheets/manageiq-v2v/base.scss
+++ b/app/assets/stylesheets/manageiq-v2v/base.scss
@@ -1,4 +1,5 @@
 @import '../../../javascript/react/app.scss';
+@import './type-ahead-select.scss';
 
 .react-wrapper {
   height: 100%;
--- a/app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostWizard/ConversionHostWizardHostsStep/ConversionHostWizardHostsStep.js
+++ b/app/javascript/react/screens/App/Settings/screens/ConversionHostsSettings/components/ConversionHostWizard/ConversionHostWizardHostsStep/ConversionHostWizardHostsStep.js
@@ -1,10 +1,26 @@
 import React from 'react';
-import { Form } from 'patternfly-react';
+import { Form, TypeAheadSelect } from 'patternfly-react';
+import { RHV, OPENSTACK } from '../../../../../../../../../common/constants';
 
-const ConversionHostWizardHostsStep = () => (
-  <Form className="form-horizontal">
-    <h2>TODO: Hosts Step Contents</h2>
-  </Form>
-);
+const ConversionHostWizardHostsStep = ({ selectedProviderType, selectedCluster }) => {
+  let hostOptions = [];
+  if (selectedProviderType === RHV) hostOptions = selectedCluster.hosts;
+  if (selectedProviderType === OPENSTACK) hostOptions = selectedCluster.vms;
+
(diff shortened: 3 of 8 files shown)


mturley:
Yay, it looks like we got your permissions fixed @mzazrivec! This one is ready to go if you have any time to review it.

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
```diff
@@ -0,0 +1,61 @@
+/*
+  Until we can properly import patternfly-react CSS extensions
+  (which will probably require changes to manageiq-ui-classic's webpack config)
```
himdel: What's the problem there @mturley ? Do we have an issue?

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+@import 'react-bootstrap-typeahead/css/Typeahead';
+
+.rbt-input-main {
+  padding: 2px 6px;
+}
+
+.rbt-input-multi {
+  height: fit-content;
```
himdel: This is problematic.


According to https://developer.mozilla.org/en-US/docs/Web/CSS/height ,
`fit-content` is only supported in chrome.

Remember we need to support IE11.

mzazrivec:
For some reason, when I login into ManageIQ for the first time and navigate to the second step of the wizzard, I get the following javascript console error:

```
Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
    in input (created by HintContainer(AutosizeInput))
    in div (created by HintContainer(AutosizeInput))
    in HintContainer(AutosizeInput) (created by TypeaheadInputMulti)
    in div (created by TypeaheadInputMulti)
    in div (created by TypeaheadInputMulti)
    in TypeaheadInputMulti (created by InputContainer(TypeaheadInputMulti))
    in InputContainer(TypeaheadInputMulti) (created by Typeahead)
    in div (created by Typeahead)
    in Typeahead (created by WrappedTypeahead)
    in WrappedTypeahead (created by TypeaheadContainer(WrappedTypeahead))
    in TypeaheadContainer(WrappedTypeahead) (created by OnClickOutside(TypeaheadContainer(WrappedTypeahead)))
    in OnClickOutside(TypeaheadContainer(WrappedTypeahead)) (created by TypeAheadSelect)
    in TypeAheadSelect (created by TypeAheadSelectField)
    in TypeAheadSelectField (created by ConnectedField)
    in ConnectedField (created by Connect(ConnectedField))
    in Connect(ConnectedField) (created by Field)
    in Field (created by ConversionHostWizardHostsStep)
    in div (created by FormGroup)
    in FormGroup (created by ConversionHostWizardHostsStep)
    in form (created by Form)
    in Form (created by ConversionHostWizardHostsStep)
    in ConversionHostWizardHostsStep (created by Form(ConversionHostWizardHostsStep))
    in Form(ConversionHostWizardHostsStep) (created by Connect(Form(ConversionHostWizardHostsStep)))
    in Connect(Form(ConversionHostWizardHostsStep)) (created by ReduxForm)
    in ReduxForm (created by Connect(ReduxForm))
    in Connect(ReduxForm) (created by ModalWizardBody)
    in div (created by WizardContents)
    in WizardContents (created by ModalWizardBody)
    in div (created by WizardMain)
    in WizardMain (created by ModalWizardBody)
    in section (created by WizardRow)
    in WizardRow (created by ModalWizardBody)
    in ModalWizardBody (created by ConversionHostWizardBody)
    in ConversionHostWizardBody (created by ConversionHostWizard)
    in div (created by ModalBody)
    in ModalBody (created by WizardBody)
    in WizardBody (created by ConversionHostWizard)
    in div (created by Wizard)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in CustomModalDialog (created by Modal)
    in Transition (created by Fade)
    in Fade (created by DialogTransition)
    in DialogTransition (created by Modal)
    in RefHolder (created by Modal)
    in div (created by Modal)
    in Portal (created by Modal)
    in Modal (created by Modal)
    in Modal (created by Wizard)
    in Wizard (created by ConversionHostWizard)
    in ConversionHostWizard (created by Connect(ConversionHostWizard))
    in Connect(ConversionHostWizard) (created by ConversionHostsSettings)
    in Spinner (created by ConversionHostsSettings)
    in ConversionHostsSettings (created by Connect(ConversionHostsSettings))
    in Connect(ConversionHostsSettings) (created by Settings)
    in div (created by TabPane)
    in Transition (created by Fade)
    in Fade (created by TabPane)
    in TabPane (created by Tab)
    in Tab (created by Settings)
    in div (created by TabContent)
    in TabContent (created by Tabs)
    in div (created by Tabs)
    in TabContainer (created by Tabs)
    in Tabs (created by Uncontrolled(Tabs))
    in Uncontrolled(Tabs) (created by Settings)
    in div (created by Settings)
    in Settings (created by Connect(Settings))
    in Connect(Settings) (created by I18nProviderWrapper(Connect(Settings)))
    in IntlProvider (created by I18nProviderWrapper(Connect(Settings)))
    in I18nProviderWrapper(Connect(Settings)) (created by Route)
    in Route (created by Routes)
    in Routes (created by App)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Connect(ConnectedRouter))
    in Connect(ConnectedRouter) (created by App)
    in App (created by Connect(App))
    in Connect(App)
    in Provider
window.console.(anonymous function) @ miq_debug.self-1351f771c2cd2fab0d83069fd3f62aadbb7effc11a01ed4942f52308793453be.js:30
```

Once I go back & forth in the wizzard, the error no longer shows. Or even when I launch the wizzard again (without logging out).

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
mturley: Yes, https://github.com/patternfly/patternfly-react/issues/673 is tracking the upstream issue. Importing the patternfly-react CSS directly results in a relative import of the Typeahead.css file which doesn't exist, when it should instead be inlined. I made a PR to fix that issue upstream, I'll keep an eye on it and remove this copied CSS when it's resolved.

I was initially thinking we the import was failing due to something in the ui-classic webpack config, but I was wrong, I'll remove that part of this comment.

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
mturley: Thanks for noticing that. This CSS came directly from patternfly-react, and it appears to me that we don't need that style rule anyway. It supposedly was added to fix a bug that I can't seem to reproduce without it. I opened the issue https://github.com/patternfly/patternfly-react/issues/1362 to track that, and I'll remove that CSS from our copy for now. If it turns out to be a real issue, I figure we can track that separately from this PR.

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
himdel: Aah, thanks :) Yeah, I saw that issue but was confused with the ui-classic mention. Thanks :)

mturley:
@mzazrivec that's strange, I can't seem to reproduce that error. We definitely pass an `onChange` handler there, I'll try and figure it out.

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
himdel: As long as this doesn't *break* anything, it's OK to be sub-optimal in IE :+1:

mturley: Yeah, it seems like even if this issue is real (I'm having trouble reproducing it), the worst that would happen is that if a user tried to select more than something like 8-10 hosts, the overflow wouldn't be visible. I can't imagine users would need that many hosts anyway.

mturley:
@mzazrivec it looks like that warning you're seeing is caused by a bug in `react-bootstrap-typeahead` that was apparently fixed in v3.2.3 of that package (https://github.com/ericgio/react-bootstrap-typeahead/issues/381). patternfly-react depends on v3.1.3, so we need to upgrade that version in patternfly-react and then upgrade patternfly-react after that release. I opened https://github.com/ManageIQ/manageiq-v2v/issues/879 to track this on our end, and https://github.com/patternfly/patternfly-react/issues/1366 to track it upstream.

I don't think this should block the PR, since it's just a dev-mode warning.

app/assets/stylesheets/manageiq-v2v/type-ahead-select.scss
mturley: I found an IE11-compatible fix for the upstream issue (https://github.com/patternfly/patternfly-react/pull/1369), so I'll go ahead and apply that fix to the copied styles here!

mturley:
Ok, @himdel @mzazrivec I think this should be ready now.

simaishi:
Hammer backport details:

```
$ git log -1
commit 19084c8d3c1334d7bb83ea5b43263cb15b33d186
Date:   Tue Feb 12 16:43:34 2019 +0100

    Merge pull request #876 from mturley/conv-host-wizard-step2
    
    Conversion Host Configuration Wizard - Step #2 - Host(s)
    
    (cherry picked from commit e64122f30b4a92d36d9bddf297674d5036583674)
    
```

PR #883, opened 2019-02-26
Conversion Host Configuration Wizard - Step #4 - Results (without passing SSH keys)

mturley:

This is part of #855.

The initial version of this wizard step **_does not_** pass authentication credentials (SSH keys) to the API, since that backend support is still in development. A follow-up PR will be made to add that support and close out the rest of #855.

# Screens

This step of the wizard generates a `POST` request to `/api/conversion_hosts` for each selected host or vm, displaying this loading state until all of those requests complete:

screenshot 2019-02-28 21 18 06

If there is an error returned from the API immediately (the task(s) could not even be queued), this error state is displayed in the wizard:

screenshot 2019-02-28 18 36 52

If the conversion host enablement task(s) are successfully queued, this message is displayed (using the same clock icon we use when a migration plan is started immediately at the end of the plan wizard):

screenshot 2019-02-28 20 58 08

Note that the queued task could still fail, but those errors will be reported in the list view, not in the wizard.

# Implementation Details

Rather than duplicate the markup for the loading and error states a third time, I created new shared components `WizardLoadingState` and `WizardErrorState` to consolidate the existing code duplication, and used them in all three wizards. The special CSS that was used for the clock icon at the end of the plan wizard is now shared between two components, so I deleted the `PlanWizardResultsStep.scss` file and moved that CSS over to `ModalWizard.scss` with a more generic class name.

Polling for the task status will be handled outside the scope of this PR.


apinnick:
The error message, "We're sorry, something went wrong. Please try again", sounds a bit lame. You might want to change it to something like "The conversion host cannot be configured. Please check your settings."

mturley:
@apinnick I agree, I was actually copying that from the error message when creating migration plans and infrastructure mappings. I'll update those too, do you think I should use that same phrasing? ("The {migration plan / infrastructure mapping} could not be created. Please check your settings.")

Also, @vconzola has told me that we shouldn't say "Please" in the UI anyway. maybe that second sentence should just be "Check your settings and try again."

apinnick:
@mturley I agree with @vconzola. We don't say "Please" or "Sorry".  :-) 
Your version is better.

simaishi:
Hammer backport details:

```
$ git log -1
commit e328552343e8178a634cb36fbba15cbb04a01d92
Date:   Mon Mar 4 14:34:33 2019 +0100

    Merge pull request #883 from mturley/conv-host-wizard-step4
    
    Conversion Host Configuration Wizard - Step #4 - Results (without passing SSH keys)
    
    (cherry picked from commit 200c02265bb70421487574e81a92b70bb7c54615)
    
```

PR #893, opened 2019-03-05
Get rid of dangerouslySetInnerHTML

mzazrivec:
Seeing that `mappingToDelete.name` is user-generated data, using `dangerouslySetInnerHTML` here would pose a security risk.

### DIFF
--- a/app/javascript/react/screens/App/Mappings/components/DeleteInfrastructureMappingConfirmationModal/DeleteInfrastructureMappingConfirmationModal.js
+++ b/app/javascript/react/screens/App/Mappings/components/DeleteInfrastructureMappingConfirmationModal/DeleteInfrastructureMappingConfirmationModal.js
@@ -27,11 +27,13 @@ const unUsedMappingInPlans = (mappingToDelete, notStartedTransformationPlans, fi
 
     const deleteMessageAboutPlansUsingMapping = (
       <div className="warning-modal-body-enable-scrollbar">
-        <strong>
-          {planNames.map(plan => (
-            <ul key={plan.id}>{plan.name}</ul>
-          ))}
-        </strong>
+        <h4>
+          <strong>
+            {planNames.map(plan => (
+              <ul key={plan.id}>{plan.name}</ul>
+            ))}
+          </strong>
+        </h4>
       </div>
     );
     return (
@@ -42,15 +44,6 @@ const unUsedMappingInPlans = (mappingToDelete, notStartedTransformationPlans, fi
     );
   }
   return '';
-};
-
-const displayDeleteMessage = mappingToDelete => {
-  const mappingNameStyled = `<strong>${mappingToDelete.name}</strong>`;
-  const regularDeleteMessage = sprintf(
-    __('Are you sure you want to delete the infrastructure mapping %s ?'),
-    mappingNameStyled
-  );
-  return <div dangerouslySetInnerHTML={{ __html: regularDeleteMessage }} />;
 };
 
 const DeleteInfrastructureMappingConfirmationModal = ({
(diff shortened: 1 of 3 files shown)


app/javascript/react/screens/App/Mappings/components/DeleteInfrastructureMappingConfirmationModal/DeleteInfrastructureMappingConfirmationModal.js
```diff
@@ -75,7 +66,10 @@ const DeleteInfrastructureMappingConfirmationModal = ({
           {mappingToDelete &&
             unUsedMappingInPlans(mappingToDelete, notStartedTransformationPlans, finishedWithErrorTransformationPlans)}
         </h4>
-        <h4>{mappingToDelete && displayDeleteMessage(mappingToDelete)}</h4>
+        <h4>
+          {mappingToDelete &&
+            sprintf('Are you sure you want to delete the infrastructure mapping %s ?', mappingToDelete.name)}
```
mturley: We still need the `__(...)` gettext call for i18n here. and we also lose the `<strong>` around the name. maybe instead we should just do:
```
<h4>
  {__('Are you sure you want to delete the infrastructure mapping ')}
  <strong>{mappingToDelete && mappingToDelete.name}</strong>
  {' ?'}
</h4>
```

mzazrivec: The thing is that we need to avoid doing string math in gettext strings as much as possible.

In this particular example, it would not be obvious that the string is an actual question, since the question mark comes separately. That would make it impossible to translate it correctly into Spanish language for example (¿ at the beginning of interrogative sentence).

What do you think about the following?

infra-mapping-delete-01

mturley:
@mzazrivec would you mind also adding the following eslint rule to `.eslintrc.js`?
```
'react/no-danger': 'error'
```
That way, any future introduction of `dangerouslySetInnerHTML` will cause CI to fail.

app/javascript/react/screens/App/Mappings/components/DeleteInfrastructureMappingConfirmationModal/DeleteInfrastructureMappingConfirmationModal.js
mturley: That looks good to me. @vconzola are you ok with this slight design change? (previously the name of the mapping was a part of the sentence).

vconzola:
@mturley Yeah, it's fine.

mzazrivec:
> @mzazrivec would you mind also adding the following eslint rule to `.eslintrc.js`?
> 
> ```
> 'react/no-danger': 'error'
> ```

Done. The PR is ready for review.

I also had to change a stylesheet slightly, since the new content wouldn't always fit in.

mturley:
@mzazrivec do you know if we ever opened a BZ for this? Is this the one that we got a security team involved with?

mzazrivec:
@mturley I did not open a bugzilla report for this problem, but there already is an older report (a private one) describing this very issue, created by security team.

mturley:
Ok, I was just curious for the sake of backporting this.

simaishi:
Hammer backport details:

```
$ git log -1
commit 6d8599a40a54b0cb679d247d284fbbc267534139
Date:   Wed Mar 13 09:26:49 2019 -0400

    Merge pull request #893 from mzazrivec/get_rid_of_dangerously_set_inner_html
    
    Get rid of dangerouslySetInnerHTML
    
    (cherry picked from commit 728ea564c29303663919137dad0f4b842be8bebf)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1714739
```

PR #907, opened 2019-03-13
Conversion Host Settings page: Ensure task.meta will always be defined

mturley:
This prevents an unlikely error that would occur if the task name field from the API is malformed. We parse it with a regular expression, and if the format of it changes in the future, that would cause console errors before when we filter on `task.meta.operation`, and with this change a malformed task name will just cause that task to be filtered out.


simaishi:
Hammer backport details:

```
$ git log -1
commit d474c1ce19d48bf6c7e4fdf5a1cd257b6780e43b
Date:   Thu Mar 14 10:30:56 2019 +0100

    Merge pull request #907 from mturley/conv-host-tasks-meta-fix
    
    Conversion Host Settings page: Ensure task.meta will always be defined
    
    (cherry picked from commit dc8eb8ba5c87b14b1284d9c66d6efe827f09edef)
    
```

PR #916, opened 2019-03-21
[#901] Conversion Host Enablement - List View - Add Retry button with modal for re-entering SSH keys

### LINKED ISSUE #901: Conversion Host Enablement - List View - Retry button on a failed enablement task - Need API support
Originally part of #862, I'm hiding this button until the back end support is ready.

**We could release the Conversion Host settings page with this button removed**, and just require users to Remove and then go through the wizard again.

When a conversion host enablement task fails, we have a Retry button that needs to submit some kind of API request to attempt the enablement again.

screenshot 2019-03-08 12 32 48

When we discussed this it was mentioned that the UI should just submit the original enable request again, but there is currently no way to obtain all of the necessary parameters from that original request based on the failed task object. Especially in the case of SSH private keys, I don't think we would be able to store and retrieve all of this necessary data.

~~We may instead need a special API action for retrying the task.~~ see below comments, we decided instead to prompt the user to re-enter their SSH keys.

mturley:
Closes #901.

Note: this depends on https://github.com/ManageIQ/manageiq/pull/18541 to be tested, but we can merge it ahead since the conversion hosts page is still hidden. The retry button will only appear on tasks that were created with that back end change in place.

While we wait for that back end PR to be ready (I'm not sure it even creates the `request_params` yet), to test this PR you can fake some context data by making a change like this: https://github.com/mturley/manageiq-v2v/commit/0acaab26a00d3358a20304d54e9e322417b7ca57

We can also test against this database which has one failed task at the end of the list (but will require a hack like the above, since the task has no context data): https://drive.google.com/open?id=1l26rbhp36v6lCfpwKGASxEJknKnI32oe

# Screens

This PR adds a Retry button to the list view rows on the conversion hosts settings page, which will appear on any failed enablement task row which has `context_data.request_params` in the task object:

Screenshot 2019-03-21 17 07 59

When clicked, a modal will appear explaining that the user needs to re-enter any SSH private keys they provided the first time through the wizard, as they were not saved. The SSH key fields shown match what was shown in the wizard, depending on their original selection of a transformation method. If SSH was originally selected in the wizard, both the conversion host SSH private key and the vmware hypervisors SSH private key fields will appear:

Screenshot 2019-03-20 17 04 39

If VDDK was originally selected in the wizard, only the conversion host SSH private key field will appear:

Screenshot 2019-03-21 10 47 46

The fields are required, and when they are filled the "Retry" button at the bottom will become enabled:

Screenshot 2019-03-21 10 47 54

When the user clicks Retry, the original properties from the wizard are recovered from the task's context data, combined with these re-entered SSH key(s), and sent to the API with the same action the wizard uses to submit.

# Implementation Details

* In order for the redux-form state to be reset when the modal closes, it needs to be fully unmounted when it's not visible. We didn't need this for the Remove/Delete modal, because it has no state, only a button, so leaving that modal always mounted and passing a single `conversionHostDeleteModalVisible` boolean to the `show` prop was sufficient. For this modal however, just like the wizards, if we had used the single boolean as a condition of mounting and always passed `show={true}`, we would lose the animation that happens when `show` turns to false. So just like the wizards, we needed distinct `conversionHostRetryModalMounted` and `conversionHostRetryModalVisible` booleans, toggled on by the show action together, and toggled off by the actions we pass to `onHide` and `onExited` separately. `onExited` gets called when the hide animation completes.
* Because the modal uses the same two SSH key fields from the Authentication step of the wizard, I abstracted those out into a new `SSHKeyField` component which wraps a `TextFileInput` in a redux-form `Field`. I also moved the logic for specifying the info popover message of the conversion host SSH private key into a helper function to be reused.
* The presence or absence of the `vmware_vddk_package_url` in a task's `context_data.request_params` determines whether the modal should use a VDDK transformation method or SSH transformation method.
* There is quite a lot of "prop drilling" from ConversionHostsSettings -> ConversionHostsList -> ConversionHostsListItem. We could clean this up in the future by connecting more of the inner components to redux directly (and all of the components used by ConversionHostsList could be in their own directory) but I decided against a big refactor as part of this PR. I might come back and do that as part of unit testing.


mturley:
Weird, https://travis-ci.org/ManageIQ/manageiq-v2v/builds/510050291 passed but didn't update the status on the PR. Closing and reopening one more time to try it again... @mzazrivec I think I'm just going to stop using draft PRs until we have time to figure out why they break Travis.

simaishi:
Hammer backport details:

```
$ git log -1
commit e5b7742b2506071b8b902e3c8320fd6c596c5d6c
Date:   Tue Mar 26 17:53:41 2019 +0100

    Merge pull request #916 from mturley/901-conv-host-retry
    
    [#901] Conversion Host Enablement - List View - Add Retry button with modal for re-entering SSH keys
    
    (cherry picked from commit 4d7e37cea2aaf83cc5ce8a230c1f556e9b33134f)
    
```

PR #920, opened 2019-03-26
[BZ#1680521] Filter cloud_tenants to Openstack providers explicitly (prevents errors with RHV-4.3 providers)

mturley:
When loading cloud_tenants in the mappings page, overview page, and mapping wizard, we need to be sure to load only OpenStack cloud_tenants, since RHV can now also provide cloud_tenants.

Without this filter, requesting the cloud_volume_types for all cloud_tenants will cause errors (since RHV does not have cloud_volume_types), and RHV tenants will incorrectly appear in the list of target OSP projects in the mapping wizard (causing further errors if chosen for a mapping).

This PR adds an explicit filter on requests for lists of cloud_tenants restricting these requests to those tenants with an OpenStack provider, which prevents these errors.

The error can be reproduced (and the fix verified) by using this database: https://drive.google.com/open?id=1KydneIrh1diGPtM4bqUkH_MYCkz1_3D- and navigating to the mappings page.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1680521

### DIFF
--- a/app/javascript/react/screens/App/Mappings/MappingsConstants.js
+++ b/app/javascript/react/screens/App/Mappings/MappingsConstants.js
@@ -12,7 +12,8 @@ export const SET_V2V_MAPPING_TO_DELETE = 'SET_V2V_MAPPING_TO_DELETE';
 export const SHOW_V2V_DELETE_CONFIRMATION_MODAL = 'SHOW_V2V_DELETE_CONFIRMATION_MODAL';
 export const YES_TO_DELETE_AND_HIDE_DELETE_CONFIRMATION_MODAL = 'YES_TO_DELETE_AND_HIDE_DELETE_CONFIRMATION_MODAL';
 
-export const FETCH_CLOUD_TENANTS_URL = '/api/cloud_tenants?expand=resources';
+export const FETCH_CLOUD_TENANTS_URL =
+  '/api/cloud_tenants?expand=resources&filter[]=ext_management_system.type=ManageIQ::Providers::Openstack::CloudManager';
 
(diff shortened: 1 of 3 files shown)


mturley:
@simaishi, this PR will not backport cleanly to hammer, so I opened a PR against hammer directly for this change: https://github.com/ManageIQ/manageiq-v2v/pull/921

mturley:
Note: this would only not backport cleanly to hammer if we tried to backport it before https://github.com/ManageIQ/manageiq-v2v/pull/871 (which we may need to).

If the conversion host feature including #871 is all backported before this, we can backport this PR instead of using the direct-to-hammer PR https://github.com/ManageIQ/manageiq-v2v/pull/921.

Sorry for the confusion, I'm just trying to keep this all straight in my mind :)

mturley:
@simaishi I switched this back to `hammer/yes` since you already backported #871 and the conflict I was anticipating above is no longer a concern (so I've closed https://github.com/ManageIQ/manageiq-v2v/pull/921).

simaishi:
Hammer backport details:

```
$ git log -1
commit d558275c48df5fc930697d7a1b06bb7bfffa46ba
Date:   Wed Mar 27 12:04:19 2019 +0100

    Merge pull request #920 from mturley/bz1680521-cloud-tenants-filter
    
    [BZ#1680521] Filter cloud_tenants to Openstack providers explicitly (prevents errors with RHV-4.3 providers)
    
    (cherry picked from commit 312673c987b37723abd617456de69889f97e0b88)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693731
```

PR #928, opened 2019-04-04
Conversion Host Settings - Unit tests for helpers, prevent corner case of reappearing failures

mturley:
Part of the Conversion Hosts UI feature BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1693339

As part of https://github.com/ManageIQ/manageiq-v2v/issues/884, I started filling in the gaps in unit testing of the conversion host settings UI code. I started with the helpers that handle generating the combined list items from parsed and indexed task and conversion host objects, on a hunch that there was a corner case I had missed, and I was right:

If you checkout ce342e9 and try to run the new tests, you'll see two failures due to the `{ ...exampleTask(20, 'enable', '11'), ...failed }` task appearing as an active enablement task even though there were successful enable and disable tasks after it. This means if a user had attempted to enable VM 11 and it failed, then attempted again and it succeeded, then removed it and it succeeded, when the removal completed it would show "enable failed" again instead of disappearing from the list.

This bug is due to the fact that the `getActiveConversionHostEnableTasks` performed these filters in the wrong order:

* filter to only enable tasks that are in progress or failed
* filter those to only the latest task per resource, and only those which don't match a configured conversion host

This was fine in the case of a failed enablement that had been retried, because the first filter would include the failure, but the second filter would remove it since it matched a configured host. But as soon as that host is removed, the second filter wouldn't remove the failure anymore, and we get the bug.

60e6246 fixes this bug by simply reversing the order of these two filters:

* filter to only the latest task per resource, and only those which don't match a configured conversion host
* filter those to only enable tasks that are in progress or failed

This way, only the latest task for each resource can be shown in the list, regardless of its state. With the filters reversed, all tasks for VM 11 are filtered out, because the latest task is a disable task, and when the disable completes, that resource is (correctly) no longer present in the list.

### DIFF
--- a/app/javascript/react/screens/App/Settings/__tests__/helpers.test.js
+++ b/app/javascript/react/screens/App/Settings/__tests__/helpers.test.js
@@ -1,0 +1,137 @@
+import {
+  parseConversionHostTasksMetadata,
+  indexConversionHostTasksByResource,
+  getActiveConversionHostEnableTasks,
+  attachTasksToConversionHosts,
+  getCombinedConversionHostListItems
+} from '../helpers';
+
+const defaultType = 'ManageIQ::Providers::Openstack::CloudManager::Vm';
+const inProgress = { state: 'Active', status: 'Ok', message: 'Example progress message' };
+const failed = { state: 'Finished', status: 'Error', message: 'Example error message' };
+const success = { state: 'Finished', status: 'Ok', message: 'Example success message' };
+
+const exampleTask = (updated_on, operation, id) => ({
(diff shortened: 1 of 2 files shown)


simaishi:
Hammer backport details:

```
$ git log -1
commit 11a670309d45854610682787dccf372b1cff3b42
Date:   Fri Apr 5 10:19:52 2019 +0200

    Merge pull request #928 from mturley/settings-helper-tests
    
    Conversion Host Settings - Unit tests for helpers, prevent corner case of reappearing failures
    
    (cherry picked from commit f6dfc222053d10d30de268c253176aeea32dabbc)
    
```

PR #930, opened 2019-04-08
Add RSpec framework files

Fryguy:
@djberg96 Please review.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,5 @@
+--require manageiq/spec/spec_helper
+--require spec_helper
+--color
+--order random
+--exclude-pattern "spec/manageiq/**/*_spec.rb"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,28 @@
-language: node_js
-node_js:
-  - '8' # current stable
-script: yarn test --runInBand
+dist: xenial
+matrix:
+  include:
+  - name: "Node.js: 8"
+    language: node_js
+    node_js:
+    - '8' # current stable
+    cache: yarn
+    script: yarn test --runInBand
+  - name: "Ruby: 2.4.5"
+    language: ruby
+    rvm:
+    - 2.4.5
+    cache: bundler
+    addons:
+      postgresql: '10'
+    install: bin/setup
+  - name: "Ruby: 2.5.5"
+    language: ruby
+    rvm:
+    - 2.5.5
+    cache: bundler
+    addons:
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -1,7 +1,17 @@
-language: node_js
-node_js:
+matrix:
```
Fryguy: I'm not sure if this is the right way to do this...I basically need two completely different languages, and reading the docs, it seems the matrix is the way to do this.

Fryguy: And I didn't want to do it the way the ui-classic does it because they nvm install on demand, and I was trying to find a way to use the node travis image directly.

Fryguy:
cc @jrafanie @juliancheal I just wanted you to see what changes I had to do here to get multi-language support, and so if you have to blast out ruby upgrades, this repo is a little different.

.rspec
```diff
@@ -0,0 +1,5 @@
+--require manageiq/spec/spec_helper
+--require spec_helper
```
djberg96: Unless having a local spec_helper.rb is required, I would leave this line out.

spec/example_spec.rb
```diff
@@ -0,0 +1,5 @@
+describe MigrationLogController do
+  it "stuff" do
+    expect(1).to eq 1
+  end
+end
```
djberg96: Is this file required?

.rspec
djberg96: Oops, nm, I see that you added it.

djberg96:
@Fryguy overall LGTM, though I'm not an expert on the JS stuff. @mturley or @vconzola Can either of you comment?

spec/example_spec.rb
Fryguy: It was my placeholder for your tests otherwise travis would have nothing to run.

djberg96: Ah, ok, thanks.

.travis.yml
```diff
@@ -1,7 +1,28 @@
...
-script: yarn test --runInBand
+dist: xenial
+matrix:
+  include:
+  - name: "Node.js: 8"
+    language: node_js
+    node_js:
+    - '8' # current stable
+    cache: yarn
+    script: yarn test --runInBand
```
Fryguy: This part is basically the same as before, but added `cache: yarn` for performance.

mturley: Sounds good to me.

PR #934, opened 2019-04-16
[#927] Add option to sort plans by Completed time, and fix alignment in plan list views

### LINKED ISSUE #927: [BZ#1694741] Completed Migration plans cannot be ordered by execution order
https://bugzilla.redhat.com/show_bug.cgi?id=1694741

In the Migration Plans Completed and Migration Plans Archived views, add a sort option for the creation date of the latest plan request. This should be made available as a property on the plan object for reactabular to use, as part of processing the plan objects for display here: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Overview/helpers.js#L7

mturley:
Closes #927.
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1694741

This PR:
* Adds a new sort option called "Completed" to the Completed Plans and Archived Plans list views, and sets it as the default sort.
* Adds the Completed timestamp to the list view items, to the right of the elapsed time. If the plan is scheduled for future (retry) migration, the Completed timestamp is not shown, because the Completed and Scheduled timestamps would not both fit comfortably in the row.
* Overrides the width of the list view's two main sections (name/description and additional info) from 50%/50% to 25% and 75% width respectively. This is in order to make more room in the additional info for the Completed timestamp and prevent the row from wrapping in an ugly way. This applies to all 3 list views (including Not Started) for consistency.
* Adds a `Visibility` helper component which conditionally applies `visibility: hidden` CSS to its children
* Uses the `Visibility` helper instead of conditional rendering to hide action buttons in the list view that should not appear. This ensures that the `.list-view-pf-actions` element will always have the same width in each row of the list (without having to specify a static width), which fixes the issues we were having with alignment of the details in the center of the list items (this also applies to the Plans Not Started list view).
* Refactors the `ScheduleMigrationButtons` component into two separate components `ScheduleMigrationButton` and `EditScheduleMenuItems` to make it easier to apply the visibility helper to the Schedule button without messing up the kebab menu. These shouldn't have been in the same component anyway ever since we moved the Edit/Delete schedule items into the kebab menu, so this cleanup eliminates some confusing code.

# Screens

"Completed" sort option:
Screenshot 2019-05-06 14 08 06

"Completed" timestamps in the list:
Screenshot 2019-05-06 14 01 05

Scheduled timestamp taking precedence over Completed timestamp:
Screenshot 2019-05-06 14 00 43

Alignment fixed in the Not Started list view:
Screenshot 2019-05-06 13 56 42

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/EditScheduleMenuItems.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/EditScheduleMenuItems.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Button, MenuItem, Icon } from 'patternfly-react';
+import { MenuItem, Icon } from 'patternfly-react';
 import { formatDateTime } from '../../../../../../components/dates/MomentDate';
 
-const ScheduleMigrationButtons = ({
+const EditScheduleMenuItems = ({
   showConfirmModalAction,
   hideConfirmModalAction,
   loading,
@@ -14,8 +14,7 @@ const ScheduleMigrationButtons = ({
   plan,
   isMissingMapping,
   migrationScheduled,
-  migrationStarting,
-  showInitialScheduleButton
+  migrationStarting
 }) => {
   const confirmationWarningText = (
     <React.Fragment>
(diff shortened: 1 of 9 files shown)


mturley:
@vconzola, I think maybe we should merge this as-is, and if you want to revise the design later we can approach that as a separate issue. Are you ok with that?

vconzola:
I guess I don't understand the logic of only having the sort on completed and archived plans. I also don't understand exactly which column in the plans view you're sorting on. Can you show me in a quick BJs call?

mturley:
Sure.

vconzola:
My original designs had a timestamp in the list view, but for some reason it was never implemented.

mturley:
I can add that timestamp to the list easily if we just figure out where it fits. Maybe we can make this clock icon and/or the elapsed time into a link that opens a popover with this "Date Migrated:" timestamp in it.

Screenshot 2019-05-02 10 53 31


That's the other `needs-design` piece of this too, is "Date Migrated" a good name for this?

mturley:
Note to self: we decided on "Completion Time" for this field name, and it should be based on the last updated timestamp of the plan request. I'll need to find a place for it in the list view.

mturley:
In order to make this fit into the list view I'll have to play around with the layout a little. Still WIP for now.

vconzola:
@mturley Since you're still WIPing, please change the field label to "Completed:". Thanks.

mturley:
This is now ready for review again. Updated the PR description with some implementation info and screenshots. Still WIP only until I get confirmation from Vince about these decisions, otherwise I think it's ready to merge.

@vconzola does this all look/sound good? In particular, I wonder if you're ok with me applying the 25%/75% width override to the Plans Not Started list view as well for consistency (see the 4th screenshot above). Earlier you wanted me to get confirmation from Fabien on whether it's ok to let Scheduled time take precedence in the view over Completed time, but he wasn't originally asking me to add that timestamp to the list at all, so I think if you're ok with it we're good.

mturley:
Removed `wip`, I got confirmation from Vince that this is good.

mturley:
@vconzola @fdupont-redhat are you guys okay with the decision I made here to hide the Completed time if there is a Scheduled time for a completed(failed) plan that is scheduled for retry? @AparnaKarve pointed out that it is a little odd. There isn't room to display both Completed and Scheduled timestamps in the list, so once you schedule the plan it is still being sorted by the Completed time that isn't shown.

fdupont-redhat:
What about showing the Scheduled timestamps only if the plan hasn't run yet, and show the Completed timestamps when it has ?

mturley:
@fdupont-redhat that is essentially what will happen here, if you're referring to the newly requested run of the plan. The plan has already run and completed (failed), and is scheduled to run again, but hasn't run a second time yet. So it shows its new Scheduled time. When it runs it will move to the In Progress view and then come back to the Completed view with a new Completed timestamp showing.

fdupont-redhat:
Oh. That's right. I forgot about the different views. Then, LGTM

mturley:
Perfect. Thanks!

simaishi:
Hammer backport details:

```
$ git log -1
commit 426d378c8948fee0e95ef0b3bdc5c2371949a4eb
Author: Aparna Karve <akarve@redhat.com>
Date:   Wed May 8 13:09:46 2019 -0700

    Merge pull request #934 from mturley/927-plan-sort-by-date
    
    [#927] Add option to sort plans by Completed time, and fix alignment in plan list views
    
    (cherry picked from commit 3a6f5fc5f07c6827cdf1d9ecf82d376a02746d42)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1711034
```

Issue #935, opened 2019-04-16
We should refactor our use of `UNSAFE_componentWillReceiveProps` to use safer patterns


mturley:
We renamed the usages of this method to include the `UNSAFE_` prefix in #890 to make our codebase safe for an upgrade to React 17+, but we should eventually move away from this deprecated method entirely to a more reasonable pattern (possibly using `componentWillUpdate` or `getDerivedStateFromProps` or memoization helpers).

See https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops for more info about why this method was deprecated and what to use instead.

mturley:
Still should be done, but low priority.

vconzola:
Close as wontfix - maybe someday.

PR #936, opened 2019-04-16
[#890] Rename unsafe/deprecated React lifecycle methods

### LINKED ISSUE #890: Deprecated React lifecycle methods need to be renamed to UNSAFE_* before we upgrade to React 17
There is a need in manageiq-ui-classic to upgrade to React 16.8 to use Hooks (discusson here: https://github.com/ManageIQ/manageiq-ui-classic/pull/5293#discussion_r262050477). It would be nice for us to get this upgrade as well and be able to use React Hooks in v2v too.

That upgrade should be fine, as the deprecated method names will continue to work until React 17. But to prepare ourselves for further upgrades, we should run the [`rename-unsafe-lifecycles`](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) codemod at a minimum, and probably also move away from the deprecated methods entirely.

mturley:
Closes #890.

This PR applies the [`rename-unsafe-lifecycles`](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) codemod to our files, adding the `UNSAFE_` prefix to our usages of `componentWillReceiveProps`. If/when we upgrade to React 17+, the original names of these methods will no longer work, so we need to use the UNSAFE_ names until we refactor away these methods.

This PR only renames the methods, we should refactor them away entirely in the future (#935).

### DIFF
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
@@ -64,7 +64,7 @@ class MappingWizardDatastoresStep extends React.Component {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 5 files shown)


mturley:
@himdel when this is merged, we'll be safe for a future upgrade to React 17+

himdel:
> @himdel when this is merged, we'll be safe for a future upgrade to React 17+

@mturley thanks! :)
Cc @Hyperkid123

PR #937, opened 2019-04-16
[WIP] Fix table inline edit buttons positioning

mzazrivec:
https://bugzilla.redhat.com/show_bug.cgi?id=1693382


app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardInstancePropertiesStep/components/TableInlineEditRow/TableConfirmButtonsRow.js
```diff
@@ -96,19 +36,21 @@ class TableConfirmButtonsRow extends React.Component {
     const elements = [
       <tr ref={this.saveRowDimensions} className={rowClass} key="tableRow">
         {this.props.children}
+        <td
+          style={{
+            borderCollapse: 'collapse',
+            borderSpacing: '0px',
+            maxWidth: '0.01px',
```
mturley: Sub-pixel decimal width values is something I've never seen before 😂  I do wonder whether we can avoid that.

mturley:
@mzazrivec I have a solution for you!

I gave up on trying to make a zero-width cell or a zero-height row work with CSS, and instead found a way to just render the buttons as a child of the last cell that we already had rendered. Check it out:

YMEdQHY1ZV

This solution is about as dodgy from a React perspective as yours was from a CSS perspective, though... the tricky thing is that since the cell I wanted to modify is dynamically rendered by reactabular and buried deep in `this.props.children`, I had to dust off the [React Top-Level API](https://reactjs.org/docs/react-api.html#reactchildren) to get in there and manipulate them at render time.

Here's the commit, I think this might be our best bet: https://github.com/mturley/manageiq-v2v/commit/153b5a901975d330f1c8deab362a64e1a79da09c

These things used to be easier in jQuery days, didn't they...

mzazrivec:
@mturley nice work Mike!

I was also playing with `this.props.children` initially, but could not figure out how to drill down to the correct cell data.

I like your solution better, it's more elegant and programatically cleaner. Plus I verified, that it solves the problem.

If you'd agree, I could close my PR and you would create a new one with your changes, so that I can properly review & test it.

PR #944, opened 2019-04-22
[#884] Add missing unit tests for ConversionHostsSettings files

### LINKED ISSUE #884: Conversion Host Enablement - Unit Tests

mturley:
Closes #884.

Lots of unit tests look the same... this makes codeclimate very upset... ¯\\\_(ツ)\_/¯

This PR includes a few non-test changes that should have no effect on the user:
* In https://github.com/ManageIQ/manageiq-v2v/pull/944/commits/d872f4fef82c8fe43aedc1066d240f35b187954d I made the `mappings` CSS class more generic by renaming it to `white-bg` which is all it did. I also removed `<React.Fragment>` from a few components where it was unnecessary because those fragments only had one child.
* In https://github.com/ManageIQ/manageiq-v2v/pull/944/commits/ce166893067e5e148b6aec92c0a906ac2b751062 I changed the way the wizard steps are exported, so they could be identified by unique display names in the tests.
* In ﻿﻿https://github.com/ManageIQ/manageiq-v2v/pull/944/commits/09f18ce3dec64d9784799a69bf4ee5d5450c90d6 I removed the custom `{({ input }) => <Form.FormControl {...input} type="text" />}` I had left behind in the VDDK library path field, which was cruft from when I had to use that custom render prop to make space for the Validate button, which we ended up removing before it was ever enabled. `<Field component={FormField} type="text"/>` is sufficient here and renders the same thing.
* In https://github.com/ManageIQ/manageiq-v2v/pull/944/commits/b2f12b780922af5707326626cf3ecd799121f9e7 I removed a leftover propType for `unregisterFieldAction` that I had left behind from earlier attempts to fix an issue with redux-form field registration. This prop is never used, so I removed it to clean up the snapshot.


mturley:
Well it looks like codeclimate has an issue with unit tests being repetitive ¯\\\_(ツ)\_/¯

These are finally done, but still WIP just until https://github.com/ManageIQ/manageiq-v2v/pull/961 is merged because I want to add a test case for that bug.

mturley:
Lots of unit tests look the same... this makes codeclimate very upset... ¯\\\_(ツ)\_/¯

These tests have been trickling in as I find free time, and I'm finally done here. Marking it not WIP.

PR #955, opened 2019-05-16
[#954] [BZ#1708365] Mapping Wizard: Clusters Step - hide RSA key warning if all conversion hosts have authentications

### LINKED ISSUE #954: [BZ#1708365] Mapping Wizard: Clusters Step - OSP RSA Key Pair warning appears even when there is a conversion host which has been configured via the UI with an SSH key
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1708365
CFME Target Release: 5.10.5

As seen in the BZ, even after the user has configured a conversion host for OpenStack via the UI (which included specifying an SSH key), when they try to create a mapping to OpenStack they still get this warning in the Clusters step of the wizard:

Screenshot from 2019-05-09 13-46-39

That warning is checking for an authentication directly in the provider, which was the old-school way of authenticating a conversion host. Now the authentication in the provider is no longer required if there is an authentication present in all conversion hosts configured for OpenStack. So we need to look at the conversion hosts and hide this warning if the key is present there.

mturley:
Fixes #954.
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1708365

This can be tested using the `conversion-host-enabled` database from [my collection](https://drive.google.com/open?id=1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q). That database has conversion hosts configured via the UI with an SSH key, and an OSP provider without an SSH key. In this case, on master, this warning will appear when trying to create a mapping to OpenStack:

Screenshot 2019-05-16 15 15 43

And after this PR, it will not. The warning will still appear if the conversion host has been configured manually (the old way, directly running the playbook) without an SSH key, and there is no key in the provider.

### DIFF
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
@@ -56,6 +56,7 @@ class MappingWizardClustersStep extends React.Component {
       isRejectedTargetClusters,
       targetProvider,
       rhvConversionHosts,
+      ospConversionHosts,
       providers,
       isQueryingProviders
     } = this.props;
@@ -87,6 +88,7 @@ class MappingWizardClustersStep extends React.Component {
         isFetchingTargetClusters={isFetchingTargetClusters}
         targetProvider={targetProvider}
         rhvConversionHosts={rhvConversionHosts}
+        ospConversionHosts={ospConversionHosts}
         providers={providers}
         isQueryingProviders={isQueryingProviders}
       />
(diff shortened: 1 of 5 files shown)


simaishi:
Hammer backport details:

```
$ git log -1
commit d9447baea419141021449388739f09da8dc4d963
Date:   Fri May 17 09:50:14 2019 +0200

    Merge pull request #955 from mturley/954-osp-ssh-key-warning-fix
    
    [#954] [BZ#1708365] Mapping Wizard: Clusters Step - hide RSA key warning if all conversion hosts have authentications
    
    (cherry picked from commit ccf654f6d37681b9fee97664e1ab89755d23d3b4)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1711285
```

PR #960, opened 2019-05-23
[WIP][V2V] Move V2V-Related Models from Core to this Repo

jerryk55:
Move all V2V-related model code from the ManageIQ/manageiq (core) repo to manageiq-v2v.

In order to keep all V2V code in a common repo, the model code (and related spec tests) are being moved to this repo. A separate PR will remove this set of code from the core repo.

A second PR will move API code from the manageiq-api repo to this repo as well as another to remove that code from the api repo.

This PR has no dependencies.

@djberg96 @agrare @roliveri please review.

### DIFF
--- a/app/models/conversion_host.rb
+++ b/app/models/conversion_host.rb
@@ -1,0 +1,394 @@
+require 'resolv'
+
+class ConversionHost < ApplicationRecord
+  include NewWithTypeStiMixin
+  include AuthenticationMixin
+
+  acts_as_miq_taggable
+
+  belongs_to :resource, :polymorphic => true
+  has_many :service_template_transformation_plan_tasks, :dependent => :nullify
+  has_many :active_tasks, -> { where(:state => ['active', 'migrate']) },
+    :class_name => "ServiceTemplateTransformationPlanTask",
+    :inverse_of => :conversion_host
+
+  delegate :ext_management_system, :hostname, :ems_ref, :to => :resource, :allow_nil => true
+
+  validates :name, :presence => true
+  validates :resource, :presence => true
+
+  validates :address,
+    :uniqueness => true,
+    :format     => { :with => Resolv::AddressRegex },
+    :inclusion  => { :in => ->(conversion_host) { conversion_host.resource.ipaddresses } },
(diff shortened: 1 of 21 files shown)


jerryk55:
@miq-bot add-label hammer/no

agrare:
Hey @jerryk55 you probably want to use git filter branch to maintain the history of these files across repos.  @juliancheal has a tool to assist with this (https://github.com/juliancheal/code-extractor)

jerryk55:
Oh cool, thanks @agrare.  I'll look at the tool.

mturley:
@agrare, that `code-extractor` tool looks interesting, but can it create commits to an existing repo? It looks like the use case described there is creating an entirely new repo with the extracted code. We might have to do something like this afterwards: https://saintgimp.org/2013/01/22/merging-two-git-repositories-into-one-repository-without-losing-file-history/

agrare:
@mturley yeah you just have to merge unrelated histories afterwards, I've done this a few times on the manageiq-providers-vmware repo.

mturley:
Nice. Seems like a really useful approach.

mturley:
@jerryk55 after this is merged we should decide on how we'll share the newly combined issue/PR lists, I'm thinking maybe we put a `ui` label on all existing open issues and an `api` or `model` label on the new backend-related ones, or something like that. I'm not sure if you guys will want to use GitHub Projects here, but I'm thinking we'll distinguish the [existing project board](https://github.com/ManageIQ/manageiq-v2v/projects/1) as UI-only.

jerryk55:
Yeah @agrare I tried a test run of the tool and was confused about what the results were.  I thought it would add commits to an existing repo (which is what we want).  Instead it creates an empty git tree, apparently with two branches - one with the extracted files, and one with the original repo minus the extracted files.  Not very user friendly.  I'm trying to figure out how to use this in the way we need...

agrare:
@jerryk55 yeah you have to add the remote for the new repo, create a new branch from the master branch, and `git merge --allow-unrelated-histories` from your extracted branch.  Then push that as your PR if I remember correctly

jerryk55:
Ok closing this PR in order to go through the steps suggested by @agrare above in a separate PR.

PR #964, opened 2019-05-29
[WIP][V2V] Migrate the V2V-related Models from the Core ManageIQ repo to the manageiq-v2v repo.

jerryk55:
The subject says it all.  These files are being migrated to the v2v repo with their history.
A separate PR will pull in the V2V-related API code from the manageiq-apI repo, and finally 2 PRs will remove the code from the old repos.

@mturley @djberg96 @agrare @roliveri for your viewing pleasure.  Thanks for reading.


mturley:
Thanks for doing this @jerryk55, it'll be nice to have this all decoupled from core.

mturley:
I added the three new labels `UI`, `API` and `model` to this repo, and labelled all of the other existing open issues and PRs with `UI`. We don't have to use these necessarily, or we can change them later, but I figured it was a good starting point.

jerryk55:
Had to push 3 more factory files that I'd left out of the list the first time around.

Fryguy:
@agrare Since you have merge rights to both core and here, I'm assigning to you as this will need to be merged together with https://github.com/ManageIQ/manageiq/pull/18830

jerryk55:
@miq-bot remove_label wip

jerryk55:
@agrare re-extracted the full set of files and pushed them again (against my same v2v branch).  That seems to be right but not sure since the commit count seems to be increasing astronomically....

agrare:
@jerryk55 it looks like you have two of every commit?

jerryk55:
@agrare should I be using a new branch in this repo each time I rerun the extract, pull, and merge?  Won't that essentially create a new PR when I push that branch back to origin?  Or is there some way to "undo" my previous commits against this branch first?

agrare:
@jerryk55 you can delete and re-create a new local branch then force-push to the same branch on your fork thus keeping the same PR

mturley:
@jerryk55 you can also undo the commits against this branch with a `git reset --hard master` if you prefer that to making a new branch. After you merge to the branch again you'll need to `git push --force` to overwrite the existing history on github.

agrare:
@mturley but you need an empty branch to extract into so resetting on master wont do it.

jerryk55:
@agrare I'm extracting into a new empty branch each time.  That branch then gets used as the remote for the manageiq-v2v repo branch.  *this* branch is the one that I"m re-using.

mturley:
Ah, good point.

jerryk55:
@agrare @mturley went the reset --hard route, then -f on the push.  Seems better - down to 340 commits now.

jerryk55:
@agrare the Travis tests won't be able to succeed until https://github.com/ManageIQ/manageiq/pull/18830 is merged because the 3 factory files that are moving are being pulled in from the core repo first and then from the v2v repo.

jerryk55:
Putting the WIP handle back on until further issues are resolved.

jerryk55:
@miq-bot add_label wip

PR #980, opened 2019-06-10
[#875] Replace Configure Providers button with distinct links to infra and cloud providers

### LINKED ISSUE #875: "Configure Providers" button in empty state takes user to page for VMW and RHV providers, but not for OSP.
As pointed out by @michaelkro in https://github.com/ManageIQ/manageiq-v2v/pull/871#discussion_r253689717, when the user has insufficient providers configured, meaning they don't have at least one source provider (VMW) and at least one target provider (RHV or OSP) configured, we show a "Configure Providers" button that takes them to:

compute -> infrastructure -> providers

which is where a user would need to go to configure VMW and RHV providers, but not for OSP. Not sure what the best thing to do is here.

mturley:
Fixes #875.

This PR affects the empty state view that is shown when the user has no providers configured. To test it, you can modify the Overview/index.js file and/or the Mappings/index.js file so that the `hasSufficientProviders` prop is false.

Instead of the single "Configure Providers" button that we had before, which linked to `Compute > Infrastructure > Providers`, (which was not helpful for OpenStack providers) we now have distinct links to both that page and `Compute > Cloud > Providers`.

To reduce code duplication I consolidated these two identical empty state views into a new component called `NoProvidersEmptyState`.

As seen in the Plans view:

Screenshot 2019-06-10 16 08 39

As seen in the Mappings view:

Screenshot 2019-06-10 16 08 45

### DIFF
--- a/app/javascript/react/screens/App/Mappings/Mappings.js
+++ b/app/javascript/react/screens/App/Mappings/Mappings.js
@@ -8,8 +8,8 @@ import InfrastructureMappingsList from './components/InfrastructureMappingsList/
 import { FETCH_TRANSFORMATION_PLANS_URL, FETCH_ARCHIVED_TRANSFORMATION_PLANS_URL } from '../Overview/OverviewConstants';
 import { FETCH_TRANSFORMATION_MAPPINGS_URL, FETCH_CLOUD_TENANTS_URL } from './MappingsConstants';
 import { FETCH_V2V_PROVIDERS_URL } from '../../../../redux/common/providers/providersConstants';
-import ShowWizardEmptyState from '../common/ShowWizardEmptyState/ShowWizardEmptyState';
 import BreadcrumbPageSwitcher from '../common/BreadcrumbPageSwitcher';
+import NoProvidersEmptyState from '../common/NoProvidersEmptyState';
 
 class Mappings extends Component {
   constructor(props) {
@@ -228,14 +228,7 @@ class Mappings extends Component {
               showMappingWizardEditModeAction={showMappingWizardEditModeAction}
             />
           ) : (
-            <ShowWizardEmptyState
-              description={
-                __('The source and target providers must be configured before attempting a migration') // prettier-ignore
-              }
-              buttonText={__('Configure Providers')}
-              buttonHref="/ems_infra/show_list"
-              className="white-bg"
-            />
+            <NoProvidersEmptyState className="white-bg" />
           )}
         </Spinner>
         {mappingWizardVisible && this.mappingWizard}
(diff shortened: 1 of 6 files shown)


mzazrivec:
The change looks good and I verified that it works OK.

One thing that occured to me though is that the link saying "View VMware and Red Hat Virtualization Providers" is a bit misleading. In truth, the link takes you to list of infra providers, but it won't show any VMware or RHEV providers, since (by definition), there aren't any.

If we'd replace those links with `/ems_infra/new` and `/ems_cloud/new`, click on those would take you directly to the form for adding a new infra / cloud provider, which I guess in this particular situation would be more intuitive.

@mturley @vconzola what do you guys think?

mturley:
@mzazrivec that makes sense to me, I would also be okay with just changing the text of the links to "Configure Infrastructure Providers" and "Configure Cloud Providers", if there is some reason why we don't want to take the user directly to the `/new` pages. @vconzola I think we had discussed that, but I can't remember why we decided against linking directly to `/new`, do you remember?

vconzola:
@mturley @mzazrivec The reason I wrote it like I did is that we don't know what provider type is missing. It can be VMware, RHV, or OpenStack. My thought/hope was that the first link would take the user to a page where they could see if they have VMware/RHV providers, and if not, add what's needed. Same for the second link with respect to OpenStack providers. The logic still makes sense to me. Presumably, a user would not go to Migration if they didn't think they have providers configured, so if the migration UI is telling them a provider is missing they'd want to see what providers there are. Make sense?

mturley:
@vconzola that makes sense. I think @mzazrivec's point was just that it says "View VMware providers" and when you click it there won't be any VMware providers to view, so it's just confusing phrasing. Maybe we just use the word Configure instead of View, and change it to "Configure VMware and Red Hat Virtualization providers" and "Configure Red Hat OpenStack Platform providers"?

vconzola:
@mturley That's my point. We don't know if there are any VMware providers to view. User could be missing the VMware provider or the RHV provider. We don't know. That's why I want them to go to the infrastructure providers page to "see" what's there. I don't want to take them directly to a page where they can configure a new provider without them seeing what's there first.

mturley:
@vconzola I understand and agree with that, I was proposing we take them to the providers page as you intended but use the word Configure instead of View for the link. That matches the wording we have today for the singular Configure Providers button that takes you to the infrastructure providers page.

mturley:
But View makes perfect sense to me too, we can leave it as is.

mzazrivec:
Yup, I understand why we want to take users to the providers list screen and yes, you don't necessarily know what type of provider is missing.

But even with the above explanation it is still quite easy to create a situation where you have zero or more providers configured, none of those will be VMware / RHEV and you'll see the link saying "View VMware / RHEV providers", you click on it and there won't be any VMware / RHEV providers shown.

Or to put it differently, on that landing screen we'll have one line saying that you need *at least one* VMware / RHEV / OpenStack provider, but the next lines indicate that you can click on a link which will show you VMware / RHEV / OpenStack providers (and those links may or may not show those types of providers).

Anyway, I guess it is not a big deal. Merging.

mturley:
@mzazrivec that makes sense and I see how that could be confusing. @vconzola if you're fine with it the way you mocked it up, that's what we merged, but let me know if you want to change it and I'll make another PR.

vconzola:
@mturley I'm happy to leave it as is and see if anyone balks. But if you feel like changing it, I suggest leaving the link text alone, adding a heading that says Missing Providers, and making the current heading text a subheading. That would be more in line with the PF design guidelines for an empty state anyway.

mturley:
@vconzola I accidentally missed the missing providers screen that shows up on the conversion hosts page, so i'll make those changes while I'm in there 😄

PR #988, opened 2019-06-13
[#968] Migration Analytics #1 - Add disabled landing page

### LINKED ISSUE #968: Migration Analytics #1 - Create new page skeleton with empty state
To get started with Migration Analytics, this issue covers creating the new menu entry (and breadcrumb switcher entry) and the landing page with an empty state.

This new page should start as a disabled/hidden feature with a flag to enable it locally, similarly to how we developed the Conversion Hosts Settings view.

mturley:
Closes #968.

This PR adds a new page at `/migration#/analytics` with an empty state landing page:

Screenshot 2019-06-13 19 04 24

The first commit adds the Migration Analytics entries in the app menu and in the breadcrumb switcher, but https://github.com/ManageIQ/manageiq-v2v/commit/23241414bac9d7debd95f7158d219b4401177ddd hides them. We can test the new page prior to release by reverting that commit locally, or by flipping the `hideMigrationAnalytics` flag to false in `BreadcrumbPageSwitcher` and using the switcher, or by just entering `/migration#/analytics` in the URL bar.

The "Get started" button currently doesn't do anything, it is out of the scope of the issue #968.


lib/manageiq/v2v/engine.rb
```diff
@@ -10,12 +10,15 @@ def self.vmdb_plugin?
...
+    # TODO: Un-comment the analytics menu item below when we are ready to release it
+
     initializer 'plugin' do
       Menu::CustomLoader.register(
         Menu::Section.new(:migration, N_("Migration"), 'fa fa-plus', [
           Menu::Item.new('plans', N_("Migration Plans"), 'migration', {:feature => 'migration', :any => true}, '/migration#/plans'),
           Menu::Item.new('mappings', N_("Infrastructure Mappings"), 'mappings', {:feature => 'mappings', :any => true}, '/migration#/mappings'),
-          Menu::Item.new('settings', N_("Migration Settings"), 'migration_settings', {:feature => 'migration_settings', :any => true}, '/migration#/settings')
+          Menu::Item.new('settings', N_("Migration Settings"), 'migration_settings', {:feature => 'migration_settings', :any => true}, '/migration#/settings')# ,
+          # Menu::Item.new('analytics', N_("Migration Analytics"), 'migration_analytics', {:feature => 'migration_analytics', :any => true}, '/migration#/analytics')
```
mzazrivec: We're going to need a new product feature called `migration_analytics` for this new menu item. Is there a core PR for this somewhere? Or should I perhaps create one?

mzazrivec:
Things look good here, the new screen is working. I'll merge once we have the product feature for this in ManageIQ core repo.

lib/manageiq/v2v/engine.rb
mturley: Hmm, good point. I didn't create one, but I'm not sure if we ever created one for `mappings` or `migration_settings` either. Can you point me to where those are defined? we might be able to just have these all under the `mappings` feature. @fdupont-redhat do you know if we need these menu items to be under different CF features?

mturley:
Let's also hold off merging this until I remove/replace the word "CloudForms" in there, I didn't think of that before (thanks miq-bot). I'm not sure how to handle having the app automatically display "ManageIQ" vs "CloudForms" depending on upstream/downstream build. I'll ask @vconzola, maybe we can rephrase this so it doesn't need to be there.

lib/manageiq/v2v/engine.rb
mzazrivec: This is where the current product features for v2v are defined: https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml#L1115

mturley: oh weird, conversion hosts are listed under migration overview, they should be under migration settings

mturley: Ok, sounds like we will need a `migration_analytics` product feature then. Do you mind making the PR for that @mzazrivec? good catch.

mzazrivec: @mturley Those `conversion_host*` product features were added b/c of API it seems. Perhaps we could revisit / fix this, since the yaml structure is also propagated into UI and can be a bit confusing.

mzazrivec: Just to illustrate my point above:

roles

mturley: Yeah, that is confusing, especially since it looks like those granular features can be turned off individually for certain roles. If the user can access the Migration Settings page but they can't use the conversion_hosts API endpoints, there will be errors there. Also, I don't do any kind of role/feature checking in the BreadcrumbSwitcher, it just renders links to all 4 pages no matter what. I wonder if that needs to change, and if so how we can do those checks from React...

mzazrivec: @mturley new product feature PR: https://github.com/ManageIQ/manageiq/pull/18877

mzazrivec: > Also, I don't do any kind of role/feature checking in the BreadcrumbSwitcher ...

Yeah, this can become a problem, should the role not have some of the migration features enabled (unlikely, but can happen).

To fix this in react, we'd have to maintain our map of `bc_item: feature` and pull enabled features via api:
```
# get user's group id first
$ /api/users/...
# get enabled product features for given group id
$ /api/groups/...?attributes=miq_product_features
```

mturley: ok, thanks for that info. I'll open a separate issue to track that, we should try to get it in for 5.11 I think.

Fryguy:
Please remove the "Red Hat" terminology from ManageIQ.  If you need pluggable product names, we have that via productization code.

Fryguy:
oh LOL...I hadn't read the whole thread yet and didn't realize the bot did that 😆

lib/manageiq/v2v/engine.rb
Fryguy: @mzazrivec @mturley Product features are now pluggable, so it could live in this repo directly.

Fryguy:
@mturley Is Migration Analytics a thing that could be used in ManageIQ (i.e. does it require the Red Hat Cloud or not?)  If it require Red Hat connectivity, then I don't believe we should add it to ManageIQ at all, and instead work in a CloudForms only plugin similar to https://github.com/RedHatCloudForms/cfme-cloud_services .  cc @gtanzillo @martinpovolny

lib/manageiq/v2v/engine.rb
mturley: @mzazrivec I opened https://github.com/ManageIQ/manageiq-v2v/issues/989 to track the breadcrumb switcher thing. @Fryguy that's good to know, we should probably move that stuff to this repo then. Is that documented somewhere? I'm not sure where that needs to live in this repo.

mturley:
@Fryguy I believe it does require the Red Hat Cloud, but in this initial disconnected case since we won't be directly connecting it to the cloud (just letting the user download a payload file that they can upload in the cloud manually) I'm not sure where we would land on that distinction. Brett mentioned in an email that "we should always be upstream first from an engineering perspective", fwiw.

Making this its own CloudForms only plugin would make sense in that case, but since this repo controls the Migration submenu, I'm not sure how we would add this new menu item there if it's not defined by this repo. Maybe we'd have to put it somewhere else. I definitely need to learn more about how those CF-only plugins work.

mturley:
cc @bthurber

Fryguy:
I think that since you can't do anything with that payload but manually upload it to the Red Hat Cloud, then even in the disconnected case, it's still a Red Hat specific value-add feature

Fryguy:
I could easily see the CloudForms specific plugin depending on this plugin, then running its initializer after the other initializer.

lib/manageiq/v2v/engine.rb
Fryguy: Pluggable product features just landed in https://github.com/ManageIQ/manageiq/pull/18806.  No documentation yet.

mturley:
That makes total sense.

> I could easily see the CloudForms specific plugin having a dependency on this plugin, then running its initializer after the other initializer.

Do you know where I should look for documentation on how to do that? I'm a little out of my realm of expertise here.

Fryguy:
> Do you know where I should look for documentation on how to do that?

It's just a matter of adding a dependency in the .gemspec, though the fact that we only have git based plugins makes that a lot harder...it can be a soft dependency in the README for now.  The more important bit is ensuring that the initializer is in the correct order.  I can help out where needed.

mturley:
Closing this because the decision has been made to move this new page to a downstream plugin.

PR #995, opened 2019-06-20
Move the Migration menu section to the top level, placed before Configuration (after Compute)

### LINKED ISSUE #994: Move Migration out of Compute to its own top-level menu
As a prerequisite to Migration Analytics, we'll need to move our whole plugin's menu out of Compute and into its own menu item. This will involve deciding on an icon for it since the top-level CloudForms menus have icons.

mturley:
Closes #994.

In anticipation of adding the new Migration Analytics page (which will now be in a separate plugin), we decided Migration should be promoted to its own top-level menu section within ManageIQ/CloudForms. This PR simply moves the menu and updates the breadcrumb bars of each page to remove the Compute piece.

Screenshot 2019-06-20 17 29 19

### DIFF
--- a/app/javascript/react/screens/App/Mappings/Mappings.js
+++ b/app/javascript/react/screens/App/Mappings/Mappings.js
@@ -1,14 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Breadcrumb, Spinner } from 'patternfly-react';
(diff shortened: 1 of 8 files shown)


mturley:
Icon choice and menu placement still need confirmation from the team, so I'm leaving `needs-design` on this for now.

PR #595, opened 2018-08-21
Reset state when leaving Plan details page

michaelkro:
Fixes #594 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1618844

planRequestFailed is being persisted between page visits, causing the
error icon to appear on the Plan Not Started page when visiting after
viewing a failed plan.

# Steps to Reproduce Bug
- Create a new migration plan, but do not run
- Visit the details page of a failed migration plan
- Visit the details page of the not started plan
- You should see the error icon in the breadcrumb area

PR #996, opened 2019-06-26
Modify rails config filters for v2v

djberg96:
This PR adds additional filtered parameters for V2V. The REST API internally uses this list inside the  `Logger#log_api_request` method where it filters the params

This partially addresses #595 in that it will filter the api.log data, but doesn't affect the core production (evm) log. That will need to be a separate PR.

Replaces https://github.com/ManageIQ/manageiq-api/pull/614

### DIFF
--- a/lib/manageiq/v2v/engine.rb
+++ b/lib/manageiq/v2v/engine.rb
@@ -10,6 +10,10 @@ module ManageIQ::V2V
       app.config.assets.paths << root.join('assets', 'images').to_s
     end
 
+    initializer 'plugin.filter' do |app|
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add_reviewer @fdupont-redhat

fdupont-redhat:
@djberg96, I'm not sure we need to hide `openstack_tls_ca_certs`. It's not sensitive, because it is a CA certificate chain, not a private key. Otherwise, it's good for me.

djberg96:
@fdupont-redhat Unless you're explicitly opposed to it, I'd like to filter it. It's still a lot of unnecessary info IMO.

fdupont-redhat:
@djberg96 you're right, it's a big string and can make the log more difficult to read.

Issue #989, opened 2019-06-14
BreadcrumbPageSwitcher does not check the user's enabled features


mturley:
The BreadcrumbPageSwitcher always contains links to plans, mappings, settings, and analytics (once that page is enabled), regardless of what is set in the feature permissions in CloudForms for the current user. We need to add some API calls to check which features the user has permission to see, and hide the breadcrumb items they don't.

Even though it's probably pretty rare for a user to have some but not all of these features enabled, this is something we should do just in case.

_Originally posted by @mzazrivec in https://github.com/ManageIQ/manageiq-v2v/pull/988_:

> > Also, I don't do any kind of role/feature checking in the BreadcrumbSwitcher ...

> Yeah, this can become a problem, should the role not have some of the migration features enabled (unlikely, but can happen).

> To fix this in react, we'd have to maintain our map of `bc_item: feature` and pull enabled features via api:

```
# get user's group id first
$ /api/users/...
# get enabled product features for given group id
$ /api/groups/...?attributes=miq_product_features
```

mzazrivec:
Actually, it seems that it's much easier to pull the list of enabled product features directly from `/api` call itself (found under `identity.miq_groups`):
```
$ curl http://evil-user:smartvm@localhost:3000/api | json_reformat
{
    "name": "API",
    "description": "REST API",
    "version": "4.1.0-pre",
    "versions": [
        {
            "name": "4.1.0-pre",
            "href": "http://localhost:3000/api/v4.1.0-pre"
        }
    ],
    "settings": {
        "locale": "en",
        "asynchronous_notifications": true
    },
    "identity": {
        "userid": "evil-user",
        "name": "evil-user",
        "user_href": "http://localhost:3000/api/users/2",
        "group": "evil-group",
        "group_href": "http://localhost:3000/api/groups/21",
        "role": "evil-role",
        "role_href": "http://localhost:3000/api/roles/20",
        "tenant": "My Company",
        "groups": [
            "evil-group"
        ],
        "miq_groups": [
            {
                "href": "http://localhost:3000/api/groups/21",
                "id": "21",
                "description": "evil-group",
                "group_type": "user",
                "sequence": 20,
                "created_on": "2019-02-25T15:30:52Z",
                "updated_on": "2019-02-25T15:30:52Z",
                "settings": null,
                "tenant_id": "1",
                "sui_product_features": [

                ],
                "product_features": [
                    "migration"
                ]
...
```

PR #998, opened 2019-06-28
Breadcrumb switcher: add support for product features

mzazrivec:
Breadcrumb switcher in v2v should honor product features enabled for user.

breadcrumbs-menu

breadcrumbs-after

Fixes: https://github.com/ManageIQ/manageiq-v2v/issues/989


app/javascript/react/screens/App/Overview/__test__/__snapshots__/index.test.js.snap
```diff
@@ -39,7 +41,7 @@ Object {
   "isFetchingClusters": false,
   "isFetchingDatastores": false,
   "isFetchingNetworks": false,
-  "isFetchingProviders": true,
+  "isFetchingProviders": false,
```
mturley: Hm, it's weird that this snapshot changed `isFetchingProviders` from true to false.. I don't think any of your code changed the providers reducer or any usage of it. Testing locally I see that providers still get loaded just fine when the Overview mounts, so I'll merge anyway.. but do you know why that happened @mzazrivec ?

mturley: Hmm, actually @mzazrivec for some reason if you include both reducers in the test for `Overview/index.js` this snapshot value goes back to `true`... which it should be, because the test mounts Overview and the action in componentDidMount there should change this state. Idk why adding the productFeaturesReducer there fixes it, maybe we had some silent errors in that snapshot test. But let's pull that in anyway: https://github.com/mturley/manageiq-v2v/commit/60d1ca71c33440cf4e3b0474a6037f25bb05550c

mzazrivec: You're right, I forgot to include `productFeaturesReducer`, I just added the initial state and ran `yarn test -u`, hence the thing with `isFetchingProviders`. Thanks for the fix, I cherry-picked it to this PR.

mturley:
Thanks for fixing this @mzazrivec !

PR #1006, opened 2019-07-24
[IVANCHUK] yarn: lock versions by updating yarn.lock

himdel:
Adding `yarn.lock`, see https://github.com/ManageIQ/manageiq-ui-classic/pull/5865

v2v is special because it already has one, so just updating here.

@miq-bot assign simaishi
@miq-bot add_label dependencies


himdel:
Closing, regenerating yarn.lock yields no changes, apparently the integrity entry only gets added when updating packages, which is probably not desirable given the test failure when running with a fresh yarn.lock.

PR #1008, opened 2019-07-24
Conversion Host Wizard: Set VDDK as default transformation method

### LINKED ISSUE #1002: [Conversion Host Wizard] Make VDDK the default transformation method
On the Authentication step of the Configure Conversion Host wizard, we currently have a Transformation method dropdown with no default selection, and SSH and VDDK choices in that order:

Screenshot 2019-07-09 17 56 03

Because SSH transformation is now being viewed as a less desirable backup option, we should change the order of those options so VDDK is first, and we should pre-select VDDK by default.

mturley:
Closes #1002.

This PR updates the Authentication step of the Conversion Host Wizard so that the Transformation Method field (previously having no default) now defaults to VDDK. This means the VDDK Library Path field will be visible by default when the user first visits the step.

Screenshot 2019-07-24 15 41 28

The Transformation Method field also previously had SSH before VDDK in the list, this order has been reversed.

Screenshot 2019-07-24 15 41 21

I also fixed a unit test that broke because the initial mounted/unmounted fields were changed.


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 1f668753fc8ee91f91e08c6fb3992e681affcad0
Date:   Thu Jul 25 13:12:58 2019 +0200

    Merge pull request #1008 from mturley/1002-vddk-default-transformation-method
    
    Conversion Host Wizard: Set VDDK as default transformation method
    
    (cherry picked from commit a05d872966e48930c0117be44b83d6df441a8144)
```

PR #1015, opened 2019-08-02
[#1014][BZ#1693994] Fix conversion host name not appearing on plan details after conversion host is deleted

### LINKED ISSUE #1014: [BZ#1693994][RFE] Migration plan detail popover don't show conversion host if related conversion host gets deleted
https://bugzilla.redhat.com/show_bug.cgi?id=1693994

From Fabien:

> mike, the above PR stores the conversion host in the task options hash: task.options.conversion_host_name.
> Could you please update the UI code to use it ? I think it's a pretty simple change, so I'd like to have it in 5.11.0.

mturley:
Closes #1014 
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1693994

On the Migration Plan Request Details page (reached by clicking on an in-progress or completed migration plan), we have an info-icon-popover next to the Status text which shows, among other things, the name of the conversion host associated with that VM's migration task:

Screenshot 2019-08-02 15 49 11

Previously, we were querying the task's `conversion_host` attribute and using the name stored there. However, as reported in https://bugzilla.redhat.com/show_bug.cgi?id=1693994 (https://github.com/ManageIQ/manageiq-v2v/issues/1014), if a user deletes the conversion host and then tries to view this page, the object will no longer be present and the name cannot be shown.

In https://github.com/ManageIQ/manageiq/pull/19075, an additional property was added to store the conversion host's name on the task as `task.options.conversion_host_name`, which will persist if the conversion host is deleted. The popover now uses this value by default if present, and falls back on the old behavior otherwise (for viewing those migration tasks which were created before this backend change).

### DIFF
--- a/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
+++ b/app/javascript/react/screens/App/Plan/components/PlanRequestDetailList/PlanRequestDetailList.js
@@ -49,7 +49,7 @@ class PlanRequestDetailList extends React.Component {
   };
 
   overlayTriggerClick = task => {
-    if (!this.props.conversionHosts[task.id]) {
+    if (!this.props.conversionHosts[task.id] && !task.options.conversion_host_name) {
       this.props.fetchConversionHostAction(this.props.fetchConversionHostUrl, task.id);
     }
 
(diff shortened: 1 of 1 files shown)


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 1df94caea6513bc133046aef390f176241bb7637
Date:   Mon Aug 5 12:32:31 2019 +0200

    Merge pull request #1015 from mturley/1014-fix-conversion-host-name
    
    [#1014][BZ#1693994] Fix conversion host name not appearing on plan details after conversion host is deleted
    
    (cherry picked from commit f59328922e846f9dfc9192ccb973457ac364f890)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1693994
```

PR #1027, opened 2019-09-06
Completely remove gettext catalogs

mzazrivec:
* the translations now live in core repo only
* we don't need these files in plugins at all
* also remove gettext initialization from the plugin (it's not needed now)


mturley:
Thanks for cleaning this up @mzazrivec !
If you have the time, would you mind looking at this same stuff in cfme-migration_analytics? I'm not even sure if I set up gettext correctly there in the first place anyway.

mzazrivec:
> If you have the time, would you mind looking at this same stuff in cfme-migration_analytics? I'm not even sure if I set up gettext correctly there in the first place anyway.

Yes sir, it's in my TODO list.

PR #1031, opened 2019-09-11
Convert list views to table structure to fix alignment issues

### LINKED ISSUE #993: [BZ#1716845] Completed plan list require proper alignment
Issue for tracking resolution of BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1716845

Description of problem:
Completed plan list require proper alignment

Mike comment:
This requires some further investigation... We are using the PatternFly "list view" component, which was not designed to operate like a table, it was designed to just have left side content, center content and right side content. Because we have stuffed so many things in the center section, the normal list view spacing doesn't work well for us anymore. To fix this I have a few options, none of which are great:

1. I could specify static widths for each "column" there (to use the term loosely), which would mean that as long as the widths of each row's data for each column are less than the static width, everything would be aligned. But it would not scale down to smaller screen sizes well, and if content ends up being wider, it will still not align correctly.
2. I could convert the whole list view to a table in the underlying HTML, which would require rebuilding some of the layout code here and breaking PatternFly semantics, with custom styles to make it still look like a regular list. This would be reliably aligned, but would take a lot of time and effort.
3. I could just move the most variable-width pieces to the end of the center section, such as the mapping name (which here includes the "infrastructure mapping does not exist" message). Most of the rest of the row has a static width, so this would mostly fix the problem, but if a plan has no "scheduled/completed" timestamp, that row's mapping name would not be aligned with the other mapping names. This would be easy to implement, though.

### LINKED ISSUE #1025: Warm Migration - Convert plans page list views to table structure
This is a refactor to help us prepare for the functional warm migration changes, but it affects the existing list views and can be done separately from any warm-migration-specific work.

Due to the amount of data we want to show in each list item of these views (which is about to increase further with warm migration), we will only continue to run into alignment issues like #993 unless we use HTML structure that will guarantee these columns are truly aligned as columns. The approach I am attempting is to convert the underlying nested-divs structure of these list views to `<table><tr><td>`, and then adjust the CSS so they still look like the existing PF3 List View design. That way we keep the visual structure we have, but we can have real column alignment for all the new data we're adding.

The downside here is that we may lose the mobile-friendly "collapsed" version of these list view items at smaller screen sizes, so we'll need to address that. Perhaps we can examine how the table cells collapse on mobile in PF4 tables, which have this behavior, or we can abstract away the structure and have the view degrade back to the original nested divs when on smaller breakpoints.

This will require some CSS review to make sure we're not doing anything dumb, since we'll be sort of breaking PF3 a bit here :)

mturley:
Closes #1025
Closes #993 (https://bugzilla.redhat.com/show_bug.cgi?id=1716845)

This PR introduces the new ListViewTable and ListViewTableRow components, which match the props interface for the pf3 ListView and ListViewItem, respectively. The ListViewTable converts the floating-divs HTML structure of the ListView into table HTML, to properly align its items as real columns.

Because this breaks responsiveness on mobile viewport sizes, it is only used at desktop window size, and smaller windows will fall back to the original ListView implementation (which is possible because of the matching props interface). The BreakJS-based layout listener (which I wouldn't have added, but is already used elsewhere in pf3) swaps the two render implementations out automatically on window resize.

This PR converts the Plans Not Started and Plans Complete/Archived list views to use the new structure. In separate PRs, I'll convert the Plan Details list view, and also convert the Plans In Progress card view to this kind of list view.

I also worked out some CSS quirks, and in general slimmed down some of the margin/padding in these list view items, in anticipation of the higher-density list view items we'll need for warm migration.

# Screens

Before:

Screenshot 2019-09-11 17 06 45

After:

Screenshot 2019-09-11 17 05 50


mturley:
Fixed the CI failures here, I needed to update enzyme so it could understand the Context API.

mturley:
I think we can ignore the codeclimate errors here since they are out of the scope of fixing this bug.

mturley:
I pushed another commit to abstract the scheduled time info item into a component shared by the not started list and the completed list, to satisfy codeclimate. But, codeclimate now has a problem with the propTypes being too similar to other propTypes lists, I guess? Except the lists don't even have the same props. I think this is invalid

Reading up on codeclimate, it looks like it's possible to mark certain issues as `invalid` or `wontfix` (https://docs.codeclimate.com/docs/issues) which would be really useful for getting rid of that red X in these cases without totally ignoring codeclimate. @Fryguy @agrare do you guys know if that's something I can request permission for in codeclimate, and if so, who manages those permissions? Currently I don't have permission to set the issue statuses described in that doc.

app/javascript/react/screens/App/common/ListViewTable/ListViewTableRow.js
```diff
@@ -0,0 +1,69 @@
...
+        ))}
+      {actions && (
+        <td>
+          <ListView.Actions>{actions}</ListView.Actions>
+        </td>
+      )}
+    </tr>
+  );
+  const numColumns = row.props.children.filter(child => !!child).length;
+  console.log('TODO: use this for colspan of expanded content -- numColumns =', numColumns);
```
mzazrivec: Do we need the `console.log` here?

mzazrivec:
The changes look good to me and have no problem merging, except that one `console.log` (which may stay, will wait for Mike's opinion here).

app/javascript/react/screens/App/common/ListViewTable/ListViewTableRow.js
mturley: Ah, good catch, I forgot about that. I was intending originally to implement expandable rows in this PR, but I decided to save that for another PR and forgot to remove this.

mturley:
I staged this locally on ivanchuk and it shouldn't have merge conflicts.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit b2f23976700ffccf0f89f2218b3ae6e0180422c0
Date:   Thu Sep 19 19:59:51 2019 +0200

    Merge pull request #1031 from mturley/list-view-table-structure

    Convert list views to table structure to fix alignment issues

    (cherry picked from commit 40746f19d02a7c407c5d101accbd733766fa9d11)

    https://bugzilla.redhat.com/show_bug.cgi?id=1716845
```

Issue #1032, opened 2019-09-12
[BZ#1677677] Fatal Error in Mapping Wizard: When mapping more than one source cluster, cannot switch between clusters on storage step


mturley:
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1677677

I observed this bug while testing #1030, but it also appears to be present on master.

1. Open the Create Infrastructure Mapping wizard.
2. On the "Map Compute" step, add mappings for at least two different source clusters.
3. Proceed to the "Map Storage" step, and try to select a cluster in the dropdown.
4. An error appears, the cluster is not selected, and you cannot proceed.

![alUZOvWrVR](https://user-images.githubusercontent.com/811963/64818494-2d0c0680-d57a-11e9-8bae-f71996f27108.gif)

<img width="671" alt="Screenshot 2019-09-12 16 24 55" src="https://user-images.githubusercontent.com/811963/64818502-3006f700-d57a-11e9-87f6-e483667d117d.png">

The line that is causing the error is here: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js#L109
Apparently, there is an issue with the reference to `this.setState` being set as a default parameter value in the function definition, which is causing React to get confused about `this` internally.

The fix is pretty simple (use this.setState directly in the method body instead of assigning it and calling it by reference), but (a) I don't really understand why it doesn't work as written, and (b) I don't know why we're only seeing this now, I'm pretty sure it was working fine when that code was introduced in https://github.com/ManageIQ/manageiq-v2v/pull/796.

@himdel, @mzazrivec , @Hyperkid123 .. would you guys mind helping me figure out what's going on here and identify which ManageIQ releases are affected? I imagine we should also make sure QE is testing this wizard with more than one source cluster.. I'm surprised we didn't catch it sooner. I'll post a PR shortly to fix the bug, but I'm not sure how to proceed with damage control here.

PR #1033, opened 2019-09-12
Fix error when selecting multiple source clusters in an infrastructure mapping

mturley:
Fixes https://github.com/ManageIQ/manageiq-v2v/issues/1032.
Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1677677

The bug was also present on the Networks step of the wizard, which duplicates the problematic code (the more I read through this old mapping wizard code the more I see how rushed it was... we should probably clean up a lot of this code duplication). This PR fixes the bug in both places.

As mentioned in #1032, I'm not sure why this code ever worked and when/why the bug was introduced. I am still investigating which versions of ManageIQ are affected, and would appreciate help with this :)

### DIFF
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
@@ -27,8 +27,8 @@ class MappingWizardDatastoresStep extends React.Component {
       []
     );
 
-    const synchronousSetState = fn => {
-      this.state = { ...this.state, ...fn() };
+    const synchronousSetState = newState => {
+      this.state = { ...this.state, ...newState };
     };
 
     if (sourceClusters.length === 1 || !pristine) {
@@ -89,7 +89,7 @@ class MappingWizardDatastoresStep extends React.Component {
     }
   }
 
-  selectSourceCluster = (sourceClusterId, setState = this.setState) => {
+  selectSourceCluster = (sourceClusterId, synchronousSetState) => {
     // when dropdown selection occurs for source cluster, we go retrieve the datastores for that
     // cluster
     const {
@@ -106,10 +106,16 @@ class MappingWizardDatastoresStep extends React.Component {
 
     const { nodes: sourceClusters, ...targetCluster } = selectedClusterMapping;
 
-    setState(() => ({
+    const newState = {
       selectedCluster: sourceClusters.find(sourceCluster => sourceCluster.id === sourceClusterId),
       selectedClusterMapping
-    }));
+    };
+
+    if (synchronousSetState) {
+      synchronousSetState(newState);
+    } else {
+      this.setState(newState);
+    }
 
     fetchSourceDatastoresAction(fetchStoragesUrls.source, sourceClusterId);
     fetchTargetDatastoresAction(fetchStoragesUrls[targetProvider], targetCluster.id, targetProvider);
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardNetworksStep/MappingWizardNetworksStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardNetworksStep/MappingWizardNetworksStep.js
@@ -27,8 +27,8 @@ class MappingWizardNetworksStep extends React.Component {
       []
     );
 
-    const synchronousSetState = fn => {
-      this.state = { ...this.state, ...fn() };
+    const synchronousSetState = newState => {
+      this.state = { ...this.state, ...newState };
     };
 
     if (sourceClusters.length === 1 || !pristine) {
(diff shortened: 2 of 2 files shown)


Hyperkid123:
@mturley i was looking at this issues and the main problem i see with assigning setState as a default argument. This fixed it for me:
```diff
diff --git a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
index 17e37243..14a93824 100644
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardDatastoresStep/MappingWizardDatastoresStep.js
@@ -89,7 +89,7 @@ class MappingWizardDatastoresStep extends React.Component {
     }
   }
 
-  selectSourceCluster = (sourceClusterId, setState = this.setState) => {
+  selectSourceCluster = sourceClusterId => {
     // when dropdown selection occurs for source cluster, we go retrieve the datastores for that
     // cluster
     const {
@@ -106,7 +106,7 @@ class MappingWizardDatastoresStep extends React.Component {
 
     const { nodes: sourceClusters, ...targetCluster } = selectedClusterMapping;
 
-    setState(() => ({
+    this.setState(() => ({
       selectedCluster: sourceClusters.find(sourceCluster => sourceCluster.id === sourceClusterId),
       selectedClusterMapping
     }));
```

Hyperkid123:
But looking at the code more closely I think there should some effort put towards rewriting these very complex class base components with a lot of side effect to functional components with hooks. Doing this might significantly simplify the code base.

mturley:
@Hyperkid123 the problem with that fix is that this method is also (conditionally) used in the constructor, where we can't call this.setState, hence the alternative synchronousSetState it passes in to just make the initial selection part of the initialization of this.state. It's not pretty, it was a workaround for an issue a while back.

I agree that many of these complex components should be rewritten. There's a whole lot of technical debt here. Hooks would simplify a lot of it, but there's also a bunch of duplicated and convoluted logic. Much of this code was written (in a hurry, for a POC) with only RHV migration targets in mind, then modified (in a hurry) to also support OSP targets. I think most of these are due for a total rewrite, reconsidering their design from the ground up, and not just converting the logic we have to a hooks-based equivalent. Although that might be a good start.

Unfortunately the project has been woefully understaffed, it's basically just me and @mzazrivec now and we both have other things to work on too :) I hope we find time to get this all cleaned up, but I don't know if it will be any time soon :(

Sorry for the rant, it's not directed at you, this code just makes me sad sometimes. Thanks for taking a look :slightly_smiling:

Hyperkid123:
> @Hyperkid123 the problem with that fix is that this method is also (conditionally) used in the constructor, where we can't call this.setState

I was not aware of this.

> Sorry for the rant, it's not directed at you, this code just makes me sad sometimes. Thanks for taking a look 🙂

No worries 👍. I completely understand. It gave me a few headaches too in the past when i was doing package upgrades. 

> Unfortunately the project has been woefully understaffed, it's basically just me and @mzazrivec now and we both have other things to work on too :)

I am aware of this situation and i understand it very well. @martinpovolny do you think it would be possible to assign some of our interns to do some light refactoring when they have bit of down time (waiting for backend/api work)? I think it would be great learning experience for them and it might help @mturley and @mzazrivec a bit.

mturley:
@Hyperkid123 that would be awesome :) it might be tricky to unravel some of the logic and make sure they're not breaking anything, though. I'd be happy to help guide them if/when that happens.

@mzazrivec are you ok with merging this PR or should we try a different fix for the short term? I'm a bit embarrassed that I've let this bug sit unfixed for so long since I misunderstood the original BZ.

martinpovolny:
I am not going to prevent anyone from helping.

However:

> when they have bit of down time

That's the "funny" part.

Our backlog is looooooong. There's a lot of work needed on the UI side.

mturley:
@martinpovolny I totally understand that. Nobody will have time to do this refactoring any time soon, and I think that's ok. It's at the end of my backlog, too 😄

mturley:
New BZ broken out for only this specific bug: https://bugzilla.redhat.com/show_bug.cgi?id=1677677

mturley:
I think we can ignore the codeclimate errors here since they are out of the scope of fixing this bug.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 508186e81150b8f5904a160bd6325f21c6a4ec01
Date:   Wed Sep 18 15:18:08 2019 +0200

    Merge pull request #1033 from mturley/fix-1032

    Fix error when selecting multiple source clusters in an infrastructure mapping

    (cherry picked from commit c63daa17a300d6a19d8c61cfe883e8cb496c3815)

    https://bugzilla.redhat.com/show_bug.cgi?id=1677677
```

PR #1049, opened 2019-10-10
[#1046] Fix logic for determining if a migration task is completed successfully

### LINKED ISSUE #1046: An error icon can show on a successful migration in the plan details list view
This issue is coming up with the latest version of ManageIQ, because of this code:

https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/helpers.js#L78
```js
    taskDetails.completedSuccessfully =
        (task.options.progress && task.options.progress.current_description === 'Virtual machine migrated') ||
        (task.options.progress && task.options.progress.current_description === 'Mark source as migrated');
```

On a list item in the plan request details list, if the task is complete and its `current_description` is not one of those two magic values, there will be an error icon on the left of that list item. We need to instead check `task.state === 'finished' && task.status === 'Ok'`, which we already use for determining a successful migration elsewhere in the code (we should just reuse the `taskCompletedSuccessfully` helper at the top of that same file: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/helpers.js#L12)

This bug can be reproduced using the database dump named `status-message-issues` from [my collection](https://drive.google.com/drive/u/1/folders/1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q).

mturley:
Fix #1046 (related to https://bugzilla.redhat.com/show_bug.cgi?id=1759062).

When we originally developed the plan details page, we wrote a lot of logic based on specific magic strings in the `task.message` property instead of using the more reliable `state` and `status` properties. In the latest version of ManageIQ, some of these task messages have changed, so the logic is failing in this case. This PR switches the logic for the `taskDetails.completedSuccessfully` property to use `state` and `status`, which fixes this icon bug.

The bug can be reproduced using the database dump named `status-message-issues` from [my collection](https://drive.google.com/drive/u/1/folders/1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q). Specifically, if you look at `/migration#/plan/4`, you can see that the migration was successful (and the plan itself has a green success icon) but the individual VM list item shows an error icon:

Screenshot 2019-10-10 11 52 35

After this fix:

Screenshot 2019-10-10 11 53 19

### DIFF
--- a/app/javascript/react/screens/App/Plan/helpers.js
+++ b/app/javascript/react/screens/App/Plan/helpers.js
@@ -1,7 +1,6 @@
 import numeral from 'numeral';
 
 import {
-  STATUS_MESSAGE_KEYS,
   V2V_MIGRATION_STATUS_MESSAGES,
   V2V_DOWNLOAD_LOG_STATUS_MESSAGES,
   V2V_BACKEND_ERROR_MESSAGES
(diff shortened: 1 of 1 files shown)


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit f0faf96c91406f8985bc5bbf3f843b0f4b829352
Date:   Fri Oct 11 12:58:05 2019 +0200

    Merge pull request #1049 from mturley/1046-error-icon-logic

    [#1046] Fix logic for determining if a migration task is completed successfully

    (cherry picked from commit 158e4314279d9bdb2b1eb28d48108495c8411fb1)

    https://bugzilla.redhat.com/show_bug.cgi?id=1789433
```

Issue #1054, opened 2019-10-23
Migration Status messages need to be updated to match the new ones in the backend


mturley:
The UI keeps a list of strings that can be present in various properties for migration status. Since these messages are changing in 5.11.1, the UI strings need to be updated so translations will remain correct.

The list of messages in the UI is here: https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/PlanConstants.js#L22

The list of messages in the backend is here: https://github.com/ManageIQ/manageiq/blob/master/app/models/infra_conversion_job.rb#L76-L128

Related to BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1759062

mturley:
Note: after a brief chat with @mzazrivec about this, it sounds like the better solution would be to have these messages translated on the backend before the UI loads them, so the UI can just display them as-is. That sounds like an ideal way to resolve this to me :)

vconzola:
Resolved in PR #1076. Close.

PR #1055, opened 2019-10-23
[#1045][BZ#1759062] Rearrange migration status messages to expose more descriptive information

### LINKED ISSUE #1045: Issues with migration status messages
Still working with @fdupont-redhat to determine the exact fixes here, but what we're observing:

In places where a status message is looked up in an i18n table (https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/Plan/PlanConstants.js#L22, which is used for `task.message` and `task.options.progress.current_description`), if the string coming from the API is not present in the list of expected messages, we will get one of the following instead of the actual message:
* the string "VM migrations stalled" (probably put in place to prevent long error strings from breaking the layout)
* a blank string.

In either case, we should:
* stop hiding unexpected strings, and figure out the right way to display them (maybe if it could be too long it should go in a popover, or maybe we just display it as-is inline)
* update the list of expected strings to include any new ones.

It also seems like the `task.options.progress` strings we use may need to instead reference `task.options` directly instead of the `progress` hash, still discussing that with @fdupont-redhat .

This bug (the blank string case) can be reproduced using the database dump named `status-message-issues` from [my collection](https://drive.google.com/drive/u/1/folders/1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q).

Related to BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1759062

mturley:
Fixes #1045.
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1759062.

Before this PR, the list view had three status messages:
* On the first line of each list view item: `task.message`
* On the second line, with a label of "Status": `task.options.progress.current_description`
* Inside the info popover, with a label of "Status Detail": `task.options.progress.states[task.options.progress.current_state].message`

Based on discussions with @vconzola and @fdupont-redhat, this PR:
* Drops `task.message` since it was not very meaningful and would be redundant with new values for the other messages
* Moves `task.options.progress.current_description` up to the first line, dropping the "Status:" label
* Moves `task.options.progress.states[task.options.progress.current_state].message` out of the popover and into the second line.

It also fixes some issues:
* #1045, an unexpected message string would result in either a blank message or the message "VM migrations stalled" being displayed, which this PR removes entirely.
* There was some logic which depended on checking for the `task.message === "Pending"`, after `task.message` had i18n applied to it, so this would have failed on non-English-language systems. This logic was only for displaying the "pending" icon vs. the spinner, though, so it's not a big deal.

Note: these strings will need to be updated when @fdupont-redhat finalizes the messages in the backend. I opened https://github.com/ManageIQ/manageiq-v2v/issues/1054 to track this.

# Screens

## Before
Screenshot 2019-10-23 16 50 38
Screenshot 2019-10-23 16 50 51

## After
Screenshot 2019-10-23 18 06 35
Screenshot 2019-10-23 18 07 10

Note: ignore the broken column alignment in the After screenshots; in my next PR this list view is being converted to a ListViewTable which will have the correct alignment.


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 7902375044edfef502ac643764bf5c38262ac00a
Date:   Thu Oct 24 13:09:54 2019 +0200

    Merge pull request #1055 from mturley/1045-fix-status-messages

    [#1045][BZ#1759062] Rearrange migration status messages to expose more descriptive information

    (cherry picked from commit 2afe37a45bc937fcb681f3742c20e4aac8cdd6e0)

    https://bugzilla.redhat.com/show_bug.cgi?id=1789433
```

PR #1057, opened 2019-10-25
Add expandable row support to ListViewTable and convert PlanRequestDetailList into a ListViewTable

mturley:
Part of preparations for Warm Migration.

This PR enhances the ListViewTable component to handle more of the props handled by the original ListView, namely `children` and `checkboxInput`. It also converts the PlanRequestDetailList to use a ListViewTable (fixing column layout issues in that view). If the `ListViewTableRow`s in this view are given `children`, they will become expandable and the children will be rendered in the expanded content area.

Because the warm-migration-specific expanded content for these rows is still in development, this PR's last commit removes the placeholder `children`, disabling the expandable rows. To test the expandable row functionality, you can revert the last commit (https://github.com/ManageIQ/manageiq-v2v/pull/1057/commits/63fd1ea343caec5ee3b634b04c77c07cc4f90b8c).

Note that some funky CSS overrides were necessary to preserve expanded row styles, and we'll probably want to revisit the CSS later on with some other UXD team members. This part took me way too damn long.

# Screens

Plan details view after this PR (only visible change is the fixed column alignment):
Screenshot 2019-10-25 15 08 31

Plan details view with https://github.com/ManageIQ/manageiq-v2v/pull/1057/commits/63fd1ea343caec5ee3b634b04c77c07cc4f90b8c reverted to enable expandable content:
Screenshot 2019-10-25 15 07 06

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MappingNameInfoItem.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MappingNameInfoItem.js
@@ -1,15 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { ListView, Icon } from 'patternfly-react';
+import { Icon } from 'patternfly-react';
+import ListViewTable from '../../../common/ListViewTable/ListViewTable';
 
 const MappingNameInfoItem = ({ plan }) => (
-  <ListView.InfoItem>
+  <ListViewTable.InfoItem>
     {plan.infraMappingName || (
       <React.Fragment>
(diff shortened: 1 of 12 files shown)


mturley:
Argh. The borders don't look right on Firefox. Fixing that...

mturley:
Ok, fixed. Should be good now.

mturley:
This will need to be rebased on https://github.com/ManageIQ/manageiq-v2v/pull/1066 to fix the CI failures.

PR #1059, opened 2019-10-29
[#1040][#1743703] Don't display invalid conversion hosts

### LINKED ISSUE #1040: [Part of BZ#1743703] Hide invalid conversion hosts in the UI
This is a temporary fix related to the larger BZ about handling related resources when a provider is refreshed or deleted: https://bugzilla.redhat.com/show_bug.cgi?id=1743703

This issue covers the temporary step mentioned in that BZ, we should hide conversion hosts which have a `resource.ems_id` of `null`. This will prevent a user from being able to trigger the "remove" playbook on a conversion host where the actual host/vm can no longer be found, which would cause errors.

mturley:
Fix #1040 
Related to BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1743703


mturley:
This will need to be rebased on https://github.com/ManageIQ/manageiq-v2v/pull/1066 to fix the CI failures.

PR #1070, opened 2019-11-25
Schedule migration modal changes for warm migration

mzazrivec:
Schedule & Migrate buttons got merged into one
warm-scheduling-01

Schedule migration now
warm-scheduling-02

Schedule migration later
warm-scheduling-03

Schedule retry button
warm-scheduling-04

Schedule retry modal
warm-scheduling-05

Schedule cutover button
warm-scheduling-06

Schedule cutover modal
warm-scheduling-07

### DIFF
--- a/app/javascript/react/screens/App/Overview/components/Migrations/MigrationInProgressListItem.js
+++ b/app/javascript/react/screens/App/Overview/components/Migrations/MigrationInProgressListItem.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Spinner, Button, Icon, UtilizationBar } from 'patternfly-react';
+import { Spinner, Button, Icon, UtilizationBar, DropdownKebab, MenuItem } from 'patternfly-react';
 import EllipsisWithTooltip from 'react-ellipsis-with-tooltip';
 
 import ListViewTable from '../../../common/ListViewTable/ListViewTable';
@@ -18,6 +18,9 @@ import {
 } from './helpers/inProgressHelpers';
 import InProgressRow from './InProgressRow';
 import ProgressBarTooltip from './ProgressBarTooltip';
+import ScheduleMigrationButton from './ScheduleMigrationButton';
+import CutoverTimeInfoItem from './CutoverTimeInfoItem';
+import StopPropagationOnClick from '../../../common/StopPropagationOnClick';
 
 const MigrationInProgressListItem = ({
   plan,
@@ -34,12 +37,19 @@ const MigrationInProgressListItem = ({
   setMigrationsFilterAction,
   cancelPlanRequestAction,
   isCancellingPlanRequest,
-  requestsProcessingCancellation
+  requestsProcessingCancellation,
+  loading,
+  toggleScheduleMigrationModal
 }) => {
   const { requestsOfAssociatedPlan, mostRecentRequest } = getRequestsOfPlan({ plan, allRequestsWithTasks });
   const waitingForConversionHost = isWaitingForConversionHost(mostRecentRequest);
 
   const isInitiating = reloadCard || !mostRecentRequest || mostRecentRequest.request_state === 'pending';
+
+  const showScheduleMigrationButton =
+    plan.options.config_info.warm_migration && !plan.options.config_info.warm_migration_cutover_datetime;
+  const showWarmMigrationKebab =
+    plan.options.config_info.warm_migration && plan.options.config_info.warm_migration_cutover_datetime;
 
   // Plan request state: initiating
   if (isInitiating) {
@@ -50,7 +60,8 @@ const MigrationInProgressListItem = ({
           <ListViewTable.InfoItem key="initiating">
             <Spinner size="sm" inline loading />
             {__('Initiating migration. This might take a few minutes.')}
-          </ListViewTable.InfoItem>
+          </ListViewTable.InfoItem>,
+          <CutoverTimeInfoItem plan={plan} />
         ]}
       />
     );
@@ -81,7 +92,8 @@ const MigrationInProgressListItem = ({
             <EllipsisWithTooltip style={{ maxWidth: 300 }}>
               {__('Waiting for an available conversion host. You can continue waiting or go to the Migration Settings page to increase the number of migrations per host.') /* prettier-ignore */}
             </EllipsisWithTooltip>
-          </ListViewTable.InfoItem>
+          </ListViewTable.InfoItem>,
+          <CutoverTimeInfoItem plan={plan} />
         ]}
         actions={
           <Button disabled={cancelButtonDisabled} onClick={onCancelClick}>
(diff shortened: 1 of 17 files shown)


mturley:
Hey @mzazrivec, I hope you enjoyed your holidays.

When you have a chance I think we should figure out how to get this and https://github.com/ManageIQ/manageiq-v2v/pull/1077 merged, I'm sure there will be conflicts and I think it'll be a mess if we try to resolve them at the last minute when the backend is ready. Considering that warm migration is already "broken" on master (you can select it in the wizard and it doesn't work properly) I think maybe we can justify merging this stuff before we test it with the backend, and then consider any issues that come up as separate bugs to be fixed when the backend is testable... does this sound reasonable or do you think we should instead merge it with flags in place to disable the new behavior?

mzazrivec:
@mturley I'd say we should try to merge our work now and resolve whatever comes up when needed. Enable / disable flags for this feature? Probably not. It would just make things more complicated.

I still need to address the tests here before this PR can be merged.

mturley:
@mzazrivec Ok sounds good. It sounds like Fabien and Martin are hoping to have the warm migration backend changes done this week and then want to get all the UI merged for the build on Monday... I'm hoping they leave us a day or two to test everything end to end but it might be a crunch 😞  I also need to finish writing unit tests for #1077 and #1082. I was thinking about asking if you are ok with merging this stuff without tests and then adding tests in another PR, but if we can get the conflicts resolved soon I guess that's not necessary. Thanks

mzazrivec:
@mturley I'm quite OK with merging the PRs without unit tests as long as the existing tests pass.

mzazrivec:
@mturley I don't have tests for this PR in place (can add in a separate PR), but I think it's ready for review.

mturley:
Sounds good @mzazrivec, I think we can get as much merged for Monday's build as possible and then follow up with unit tests. I made a new label `needs-tests` to keep track of which PRs we need to return to.

mturley:
Well, I guess Monday's build of ivanchuk won't have any of this stuff until it's all ready, so that might not matter depending on the backend.

app/javascript/react/screens/App/Overview/components/ScheduleMigrationModal/ScheduleMigrationModal.js
```diff
@@ -1,11 +1,22 @@
...
 import ScheduleMigrationModalBody from './ScheduleMigrationModalBody';
 import getPlanScheduleInfo from '../Migrations/helpers/getPlanScheduleInfo';
 
 class ScheduleMigrationModal extends React.Component {
-  state = { dateTimeInput: '' };
+  constructor(props) {
+    super(props);
+
+    const { scheduleMigrationPlan } = this.props;
+    const { migrationScheduled } = getPlanScheduleInfo(scheduleMigrationPlan) || '';
```
mturley: I think we need to look at the `getPlanScheduleInfo` helper and make sure it is aware of warm migrations, because currently the `migrationScheduled` value here will always be the cold-migration scheduled time. So if the user already scheduled a warm migration cutover and they open the modal again to edit that schedule, the state won't pre-populate with their cutover time.

app/javascript/react/screens/App/Overview/components/ScheduleMigrationModal/ScheduleMigrationModalBody.js
```diff
@@ -3,54 +3,117 @@ import { Icon } from 'patternfly-react';
 import PropTypes from 'prop-types';
 
 class ScheduleMigrationModalBody extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showDatepicker: false
```
mturley: This `showDatepicker` state value seems like it's just duplicating the `startMigrationNow` value from the parent component, since you're having to keep them in sync. It might be better to just have `showDatepicker` be a prop, and pass in `showDatepicker={!startMigrationNow}` from the parent so there's a single source of truth for that.

That way, in `componentDidUpdate` you can just detect the prop changing and use that to call the datetimepicker init stuff, and you won't have to call the `startMigrationNowHandler` in there or in `componentDidMount`. You can instead just call that handler in your `handleRadioChange` callback to update the lifted state.

Also `startMigrationNowHandler` is a bit of a confusing name, at first I thought calling it would start the migration. Maybe `setScheduleMode` or something like that would be more clear (I don't feel strongly about that though).

mturley: I guess you'd still have to make sure you call the datepicker init stuff in componentDidMount if `startMigrationNow` is already false when this mounts, but you can probably do that by moving that stuff to a helper method that you call in both componentDidMount and componentDidUpdate.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationInProgressListItem.js
```diff
@@ -162,6 +165,16 @@ const MigrationInProgressListItem = ({
             {sprintf(__('Running playbook service %s. This might take a few minutes.'), playbookName)}
           </ListViewTable.InfoItem>
         ]}
+        actions={
+          <div>
+            <ScheduleMigrationButton
```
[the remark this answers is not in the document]
mzazrivec: Yup, fixed.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationInProgressListItem.js
```diff
@@ -197,6 +210,16 @@ const MigrationInProgressListItem = ({
           </div>
         </ListViewTable.InfoItem>
       ]}
+      actions={
+        <div>
+          <ScheduleMigrationButton
```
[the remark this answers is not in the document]
mzazrivec: Fixed as well.

app/javascript/react/screens/App/Overview/components/ScheduleMigrationModal/ScheduleMigrationModal.js
mzazrivec: Yes, added `scheduleCutover` to the returned object.

app/javascript/react/screens/App/Overview/components/ScheduleMigrationModal/ScheduleMigrationModalBody.js
mzazrivec: Yup, fixed too.

app/javascript/react/screens/App/Overview/components/Migrations/MigrationInProgressListItem.js
```diff
@@ -34,13 +37,20 @@ const MigrationInProgressListItem = ({
...
+  loading,
+  toggleScheduleMigrationModal
 }) => {
   const { requestsOfAssociatedPlan, mostRecentRequest } = getRequestsOfPlan({ plan, allRequestsWithTasks });
   const waitingForConversionHost = isWaitingForConversionHost(mostRecentRequest);
 
   const isInitiating = reloadCard || !mostRecentRequest || mostRecentRequest.request_state === 'pending';
 
+  const showScheduleMigrationButton =
+    plan.options.config_info.warm_migration && plan.options.config_info.warm_migration_cutover_datetime === null;
```
mturley: Didn't notice this before, but I think the `=== null` check here is too strict, since a plan will initially not have a `warm_migration_cutover_datetime` defined. We should probably just have this check for a falsy value.

mzazrivec: Fixed (I think :smile:).

PR #1075, opened 2019-12-03
Add popup for warm migrations

mzazrivec:
In case some of the selected VMs can't be pre-copied, the plan creation wizard would disable the warm migration option and render an icon with a popover text.

Warm migration is possible:
warm-wizard-01

Warm migration is not possible (with the warm radio button disabled):
warm-wizard-02

Popup:
warm-wizard-03

### DIFF
--- a/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardScheduleStep/PlanWizardScheduleStep.js
+++ b/app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardScheduleStep/PlanWizardScheduleStep.js
@@ -4,7 +4,12 @@ import { Field, reduxForm } from 'redux-form';
 import { Form, Icon } from 'patternfly-react';
 import { FormField } from '../../../../../common/forms/FormField';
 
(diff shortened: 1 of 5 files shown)


app/javascript/react/screens/App/Overview/screens/PlanWizard/components/PlanWizardScheduleStep/PlanWizardScheduleStep.js
```diff
@@ -13,14 +18,20 @@ export const PlanWizardScheduleStep = ({ targetProvider, migration_plan_type_rad
       label={__('Migration Type')}
       inline_label={false}
       controlWidth={12}
+      info={
```
mturley: This info icon should instead be a warning icon (https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/common/forms/FormField.js#L92 should render `name="warning-triangle-o"` on that `Icon`), so we probably need to add a prop like `infoIconName` to `FormField` that defaults to `'info'` but can be overridden here.

mzazrivec: Yes sir, fixed & updated the screenshots as well.

PR #1076, opened 2019-12-05
[#1054] Update migration status message map to match latest backend strings

mturley:
Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1759062

This doesn't finish off https://github.com/ManageIQ/manageiq-v2v/issues/1054 because we want to revisit this issue by performing these translations in the backend, but this PR is a stopgap to  unblock the BZ so regression fixes can go out.

I modified this message map defensively, leaving the old versions of messages in place just in case they still show up. If one of you guys can help me confidently determine what will and will not appear in the relevant API properties (`task.message`, `task.options.progress.current_description`, `task.options.progress.states[task.options.progress.current_state].message`), we can pare down this map to only the messages that are possible.

I'm not sure if these strings are final; some messages are in imperative tense and some aren't (e.g. "Restore" vs "Restoring", "Convert" vs "Converting"). I got these strings from https://github.com/ManageIQ/manageiq/blob/master/app/models/infra_conversion_job.rb#L76-L128. @fdupont-redhat can you confirm that these are the latest strings from the backend? Am I looking in the right place?

Note: I also fix a small bug here in PlanRequestDetailListConstants which was causing the wrong message to appear for the "VM migrations failed" filter option. This was introduced in https://github.com/ManageIQ/manageiq-v2v/pull/1055 which hasn't been backported/released yet.


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 4238a72b6eed9ee02ec4f578700009f2034163f2
Date:   Fri Dec 6 11:25:34 2019 +0100

    Merge pull request #1076 from mturley/1054-status-message-updates

    [#1054] Update migration status message map to match latest backend strings

    (cherry picked from commit e7eaa28723349161b4ff73a61a0038244a89e5ca)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1789433
```

PR #1082, opened 2020-01-03
Check for VDDK conversion hosts before allowing warm migration

mturley:
On the Schedule step of the Migration Plan wizard, this PR replaces the warning icon (for VMs without snapshots) with two inline toast alerts, one for each of the potential reasons why warm migration might be unsupported. One or both will appear on mount of that step and disappear on unmount, if warm migration is blocked. The new alert appears if there are no VDDK-configured conversion hosts associated with the selected VMs.

In order to support more than one alert message at a time in the wizard, I changed the way we handle those alerts in Redux. The `showAlertAction` now takes an object of shape `{ alertText, alertType, alertId }` (if no alertId is passed, all other alerts will be hidden and only the new one will show, matching the old behavior). The `hideAlertAction` now takes an optional `alertId` argument (if no alertId is passed, all alerts will be hidden). I made these changes only in the Plan wizard, but we also had the same `alertText` redux patterns duplicated in the mapping wizard, so if we ever need concurrent alerts there we can move this stuff to somewhere common.

Screenshot 2020-01-07 17 06 55

I updated/fixed the existing tests that were failing, but I'll add tests for the new code in a followup PR.


mzazrivec:
The current CI failures are related.

mturley:
That's weird... prettier on my local env and on CI aren't agreeing. I must have screwed something up.

mturley:
@mzazrivec are you able to run `yarn test` without failures on master? It fails for me...

mturley:
I think it comes down to some version changes that happened when we removed `yarn.lock` in https://github.com/ManageIQ/manageiq-v2v/pull/1079, I guess prettier changed their rules. I'll make a PR to explicitly change back to the versions we were using..

mzazrivec:
@mturley Even with https://github.com/ManageIQ/manageiq-v2v/pull/1079 now in place, I see `yarn lint` failing (localy and on travis too).

mzazrivec:
@mturley after I run `yarn lint:fix` localy, `yarn test` still fails on my end, so 🤷 ...

mturley:
As for `yarn lint` failing, that must have been my editor using the newer prettier rules before we switched back in #1079, fixed. The other failures were because I hadn't run/fixed the tests yet, I've now fixed existing tests. I guess let's go ahead and get this merged and I'll add the new tests in another PR.

PR #1083, opened 2020-01-06
Roll back to previously locked versions of packages that are causing test failures

mturley:
When @mzazrivec mentioned the CI failures in https://github.com/ManageIQ/manageiq-v2v/pull/1082#issuecomment-571088866 I noticed many of them were unrelated to the code changes in that PR. I realized that since we removed `yarn.lock` in https://github.com/ManageIQ/manageiq-v2v/pull/1079 and we use the `^` version approximation syntax in our package.json, we allowed some packages to be upgraded on a fresh install. I verified this by checking out the master branch, removing my local yarn.lock, re-running `yarn install` and then running `yarn test`, which failed.

The failures were related to:
* changes to lint rules due to updated versions of `prettier` and `eslint-plugin-react`
* unexpected new props in snapshots due to an updated version of `react-ellipsis-with-tooltip`
* errors in certain unit tests which create redux stores without all the reducers matching the properties passed into the store, which is a check introduced in an updated version of `redux`

This PR reverts those packages to the specific versions that were locked in the `yarn.lock` file we removed, by removing the `^` and just using version numbers alone. The tests pass now.

We should possibly eventually upgrade these packages and make the code changes necessary to make the tests pass with the new versions, but I didn't think our CI should fail just because something outside the repo changed. We can make those changes when we're ready to.

I'm also trusting (hoping) that all the other packages we are now pulling in newer versions of are honoring their semantic versioning and not releasing any breaking changes that we'll get with this package.json, and that this kind of thing will only break unit tests that assume the lack of new features. I really hope I'm right about that.. maybe we should be avoiding the `^` syntax entirely.

This will unblock all other open PRs, which should be rebased on these changes.


mzazrivec:
I'm confirming that the specific package versions now make `yarn prettier` and `yarn test` pass.

Merging.

himdel:
:+1: I commented in https://github.com/ManageIQ/manageiq-v2v/pull/1079#issuecomment-571580747,
but I think the problem is that `^` is simply too optimistic and `~` usually does the right thing.

PR #1086, opened 2020-01-10
[#1072] [BZ#1775998] Fix logic for providers check on Conversion Hosts Settings page

### LINKED ISSUE #1072: [BZ#1775998] Inaccurate message is displayed in "Conversion Hosts" page while there is only one provider
https://bugzilla.redhat.com/show_bug.cgi?id=1775998

```
Description of problem:
The "Missing Providers" message displayed is not accurate, in case of VMware -> RHV migration

Version-Release number of selected component (if applicable):
5.11.1.0

How reproducible:
100%

Steps to Reproduce:
Using fresh CFME:
1. Add RHV provider
2. Go to "Migration Settings" -> "Conversion Hosts"

Actual results:
Can't  add conversion hosts due missing providers:
"There are no providers available from which to configure a conversion host. You must configure a target provider before configuring conversion hosts"

Once adding VMware provider -> there is the option to set RHV conversion hosts

Expected results:
The message should be that there is no Source provider
OR to allow to set conversion host while only the RHV provider is configured.
```

It looks like the issue here is that for the conversion hosts page we reuse the "has sufficient providers" check we already had on plans/mappings pages, which checks that you have both VMWare and one of either RHV or OSP.

@vconzola from a UX perspective, which fix would you prefer?
* Change the message so it also mentions configuring a source provider
* Change the behavior so you don't need a VMware provider to get to the conversion hosts page

mturley:
Fix #1072 
The Conversion Hosts Settings page was sharing the same `hasSufficientProviders` logic as other pages, but the error messaging was inconsistent with that logic. Since you only need a target provider configured to set up a conversion host, I changed the logic to match the error messaging.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1775998

### DIFF
--- a/app/javascript/react/screens/App/Overview/__test__/__snapshots__/index.test.js.snap
+++ b/app/javascript/react/screens/App/Overview/__test__/__snapshots__/index.test.js.snap
@@ -34,6 +34,7 @@ Object {
   "fetchTransformationPlansUrl": "/api/dummyTransformationPlans",
   "finishedTransformationPlans": Array [],
   "hasSufficientProviders": false,
+  "hasTargetProvider": false,
   "hideConfirmModalAction": [Function],
   "hideEditPlanNameModalAction": [Function],
   "hideMappingWizard": false,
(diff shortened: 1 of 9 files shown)


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 6738c60d5e8710378c8a4b206960134643c7ba10
Date:   Fri Jan 10 12:28:42 2020 +0100

    Merge pull request #1086 from mturley/1072-fix-sufficient-providers-check

    [#1072] [BZ#1775998] Fix logic for providers check on Conversion Hosts Settings page

    (cherry picked from commit 5b4ee3b59b4b19654417a63ed072fd19e548f0c2)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1790540
```

PR #1092, opened 2020-01-19
Fix logic for detecting SSH transport method on a conversion host configuration task

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1784106

In the event that a conversion host configuration is in progress or failed, the transport method displayed on its item in the Conversion Hosts list is based on the `context_data.request_params` in the MiqTask created by the wizard. Previously, SSH was detected by looking for the `vmware_ssh_private_key` param, but the backend now [omits that param when storing `request_params`](https://github.com/ManageIQ/manageiq/blob/b8ab6d4d0ba7c47e7a810f62fdc8417d6ee2c9cf/app/models/conversion_host/configurations.rb#L45), so those tasks were being rendered without the SSH label.

This PR changes the logic to assume SSH transport if the VDDK property is not present.

The bug can be reproduced using the database `failed-conv-host-type` from [here](https://drive.google.com/drive/folders/1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q?usp=sharing).

Before:
Screenshot 2020-01-19 16 40 37

After:
Screenshot 2020-01-19 16 37 12

### DIFF
--- a/app/javascript/react/screens/App/Settings/__tests__/helpers.test.js
+++ b/app/javascript/react/screens/App/Settings/__tests__/helpers.test.js
@@ -121,26 +121,14 @@ describe('transport method helper', () => {
       expect(inferTransportMethod(task)).toEqual(__('VDDK'));
     });
 
-    it('detects SSH properly', () => {
-      const task = {
-        meta: { isTask: true },
-        context_data: {
-          request_params: {
-            vmware_ssh_private_key: 'foo'
-          }
-        }
-      };
-      expect(inferTransportMethod(task)).toEqual(__('SSH'));
-    });
-
-    it('returns null if neither match', () => {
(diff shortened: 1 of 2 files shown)


simaishi:
Ivanchuk backport details:

```
$ git log -1
commit b52aca1c33121fde96fda95481c278c25230105f
Date:   Mon Jan 20 13:32:56 2020 +0100

    Merge pull request #1092 from mturley/bz1784106-conv-host-type-when-failed

    Fix logic for detecting SSH transport method on a conversion host configuration task

    (cherry picked from commit d25db9f59856d92e6a1b8bd4e407702862a29fa8)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1784106
```

PR #1100, opened 2020-02-03
Relocate ConversionHost in V2V plugin

fdupont-redhat:
When we started coding the migration component, the logic was in Automate, including the methods for the conversion host management. Since, we created a normal class in ManageIQ/manageiq. Though, it would be more logical to host the migration code under the ManageIQ/manageiq-v2v repository, in order to follow the plugins structure. The first class that we can move is ConversionHost, as it inherits only from ActiveRecord.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1797507


fdupont-redhat:
@miq-bot add-label v2v, refactor, ivanchuk/yes

agrare:
Hey @fdupont-redhat I think you want to use the [code-extractor](https://github.com/juliancheal/code-extractor) (or similar commands manually which is what I usually do) to move the commits to those files over as well so we don't lose the history.

Fryguy:
I think this commit is busted

`fdupont-redhat wants to merge 10,000 commits into ManageIQ:master from fdupont-redhat:v2v_relocate_conversion_host_class`

Fryguy:
cc @jrafanie 

I believe moving this class in isolation would end up creating a dependency of core on the v2v repo, which is undesirable.  @fdupont-redhat would that happen in this case, and if so, is there a way to avoid it?

jrafanie:
> cc @jrafanie
> 
> I believe moving this class in isolation would end up creating a dependency of core on the v2v repo, which is undesirable. @fdupont-redhat would that happen in this case, and if so, is there a way to avoid it?

Yeah, it looks like this code depends on the Auth*, MiqTask, etc. constants from core.  From the other direction, it looks like core is looking for `ConversionHost` here: 

```
app/models/conversion_host/configurations.rb:module ConversionHost::Configurations
```

Perhaps that should be moved also so core code doesn't use v2v constants.

Fryguy:
`v2v -> core` (Auth, MiqTask, etc) dependency is ok, but I'm concerned about `core -> v2v`

PR #1095, opened 2020-01-24
Add functionality for deleting warm migration cutover date

mzazrivec:
New kebab menu item for deletion
delete-cutover-01

Confirmation dialog
delete-cutover-02

This needs to be tested with https://github.com/ManageIQ/manageiq/pull/19759 in place

PR #1104, opened 2020-02-07
Add tests for warm migration modal changes

mzazrivec:
Tests for https://github.com/ManageIQ/manageiq-v2v/pull/1070 and https://github.com/ManageIQ/manageiq-v2v/pull/1095

Part of the changes in this PR is that I'm making the tests think they're running in `UTC` timezone.

### DIFF
--- a/.jest-global-setup.js
+++ b/.jest-global-setup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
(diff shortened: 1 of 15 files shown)


.jest-global-setup.js
```diff
@@ -0,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
```
mturley: Just curious, was this not something that would work in the existing `.jest-setup` file?

mturley:
Oh, I hadn't realized the Travis CI build failed

mturley:
Just a little linter error because of an unused import:

```
/home/travis/build/ManageIQ/manageiq-v2v/app/javascript/react/screens/App/Overview/components/ScheduleMigrationModal/__test__/ScheduleMigrationModal.test.js

  2:10  error  'mount' is defined but never used  no-unused-vars
```

.jest-global-setup.js
mzazrivec: Yep, for some I reason this needs to be in `globalSetup` part of `jest` setup, otherwise `process.env.TZ` has no effect here.

mzazrivec:
@mturley fixed the unused import. Thanks.

Issue #1107, opened 2020-02-17
JS / React error in mapping wizard


mzazrivec:
1. make sure you have no openstack conversion host available & launch mapping wizard
2. select OpenStack & hit next button (to get to the `Map Compute` step)
3. see the following error in console:

```
miq_debug.self-1351f771c2cd2fab0d83069fd3f62aadbb7effc11a01ed4942f52308793453be.js?body=1:30 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ModalWizardBody`. See https://fb.me/react-warning-keys for more information.
    in Alert (created by ModalWizardBody)
    in ModalWizardBody (created by MappingWizardBody)
    in MappingWizardBody (created by MappingWizard)
    in div (created by ModalBody)
    in ModalBody (created by WizardBody)
    in WizardBody (created by MappingWizard)
    in div (created by Wizard)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in CustomModalDialog (created by Modal)
    in Transition (created by Fade)
    in Fade (created by DialogTransition)
    in DialogTransition (created by Modal)
    in RefHolder (created by Modal)
    in div (created by Modal)
    in Portal (created by Modal)
    in Modal (created by Modal)
    in Modal (created by Wizard)
    in Wizard (created by MappingWizard)
    in MappingWizard (created by ConnectFunction)
    in ConnectFunction (created by I18nProviderWrapper(Connect(MappingWizard)))
    in IntlProvider (created by I18nProviderWrapper(Connect(MappingWizard)))
    in I18nProviderWrapper(Connect(MappingWizard))
    in Mappings (created by ConnectFunction)
    in ConnectFunction (created by I18nProviderWrapper(Connect(Mappings)))
    in IntlProvider (created by I18nProviderWrapper(Connect(Mappings)))
    in I18nProviderWrapper(Connect(Mappings)) (created by Route)
    in Route (created by Routes)
    in Routes (created by App)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Context.Consumer)
    in ConnectedRouterWithContext (created by ConnectFunction)
    in ConnectFunction (created by App)
    in App (created by ConnectFunction)
    in ConnectFunction
    in Provider
window.console.<computed> @ miq_debug.self-1351f771c2cd2fab0d83069fd3f62aadbb7effc11a01ed4942f52308793453be.js?body=1:30
[...]

PR #1108, opened 2020-02-18
[#1069] [BZ#1775385] Mapping Wizard: Fix OSP conversion host check to be more strict

### LINKED ISSUE #1069: [BZ#1775385] Mapping Wizard: OSP conversion host warning check is too broad, needs to look at specific providers
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1775385

When we implemented OSP target support, we made a mistake (in https://github.com/ManageIQ/manageiq-v2v/pull/812) in the logic that detects whether the user needs to be warned that there is no conversion host configured.

Currently, if there are no OSP conversion hosts anywhere the appliance can find, the UI displays a warning on the clusters step of the wizard reading "No OpenStack conversion hosts are configured. You can continue to create an infrastructure mapping, but you must configure a conversion host before migration execution."

However, this warning just checks whether OSP conversion hosts exist at all, it does not take into account which provider/EMS they reside in. If the user has more than one OSP provider configured, they could in theory have a conversion host in provider A (and therefore not see the warning), but create a mapping to provider B which would fail.

The conversion host warning for OSP needs to be similar to the one for RHV (a warning icon on each target cluster in the list), but instead of looking for the conversion host in that specific target cluster/project, it should look for a conversion host with the same EMS id as that target cluster/project.

mturley:
Fixes #1069 
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1775385

In the case where the user has more than one OpenStack provider configured, but only has conversion hosts configured in one of them, the mapping wizard was not able to warn them that migrations to the provider with no conversion hosts would fail (it only showed a warning if there were no OSP conversion hosts anywhere).

This PR moves the alert from a wizard-level inline alert into individual alert icons next to each target project (just like RHV conversion host warnings) and appears on any project which is in a provider where there are no OSP conversion hosts.

This also updates the text of the existing RHV warning to make it consistent with the new OSP one, based on @vconzola 's copy here: https://github.com/ManageIQ/manageiq-v2v/issues/1069#issuecomment-557281392

Screenshot 2020-02-18 16 42 15

### DIFF
--- a/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
+++ b/app/javascript/react/screens/App/Mappings/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStep.js
@@ -6,22 +6,10 @@ import { length } from 'redux-form-validators';
 
 import ClustersStepForm from './components/ClustersStepForm/ClustersStepForm';
 import { FETCH_TARGET_COMPUTE_URLS } from '../../../../../../../../redux/common/targetResources/targetResourcesConstants';
-import { OPENSTACK } from '../../MappingWizardConstants';
 
 class MappingWizardClustersStep extends React.Component {
   componentDidMount() {
-    const { targetProvider, ospConversionHosts, showAlertAction } = this.props;
     this.fetchClusters();
-    if (targetProvider === OPENSTACK && ospConversionHosts.length === 0) {
-      showAlertAction(
-        __('No OpenStack conversion hosts are configured. You can continue to create an infrastructure mapping, but you must configure a conversion host before migration execution.'), // prettier-ignore
-        'warning'
-      );
-    }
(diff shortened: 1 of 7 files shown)


mzazrivec:
This PR effectively fixes https://github.com/ManageIQ/manageiq-v2v/issues/1107

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit e7e9b2ac4f63b4d98466fc85bb56d06be7bafe86
Date:   Wed Feb 19 15:21:26 2020 +0100

    Merge pull request #1108 from mturley/1069-fix-osp-conversion-host-check

    [#1069] [BZ#1775385] Mapping Wizard: Fix OSP conversion host check to be more strict

    (cherry picked from commit a879780e9ff47cc2051bc5fef3c28990455db0c9)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805910
```

PR #1109, opened 2020-02-19
[Part 3/3 of BZ#1789479] UI changes for UCI: Remove OpenStack User field from conversion host wizard, always pass "cloud-user".

mturley:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1789479
JIRA: https://issues.redhat.com/browse/MIGENG-316

UCI conversion hosts all use the username `cloud-user`, so there is no longer a reason to ask for it for OSP or default it to `root` for RHV.

Screenshot 2020-02-19 13 19 55

Screenshot 2020-02-19 13 22 42

This PR also fixes a minor issue with the `caCerts` field where it would still be included in the API request if the user entered data there and then turned off the "verify certs" switch. I added a new unit test for that while I was updating these tests.


mturley:
@simaishi it looks like you'll get conflicts here too, I'm staging a PR for it now

simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-v2v/pull/1114

PR #1110, opened 2020-02-21
Fix update of vertical menu styles after menu click

mzazrivec:
The layout, styles & classes changed a little bit when the vertical menu was re-done into React.

Previously, the menu items would not be correctly marked as inactive when navigating between sections. See screenshot.
Screenshot from 2020-02-21 11-47-49

### DIFF
--- a/app/javascript/common/menu.js
+++ b/app/javascript/common/menu.js
@@ -2,7 +2,7 @@ import http from './http';
 import { links } from '../react/config/config';
 
 const updateUrl = '/migration/menu_section_url?section=migration&url=/migration%23';
-const activeMenuItemSelector = '[data-target="#menu-migration"] > div > ul > li[class="list-group-item active"]';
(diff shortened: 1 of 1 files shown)


himdel:
Matches the menu changes from https://github.com/ManageIQ/manageiq-ui-classic/pull/5854 (and https://github.com/ManageIQ/manageiq-ui-classic/pull/6197 + https://github.com/ManageIQ/manageiq-ui-classic/pull/6463) :+1:

Ideally we should be updating the menu to read the router history object instead of manually tweaking this. But this updates the existing code to match the right elements again, so, step forward :).

himdel:
```
app/javascript/common/menu.js:21
  const newActiveMenuItem = document.querySelector(`[id="${link_item.menu_item_id}"]`);
```

won't that need an id prefix update too?

mturley:
@himdel thanks for merging these, I had been meaning to get around to them

Issue #559, opened 2018-08-09
Remove react-intl


priley86:
Per [discussion](https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/533#issuecomment-411752992) in #533 , we can safely remove the react-intl from this project. 

This can be done POST GA as a small cleanup task (should not be pressing now).

AllenBW:
Woops, so we *do* actually use this dep a bit, formatting times yes, could be substituted with moment, [but we also do a bit of local formatting here](https://github.com/ManageIQ/manageiq-v2v/blob/4ab98e0f9323bdedac1365aac1c94a2fd81c5d6a/app/javascript/common/i18n.js)

mturley:
Very low priority, but I'm not sure whether we can remove this or whether it would matter at this point. @mzazrivec since you have been involved with i18n, do you have any thoughts on this? Some of the places we use it in our code seem to be stale but I'm not sure.

mzazrivec:
Yes, I think `react-intl` can be removed.

PR #1111, opened 2020-02-21
Remove react-intl & related code

mzazrivec:
Fixes https://github.com/ManageIQ/manageiq-v2v/issues/559


himdel:
LGTM, neigther the library nor the date time helpers are used anywhere :)
And it still builds :)

PR #1115, opened 2020-02-26
[BZ#1807477] Replace info message for Conversion Host SSH Private Key field

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1807477

After discussing with @fdupont-redhat and @vconzola we determined that for both RHV and OpenStack, the info message on this field should say "This is the private key file used to connect to the conversion host instance for cloud-user", which is almost the same as the existing message for OpenStack (except that the old message referred to the OpenStack User field we removed, so instead it should specifically reference the `cloud-user` username).

Screenshot 2020-02-26 10 29 40


mturley:
@simaishi, although this is now targeted for 5.11.5, I wanted to get ahead of backporting conflicts so I opened this PR early: https://github.com/ManageIQ/manageiq-v2v/pull/1116

mturley:
cc @fdupont-redhat, I think if there are any critical 5.11.4 BZs that will cause QE to have to do another build, we should maybe ask them if we can move this back to 5.11.4?

simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-v2v/pull/1116

mturley:
@simaishi since https://github.com/ManageIQ/manageiq-v2v/pull/1116 was reverted and the BZ is still in POST, should this be set back to `ivanchuk/yes`?

PR #1119, opened 2020-03-20
Remaining warm migration UI changes (Plans and Plan Details views)

mturley:
Replacement for #1077
Closes https://github.com/ManageIQ/manageiq-v2v/issues/1023
Closes https://github.com/ManageIQ/manageiq-v2v/issues/433

Should be the final piece for https://bugzilla.redhat.com/show_bug.cgi?id=1814281.

This PR implements the remaining changes from [Vince's warm migration mockups](https://docs.google.com/presentation/d/1DeL3tn2WtoOZFdfQK4wHim99l8vrnfCCnsAqKlWssZs/edit#slide=id.g5bc20f1595_0_2) now that we have real warm migration data to test against.

I saved several database dumps from various stages of UCI-based migrations (warm and cold) from an appliance last week, which can be found here: https://drive.google.com/drive/u/1/folders/1edZSrfZvSw8jjLtuI-HEuAynKIio3hhN

From that collection, here are a few useful databases to reproduce various states of warm migrations:
* `uci-migrations-7` has a cold migration and a warm migration in progress at the same time, to demonstrate the `colSpan` layout. Scheduling a cutover on the warm migration shows how the cold migration progress bar spans both the pre-copy status and the cutover time info items.
* `uci-migrations-12` has a warm migration initializing.
* `uci-migrations-13` has a warm migration running its initial pre-copy (progress bar visible on the plan details page).
* `uci-migrations-20-many-copies` has a warm migration with several pre-copies completed, ready for cutover but with no cutover scheduled.
* `uci-migrations-22-shutting-down` has a warm migration just after cutover starts, shutting down its source VM.
* `uci-migrations-23-transforming` has a warm migration mid-cutover, transforming the VM.
* `uci-migrations-27-success` has a warm migration in the successfully complete state.

I don't currently have any databases with failed pre-copies, but you can test those views by modifying the return value of `reduceCopies` in `warmMigrationHelpers.js` and set `finishedWithErrors` to true. We don't yet have error status/messages for each pre-copy in the API, this naive error case will only occur if a pre-copy ended without copying all of its data.

This PR also rearranges some of the columns and labels on the plans not started view and the plans completed view, based on these mockups.

# Screens

These were taken with database dumps about 9 days old, so disregard the durations and timestamps.

## Running initial pre-copy

In the plans in progress view:
Screenshot 2020-03-19 23 01 12-plans-initial-precopy

In the plan details view (expanded):
Screenshot 2020-03-19 22 57 06-details-initial-precopy

[...]

PR #1123, opened 2020-04-03
[BZ#1817096] Support for two-phase cold migrations

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=1817096

Per @fdupont-redhat, when a migration task has `task.options.two_phase`, its progress will be reported in the same new format as warm migration (`virtv2v_disks[].copies`), just with only one copy object per disk.

This PR checks for that flag, and if present, reuses the same code written for the "latest precopy" progress bar in warm migration instead of the overall disk copy progress bar previously used. Just in case a migration may be in progress and not have a `copies` object yet or have one with a `to_copy: 0` value, it ensures the progress bar will always render and just have "Unknown" for the total value text in those cases.

This PR also fixes an issue caused by #1119 where console errors could occur when viewing the plan details page on smaller screens (8c952e0).

Depends on https://github.com/ManageIQ/manageiq/pull/20011


mzazrivec:
The changes here look good, though I will wait for a DB to properly test this with.

mturley:
@mzazrivec I captured some databases we can use to test this here: https://drive.google.com/drive/folders/1sdcGfLfGwwppPqCvuXHVKo8rIU-AvLHW?usp=sharing

I dumped the db several times while running a migration, any of the ones with the `copying` suffix in the name will show the new progress bar.

simaishi:
Jansa backport details:

```
$ git log -1
commit 57b015c62d2ef0a764c14f9f1b592f5b0390f589
Date:   Wed Apr 8 13:53:22 2020 +0200

    Merge pull request #1123 from mturley/bz1817096-two-phase-cold-support

    [BZ#1817096] Support for two-phase cold migrations

    (cherry picked from commit 972ee7960cdaa48e59d9055a6602973e1f7c0f3f)

    https://bugzilla.redhat.com/show_bug.cgi?id=1817096
```

PR #1129, opened 2020-05-05
[#1047][BZ#1760001] Fix inconsistent scrollbar behavior by removing fixed headers and just having one scroll body for each page

### LINKED ISSUE #1047: [BZ#1760001] Scrollbar container height issues on the migration plan details page
https://bugzilla.redhat.com/show_bug.cgi?id=1760001

Description of problem:

When viewing the details page of a migration in progress, if there are more VMs than can fit in one page, scrolling the window to the bottom does not work correctly.

The problem presents itself differently in different browsers. In Firefox, the page stops scrolling before the bottom, so it is impossible to reach the pagination controls. In Chrome, the page scrolls beyond the bottom, leaving a bunch of white space below the list.

See attached screenshots taken in Firefox and Chrome while scrolled as far down as the browser will allow.

Firefox:
scroll-issue-firefox

Chrome:
scroll-issue-chrome

Version-Release number of selected component (if applicable):

How reproducible:
100%

Steps to Reproduce:
1. Open a plan details page with a large number if VMs
2. Scroll to the bottom

Actual results:
The pagination controls are either unreachable or move past the bottom of the window

Expected results:
The pagination controls are visible just at the bottom of the window

----

This bug can be reproduced using the `large-migration-in-progress` database dump from [my collection](https://drive.google.com/drive/u/1/folders/1psxWLopcPUFeJCoHcnMiOgF40x2V0h3Q).

mturley:
Closes #1047 
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1760001

In manageiq-ui-classic, the bodies of tables and list views have their height [set by JavaScript when the window resizes](https://github.com/ManageIQ/manageiq-ui-classic/blob/ba88d4c4dab3620c22476e69f5d3a070de460173/app/assets/javascripts/miq_application.js#L1324-L1342) so that they can have their own scrollbar and the breadcrumb, toolbars, and footer can be pinned outside that scrollbar.

We had tried to replicate that with CSS since we can't reuse that same window resize code (it would be overridden by React whenever our components re-rendered). The CSS solution was broken though, causing the issue in this BZ. To properly fix it would have meant restructuring the container to use proper flexbox CSS, which is difficult with the PatternFly 3 CSS (which is probably why JS was used to solve it in ui-classic) and may have had unintended consequences.

This PR simply removes the attempt to have these dynamic-height content areas, and makes each of our pages one whole scroll container. This fixes the bug in a predictable way with minimal drawback.

### DIFF
--- a/app/javascript/react/screens/App/App.scss
+++ b/app/javascript/react/screens/App/App.scss
@@ -11,3 +11,17 @@
 .required-asterisk {
   color: $color-pf-red-100;
 }
+
+.main-scroll-container {
+  height: 100%;
+  overflow-y: auto;
+  overflow-x: hidden;
+  margin-left: -20px;
+  margin-right: -20px;
+  padding-left: 20px;
+  padding-right: 20px;
+}
+
+.main-body-content {
+  margin-bottom: 20px;
+}
--- a/app/javascript/react/screens/App/Mappings/Mappings.js
+++ b/app/javascript/react/screens/App/Mappings/Mappings.js
@@ -176,7 +176,7 @@ class Mappings extends Component {
     } = this.props;
 
     return (
-      <React.Fragment>
+      <div className="main-scroll-container">
         <MigrationBreadcrumbBar activeHref="#/mappings" productFeatures={productFeatures} />
         <Spinner
           loading={
@@ -228,7 +228,7 @@ class Mappings extends Component {
           )}
         </Spinner>
         {mappingWizardVisible && this.mappingWizard}
-      </React.Fragment>
+      </div>
     );
   }
 }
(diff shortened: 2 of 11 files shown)


simaishi:
Backported to ivanchuk via https://github.com/ManageIQ/manageiq-v2v/pull/1130

simaishi:
Jansa backport details:

```
$ git log -1
commit d162003e511d59cae8425b9d33a1c3111d112c61
Date:   Mon May 11 13:55:22 2020 +0200

    Merge pull request #1129 from mturley/1047-scrollbar-bug

    [#1047][BZ#1760001] Fix inconsistent scrollbar behavior by removing fixed headers and just having one scroll body for each page

    (cherry picked from commit 4d797aaade4bea99d1374612a4b97b030deb2a35)

    https://bugzilla.redhat.com/show_bug.cgi?id=1760001
```

PR #1149, opened 2020-11-18
[KASPAROV] yarn: lock versions by adding yarn.lock

himdel:
parent https://github.com/ManageIQ/manageiq-ui-classic/pull/7515


himdel:
(travis goes green with https://github.com/ManageIQ/manageiq-v2v/pull/1151)

PR #1153, opened 2021-01-27
Use TEST_SUITE env var to drive specs

agrare:
This probably isn't the objectively best way to do this, however it does follow the TEST_SUITE pattern used by core and ui-classic and makes it easier for us to run with manageiq-cross_repo

https://github.com/ManageIQ/manageiq-cross_repo/pull/69

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,15 @@
-matrix:
-  include:
-  - name: "Node.js: 12"
-    language: node_js
-    cache: yarn
-    node_js:
-    - '12' # current stable
-    script: yarn test --runInBand
-  - name: "Ruby: 2.5.8"
-    language: ruby
-    cache: bundler
-    rvm:
-    - 2.5.8
-    addons:
-      postgresql: '10'
-    install: bin/setup
-    after_script: bin/ci/after_script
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,26 +1,19 @@
...
+language: ruby
+node_js: 12
+cache:
+  bundler: true
+  yarn: true
+addons:
+  postgresql: '10'
+rvm:
+- 2.5.8
+- 2.6.6
```
Fryguy, line 11: Super minor, but I think we should keep the languages/versions together, so can this be moved up between the language and node_js key?

Fryguy (approved):

.travis.yml
agrare: :+1: done

agrare:
Having a hell of a time getting the javascript tests to pass 😱

agrare:
Okay got this running.  Instead of having `bundle exec rake` and the default rake task switching on the `ENV["TEST_SUITE"]` I added a `bin/ci/script.sh` so that the javascript tests don't have to `bin/setup` which takes an order of magnitude longer to `bundle install` and setup the database when it won't end up being used anyway.

